PR #12, opened 2024-06-24
MWPW-152283: add acom wcs endpoint

3ch023:
Resolve https://jira.corp.adobe.com/browse/MWPW-152283

Oversimplified ability to test acom wcs domain performance (should take away OPTIONS request from stage.adobe.com)
To see this in action add `?commerce.env=STAGE&domain.switch=true`

image

Test URLs:
- Before: https://main--milo--adobecom.hlx.page/drafts/mili/promo-placeholders?martech=off&commerce.env=STAGE&domain.switch=true
- After: https://mwpw-152283--milo--adobecom.hlx.page/drafts/mili/promo-placeholders?martech=off&commerce.env=STAGE&domain.switch=true

### DIFF
--- a/commerce/src/defaults.js
+++ b/commerce/src/defaults.js
@@ -22,6 +22,7 @@ export const Defaults = Object.freeze({
     entitlement: false,
     extraOptions: {},
     modal: false,
+    upgrade: false,
     promotionCode: '',
     quantity: 1,
     wcsApiKey: 'wcms-commerce-ims-ro-user-milo',
--- a/commerce/src/defaults.js
+++ b/commerce/src/defaults.js
@@ -22,7 +22,6 @@ export const Defaults = Object.freeze({
     entitlement: false,
     extraOptions: {},
     modal: false,
-    upgrade: false,
     promotionCode: '',
     quantity: 1,
     wcsApiKey: 'wcms-commerce-ims-ro-user-milo',
--- a/commerce/src/index.d.ts
+++ b/commerce/src/index.d.ts
@@ -576,6 +576,7 @@ declare global {
                 wcsBufferDelay: number;
                 wcsBufferLimit: number;
                 wcsEnv: Environment;
+                domainSwitch: boolean;
             }
         }
     }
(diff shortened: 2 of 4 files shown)


commerce/src/wcs.js
```diff
@@ -32,12 +33,12 @@ const WcsBaseUrl = {
  */
 export function Wcs({ settings }) {
     const log = Log.module('wcs');
-    const { env, wcsApiKey: apiKey } = settings;
-
+    const { env, domainSwitch, wcsApiKey: apiKey } = settings;
+    const baseUrl = domainSwitch ? WcsBaseUrl['STAGE_ACOM'] : WcsBaseUrl[env];
```
npeltier, line 37: this is problematic in case you are domainSwitch on and env production, no? i guess we won't have both normally, but still looks weird

commerce/src/settings.js
```diff
@@ -217,6 +219,7 @@ function getSettings(config = {}) {
         entitlement,
         extraOptions: Defaults.extraOptions,
         modal,
+        upgrade,
```
npeltier, line 222: please don't mix things up

commerce/src/defaults.js
```diff
@@ -22,6 +22,7 @@ export const Defaults = Object.freeze({
     entitlement: false,
     extraOptions: {},
     modal: false,
+    upgrade: false,
```
npeltier: please don't mix things up

commerce/src/settings.js
3ch023: okie, will revert

commerce/src/wcs.js
3ch023: i added acom prod too

commerce/src/index.d.ts
```diff
@@ -576,6 +576,7 @@ declare global {
                 wcsBufferDelay: number;
                 wcsBufferLimit: number;
                 wcsEnv: Environment;
+                domainSwitch: boolean;
```
yesil, line 579: do we really need to introduce a new parameter with a confusing name:`domainSwitch` ?
I would like to avoid having to support temporarily introduced parameters for ever.
Can we just switch to this new domain?

yesil (commented):
not sure `domainSwitch` param was discussed before, but I think we should avoid introducing new params.

commerce/src/index.d.ts
npeltier: @yesil we need both domains to work so we can compare stuff from same domain

npeltier (approved):

PR #19, opened 2024-06-27
MWPW-149125 and MWPW-147035

bozojovicic:
These are 2 TWP tickets reported long time ago but I could not implement them before TWP code in Milo is merged to Stage. [MWPW-147035](https://jira.corp.adobe.com/browse/MWPW-147035) is also blocked by one bug in TWP in Milo code (Card title not displayed) but I will provide the fix for that bug in Milo PR after this PR is merged.

Fix # [MWPW-149125](https://jira.corp.adobe.com/browse/MWPW-149125) Suppress Stock Checkbox on all modals for mobile users only
Fix # [MWPW-147035](https://jira.corp.adobe.com/browse/MWPW-147035) Ability to deeplink into a card or tab within the modal

Test URLs:
- After: https://mwpw147035--mas--bozojovicic.hlx.live

### DIFF
--- a/web-components/src/merch-twp-d2p.js
+++ b/web-components/src/merch-twp-d2p.js
@@ -216,7 +216,7 @@ export class MerchTwpD2P extends LitElement {
         this.singleCard = card;
     }
 
-    unSelectSingleCard() {
+    async unSelectSingleCard() {
         if (!this.singleCard) return;
         this.singleCard.setAttribute(
             'slot',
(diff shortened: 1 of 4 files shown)


web-components/src/merch-twp-d2p.js
```diff
@@ -370,7 +374,9 @@ export class MerchTwpD2P extends LitElement {
     }
 
     get preselectedCardId() {
-        const preselectedCardIds = parseState()['select-cards']?.split(',').reduce((res, item) => {
+        const params = new URLSearchParams(window.location.search);
```
yesil, line 377: we should not use the `search` params but `hash`.

npeltier:
back to backlog

PR #25, opened 2024-07-05
MWPW-153658 audit script

npeltier:
- parse one or several url or sitemap files (can be added in manifest),
- for each url from sitemap, parse page for personnalization fragment or fragment,
- for each ost link found in html, keep usage,
- for each osi, get wcs entry,
- output everything as CSV

Test URLs:
- After: https://MWPW-153658--mas--npeltier.hlx.live/


misc/audit.mjs
```diff
@@ -0,0 +1,297 @@
...
+            }
+        });
+    }
+    return result;
+}
+
+const prefixWcsKey = (key) => ('wcs ' + key);
+
+async function getCommerceData(osi) {
+    const response = await fetch(wcsUrl(osi));
```
yesil, line 127: Is that okay if the OSI doesn't return an offer anymore via WCS?  AOS can always deliver the parameters of an OSI.

yesil (approved):
It looks very good to me.

In https://github.com/adobecom/mas/pull/22, there is eslint/prettier lint rules. If possible I suggest to lint your code with it before the initial merge.

misc/audit-manifest.txt
```diff
@@ -0,0 +1,186 @@
+https://www.adobe.com/cc-shared/assets/sitemap.xml
+https://www.adobe.com/au/cc-shared/assets/sitemap.xml
+https://www.adobe.com/cn/cc-shared/assets/sitemap.xml
```
3ch023, line 3: where does the list of sitemaps coming from?

3ch023 (approved):
approving, although sitemaps list is missing homepage, but we can add it later

misc/audit-manifest.txt
npeltier: i created it out of the locales we have in our code

3ch023: we should probably add a disclamer that these are not all Milo pages, wdyt?

misc/audit.mjs
npeltier: i think i prefer usage of WCS as blank answer there is also something we should be worried about

misc/audit-manifest.txt
npeltier: this manifest is here for somebody (me for now) to use. Nothing say this manifest is here to get everything :) (i wait for discussion with business for this)

PR #29, opened 2024-07-09
MWPW-152539 Strikethrough price is not displayed for legacy promo offers

bozojovicic:
This is requested by Drew few months ago (to show old prices for legacy promo offers [example page](https://main--cc--adobecom.hlx.page/drafts/drew/old-price-drew-copy)) and I implemented it and it was part of another PR that we decided we will not merge for some time. **I will never combine multiple fixes in one PR again.** 
Anyway, from the commerce point of view this code is fine. But there is one more problem : For these legacy promo offers we always get `data-display-old-price="true"` even when we have old=false in merch link and that makes my change here pointless. Simply this method https://github.com/adobecom/milo/blob/stage/libs/blocks/merch/merch.js#L424 always sets `displayOldPrice` to undefined if `context.promotionCode` is undefined, and for legacy promo offers it is undefined and then it sets `data-display-old-price` to default value `true`.
If we simply change that line in Milo code to `const displayOldPrice = params.get('old');` it will be OK, but can we do that?

NOTE: If we merge this PR and we don't do anything in Milo code we will always show old price for legacy promo offers no matter what is the value of `old` parameter in merch link. 

Fix #[MWPW-152539](https://jira.corp.adobe.com/browse/MWPW-152539)

Test URLs:
- After: https://mwpw152539legacypromo--mas--bozojovicic.hlx.live/


yesil (approved):
Looks good to me, thanks.

Axelcureno (approved):
lgtm

3ch023:
Closing this PR since it will need to be reopened in Milo directly

PR #31, opened 2024-07-09
MWPW-151445 MWPW-151238 Show per license and hide old price by default

bozojovicic:
We had this change merged to tacocat before in [PR 612](https://git.corp.adobe.com/wcms/tacocat.js/pull/612) but then we had to rollback it until we see the audit results for "per license label" and "display old price". Now we have the audit reports and at some point GWP will fix the content and this will be ready for merging. Before that happens let's just keep it open. Once again **do not merge** before GWP say it is OK.
I know I said I will not push 2 fixes in the same PR but these 2 changes share the same code, the same tests, so ...

Fix #[MWPW-151238](https://jira.corp.adobe.com/browse/MWPW-151238)
Fix #[MWPW-151445](https://jira.corp.adobe.com/browse/MWPW-151445)

Test URLs:
- After: https://mwpw151445and151238--mas--bozojovicic.hlx.live/


3ch023:
Closing this PR since it will need to be reopened in Milo directly

PR #32, opened 2024-07-10
Automation & Git hooks

3ch023:
In this PR:

### Pre Push hook
- before each push run 'npm run build:precommit' command
- 'npm run build:precommit' will generate files under /libs, but without banner indicating branch/commit
- the hook will stage the files that have changed (if any), and commit them before the push

Basically if you touched /web-components/src/constants.js, once you push there will be an additional commit containing rebuilt /libs/merch-card.js etc.

Test URLs:
- After: https://commithook--mas--3ch023.hlx.live/


.husky/pre-push
```diff
@@ -0,0 +1,22 @@
+# Check for unstaged changes
+if ! git diff-index --quiet HEAD --; then
+  echo "uou have unstaged or uncommitted files. commit, stash or discard your changes before pushing."
```
yesil, line 3: typo `uou`

.husky/pre-push
```diff
@@ -0,0 +1,22 @@
...
+#build artifacts
+npm run build:precommit
+echo "successfully built /libs and mas.js"
+if git diff-index --quiet HEAD --; then
+  echo "changes detected, will commit"
+  git add libs
+  git add mas.js
+  git commit -m '[pre-push hook] commit artifacts'
+  echo "committed /libs and mas.js"
+fi
```
yesil, line 22: eof

.husky/pre-push
yesil: before `before commiting` instead of `before pushing`?

.husky/pre-push
```diff
@@ -0,0 +1,22 @@
...
+  exit 1
+fi
+
+#build artifacts
+npm run build:precommit
+echo "successfully built /libs and mas.js"
+if git diff-index --quiet HEAD --; then
+  echo "changes detected, will commit"
+  git add libs
+  git add mas.js
```
yesil, line 19: smart 👍

.husky/pre-push
3ch023: why before committing? 
it asks to do smthg before push with uncommitted or unstaged changes
i can simplify to 'Please commit or revert your changes before push', wdyt?

yesil: is this a commit hook or git push hook?

yesil (dismissed):

.npmrc
```diff
@@ -0,0 +1 @@
+registry = https://registry.npmjs.org/
```
yesil, line 1: this is by default, is the intent here to force the default repository and ignore the local one?

yesil (approved):

vladen (approved):

.npmrc
3ch023: yes, since many of us have a local .npmrc with corp artifactory, this will prevent someone from accidentally committing the package-lock.json with corp artifactory

should save time debugging for devs

PR #37, opened 2024-07-12
MWPW-153245 [merch-card] dispatch sort & search event for analytics

bozojovicic:
When "click" event is fired from shadow dom and caught on root node, the target element is "merch card collection" so we can't really know where the click happened. So I had to fire custom events for sort and search change. It will be handled in CC code here [catalog.js](https://github.com/bozojovicic/cc/blob/mwpw153245/creativecloud/blocks/catalog/catalog.js#L50-L61)
Also have in mind that new version of `mas/libs/merch-sidenav.js` needs to be copied to  `cc/creativecloud/deps/merch-sidenav.js`
(I created the branch with the working code but can't create PR for CC before this PR is merged.)
In the event handler I had to go from the target node up the dom structure to collect values of all daa-lh attributes to manually construct the event name for analytics since this analytics request is also fired manually with 
`window._satellite?.track('event', { ... })`

Event name for sort : `alphabetical--sort|s2|b1|catalog|nopzn|catalog-devicet`
Event name for search, if user searched for "lightroom" : `lightroom--search|s2|b1|catalog|nopzn|catalog-devicet|b1|sidenav|b1|search`
Event name for Show More : `show-more|s2|b1|catalog|nopzn|catalog-devicet|all--cat`
Event name for Card Action Menu : `action-menu|s2|b1|catalog|nopzn|catalog-devicet|all--cat`
Event name for Card Icon click : `merch-icon-click|s2|b1|catalog|nopzn|catalog-devicet|photo--cat`

Test page [here](https://mwpw153245--cc--bozojovicic.hlx.live/products/catalog?milolibs=mwpw153245--milo--bozojovicic#category=all)

Fix #[MWPW-153245](https://jira.corp.adobe.com/browse/MWPW-153245)

Test URLs:
- After: https://mwpw153245--mas--bozojovicic.hlx.live/

### DIFF
--- a/web-components/src/merch-search.js
+++ b/web-components/src/merch-search.js
@@ -17,16 +17,31 @@ export class MerchSearch extends LitElement {
 
     constructor() {
         super();
-        this.handleInput = () =>
+
+        this.handleInput = () => {
             pushStateFromComponent(this, this.search.value);
+        };
+        this.handleInputAndAnalytics = () => {
+            pushStateFromComponent(this, this.search.value);
+            if (this.search.value) {
+                this.dispatchEvent(
+                    new CustomEvent('search-changed', {
+                        bubbles: true,
+                        composed: true,
+                        detail: { value: this.search.value },
+                    })
+                );
+            }
+        };
         this.handleInputDebounced = debounce(this.handleInput.bind(this));
+        this.handleChangeDebounced = debounce(this.handleInputAndAnalytics.bind(this));
     }
 
     connectedCallback() {
         super.connectedCallback();
         if (!this.search) return;
         this.search.addEventListener('input', this.handleInputDebounced);
-        this.search.addEventListener('change', this.handleInputDebounced);
+        this.search.addEventListener('change', this.handleChangeDebounced);
         this.search.addEventListener('submit', this.handleInputSubmit);
         this.updateComplete.then(() => {
             this.setStateFromURL();
--- a/web-components/src/merch-search.js
+++ b/web-components/src/merch-search.js
@@ -1,5 +1,6 @@
 import { html, LitElement, css } from 'lit';
 import { debounce } from './utils';
+import { EVENT_MERCH_CHANGE } from './constants';
 import {
     deeplink,
     pushStateFromComponent,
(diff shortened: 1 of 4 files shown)


web-components/src/merch-search.js
```diff
@@ -17,16 +17,31 @@ export class MerchSearch extends LitElement {
...
+                this.dispatchEvent(
+                    new CustomEvent('search-changed', {
+                        bubbles: true,
+                        composed: true,
+                        detail: { value: this.search.value },
+                    })
+                );
+            }
+        };
         this.handleInputDebounced = debounce(this.handleInput.bind(this));
```
npeltier, line 36: do you still need this?

web-components/src/merch-search.js
```diff
@@ -17,16 +17,31 @@ export class MerchSearch extends LitElement {
...
-        this.handleInput = () =>
+
+        this.handleInput = () => {
+            pushStateFromComponent(this, this.search.value);
+        };
+        this.handleInputAndAnalytics = () => {
             pushStateFromComponent(this, this.search.value);
+            if (this.search.value) {
+                this.dispatchEvent(
+                    new CustomEvent('search-changed', {
```
npeltier, line 28: can you use a constant?

npeltier (changes requested):
Looks great! you also need to remove listeners you added in merch-search's disconnectedCallback

web-components/src/merch-search.js
bozojovicic: Yes, now all events have the same name 
`export const EVENT_MERCH_CHANGE = 'merch-change';`
we can see in event.detail.type from where that event is thrown.

web-components/src/merch-search.js
bozojovicic: Yes, we have 2 events for that search field: `input` and `change`. 
`Input` is thrown for every new letter and search is performed in the handler for this event
`Change` is thrown when the search field loses focus and then we send the analytics request.
I cannot send it from the `input` handler, it would go after every letter is typed or deleted.

npeltier:
discussed this offline with @bozojovicic will close this PR and re-reraise a new one as this is high priority

PR #50, opened 2024-08-02
remaining Studio changes

Axelcureno:
Remaining changes needed for running RTE in merch at scale studio

Test URLs:
- After: https://studio--mas--adobecom.hlx.live/


yesil:
@Axelcureno can we clean bundles from libs that have been moved to Milo?

Axelcureno:
> @Axelcureno can we clean bundles from libs that have been moved to Milo?

@yesil yes of course, I have cleaned up those files.

mas/src/mas.js
```diff
@@ -10,14 +8,40 @@ const features = searchParams.get('features');
...
 const config = () => ({
     env: { name: envName },
     commerce: { 'commerce.env': commerceEnv },
     locale: { prefix: locale },
 });
 
-init(config);
+(async () => {
+    try {
+        await loadScript(
```
rahulgupta999, line 30: can this URL be put in a config file and based on we are running on local or hosted, we can pick the intended URL

yesil: actually it should either be www.stage.adobe.com or www.adobe.com.
We can change it later as well once we decide how to handle environments on MAS platform.

mas/src/mas.js
```diff
@@ -10,14 +8,40 @@ const features = searchParams.get('features');
 const envName = isStage ? 'stage' : 'prod';
 const commerceEnv = isStage ? 'STAGE' : 'PROD';
 
+function loadScript(url) {
```
yesil, line 11: this should not be needed. 
We should import the file not load it this way.

yesil (changes requested):

mas/src/mas.js
Axelcureno: fixed

mas/src/mas.js
Axelcureno: removed

Axelcureno:
Closing this PR since https://github.com/adobecom/mas/pull/56 contains this and more changes.

PR #53, opened 2024-08-15
ODIN-125: Update default M@S Studio Odin env bucket to QA

vladen:
Fix [JIRA ODIN-105](https://jira.corp.adobe.com/browse/ODIN-105)

Test URLs:
- After: https://odin-105--mas--adobecom.hlx.live/


Axelcureno (approved):
What will happen to previous bucket?

vladen:
Trivial

vladen:
> What will happen to previous bucket?

DEV env is very hard to configure in a correct way, QA should be preferred to save time.

PR #57, opened 2024-09-23
variant picker

npeltier:
(plus new variants)


studio/src/editors/merch-card-editor.js
```diff
@@ -1,5 +1,6 @@
 import { html, LitElement, nothing } from 'lit';
 import { unsafeHTML } from 'lit/directives/unsafe-html.js';
+import { renderVariantPicker } from './variant-picker';
```
yesil, line 3: let's keep `.js` at the end of local imports as a rule.

yesil (approved):

PR #58, opened 2024-09-24
MWPW-159027 introduce milolibs in studio

npeltier:
- whatever milo branch for mas.js can be switched to with milolibs parameter,
- you can switch to local dev with 'local' value

Test URLs:

Before: https://main--mas--adobecom.hlx.live/
After: https://MWPW-159027--mas--adobecom.hlx.live/

### DIFF
--- a/studio/libs/maslib.js
+++ b/studio/libs/maslib.js
@@ -1,0 +1,16 @@
+const getMiloLibs = () => {
+    const pageUrl = new URL(window.location.href);
+    const milolibs = new URLSearchParams(pageUrl?.searchParams)?.get('milolibs');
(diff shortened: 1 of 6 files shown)


Axelcureno (approved):

studio/libs/maslib.js
```diff
@@ -0,0 +1,16 @@
+const getMiloLibs = () => {
+    const pageUrl = new URL(window.location.href);
+    const milolibs = new URLSearchParams(pageUrl?.searchParams)?.get('milolibs');
```
yesil, line 3: nit: new URLSearchParams(window.location.search)

yesil (approved):

PR #62, opened 2024-09-27
MWPW-159269 add size field

npeltier:
Test URLs:
- After: https://MWPW-159269--mas--adobecom.hlx.live/


Axelcureno (approved):

3ch023 (commented):
Do we need to add details of how to specify the size?
e.g. S/L/M or small/large or pixels

also, another thought, we could put the 'Optional' into the name field to make it more explicit to GWP what fields are not necessary, wdyt?

3ch023 (approved):

npeltier:
> also, another thought, we could put the 'Optional' into the name field to make it more explicit to GWP what fields are not necessary, wdyt?

i think that we have not dealt with the variant -> fields mapping in authoring yet (even if we did it for hydration), and at that moment yes, we could foresee such things (even check box for slice card "is wide ?"). It makes even more sense not having size field at all for lot of variants

3ch023:
> > also, another thought, we could put the 'Optional' into the name field to make it more explicit to GWP what fields are not necessary, wdyt?
> 
> i think that we have not dealt with the variant -> fields mapping in authoring yet (even if we did it for hydration), and at that moment yes, we could foresee such things (even check box for slice card "is wide ?"). It makes even more sense not having size field at all for lot of variants

We could add it later, true.
So what are the possible values for the size field?

npeltier:
> So what are the possible values for the size field

PR #65, opened 2024-10-04
MWPW-158473: Top navigation wc

Axelcureno:
This PR adds the top navigation web component to merch at scale studio.
Screenshot 2024-10-04 at 2 24 56 PM

[MWPW-159622](https://jira.corp.adobe.com/browse/MWPW-159622)

Test URLs:
- After: https://mwpw-158473--mas--adobecom.hlx.page/studio.html


studio/src/top-nav.js
```diff
@@ -0,0 +1,80 @@
...
+                        label="Help"
+                        placement="bottom-end"
+                </sp-top-nav-item>
+                </sp-top-nav-item>
+            </sp-top-nav>
+        `;
+    }
+}
+
+customElements.define('top-nav', TopNav);
```
yesil, line 80: eof

studio/src/top-nav.js
```diff
@@ -0,0 +1,80 @@
...
+        return html`
+            <sp-top-nav>
+                <sp-top-nav-item
+                    class="logo"
+                    size="l"
+                    href="#"
+                    label="Home"
+                    quiet
+                >
+                    <svg
```
yesil, line 46: can the SVG be an external file?

yesil (approved):
it would be nice to address this in its own ticket.
I would like us to discuss the requirements w.r.t. toolbars.
I'm ok to merge as it makes Studio visually more appealing but we will have to rework this.

I guess we can use https://www.adobe.com/federal/assets/svgs/adobe-logo.svg as icon.

Only ask is to externalize SVG files.

studio/src/top-nav.js
Axelcureno: added

studio/src/top-nav.js
Axelcureno: yes, done.

3ch023 (approved):

PR #67, opened 2024-10-08
support https in the proxy

yesil:
This PR introduces mas proxy for Odin to use on the hlx domains.

It is already configured and functional.
Resolve: https://jira.corp.adobe.com/browse/MWPW-160093

Test URLs:
- After: https://mwpw-160093--mas--adobecom.hlx.live/studio.html#query=acom


3ch023 (commented):
@yesil could you add details where the proxy is configured to https://wiki.corp.adobe.com/display/adobedotcom/M@S2+Developer+Onboarding?

Meaning if tomorrow I want to switch https://mas-proxy-odin.corp.adobe.com to point to Odin Stage, how do I do that?

yesil:
@3ch023 it is already in the doc: https://github.com/adobecom/mas/tree/main/studio#running-the-proxy

3ch023 (approved):

studio/proxy-server.mjs
```diff
@@ -11,25 +13,40 @@ const corsHeaders = {
...
+    // Read the SSL certificate and private key
+    const httpsOptions = {
+        key: readFileSync(keyPath),
+        cert: readFileSync(certPath),
+    };
+
+    server = createHttpsServer(httpsOptions, requestHandler);
+    serverProtocol = 'https';
+    serverPort = 8443; // Use 8443 for HTTPS
+} else {
```
npeltier, line 41: why leaving http option? i would remove it tbh

npeltier (approved):
thanks! hope it won't be useful very long :)

studio/proxy-server.mjs
yesil: it is needed on local. If we drop local proxy support we can remove it.

PR #68, opened 2024-10-09
MWPW-160105: New search toolbar, filter/sort panel

Axelcureno:
This PR introduces a new search toolbar with sort/filter options along with a filter panel.
Filter options are not functional but provide a template for when the final UX is ready and approved.

Screenshot 2024-10-08 at 5 27 27 PM

Resolves: https://jira.corp.adobe.com/browse/MWPW-160105

Test URLs:
- After: https://MWPW-160105--mas--adobecom.hlx.live/studio.html


yesil (approved):
Looks very good to me, thanks.

3ch023 (approved):

PR #69, opened 2024-10-09
MWPW-160085: MAS Studio updates

yesil:
Issues found during the doc update & authoring of CCD cards.
Resolve: https://jira.corp.adobe.com/browse/MWPW-160085

ccd: https://mwpw-160085--mas--adobecom.hlx.live/studio.html#query=ccd (DO NOT UPDATE CARDS)
acom: https://mwpw-160085--mas--adobecom.hlx.live/studio.html#query=acom

Test URLs:
- After: https://mwpw-160085--mas--adobecom.hlx.live/studio.html


studio/libs/maslib.js
```diff
@@ -1,6 +1,6 @@
 const getMiloLibs = () => {
     const milolibs = new URLSearchParams(window.location.search)?.get('milolibs');
-    if (!milolibs) return 'https://www.adobe.com';
+    if (!milolibs) return 'https://ccd--milo--adobecom.hlx.live';
```
npeltier, line 3: can we rather add localeStorage for milolibs? and we leave real domain in code?

npeltier (changes requested):

studio/libs/maslib.js
yesil: When visitors come to https://main--mas--adobecom.hlx.live/studio.html, I would like it to work with the latest fixes without having to add `milolibs` params.
https://www.adobe.com is lagging behind and will be the same for sometime.
At someone we can revert this change, but I think it is too early for www.adobe

studio.html
```diff
@@ -6,13 +6,9 @@
         <meta name="robots" content="noindex, nofollow" />
         <meta name="nofollow-links" content="on" />
         <meta name="viewport" content="width=device-width, initial-scale=1" />
-        <meta name="aem-base-url"/>
-        <script>
-            let baseUrl = window.location.href.includes('localhost') ? 'http://localhost:8080' : 'https://mas-proxy-odin.corp.adobe.com';
```
3ch023, line 11: does it work on localhost as well without proxy? 
when we switch prod we'll still need this i think

studio/libs/maslib.js
3ch023: >can we rather add localeStorage for milolibs?
I would keep request parameter, it's so much easier to use.
you can share a link with someone or open 2 tabs with different branches

3ch023: and yes, it doesn't look like a good idea to me to point to a feature branch.
if you want to share the middle state with a visitor just add milolibs, but we should bound studio to main

npeltier: i would keep query param, just add local storage for devs like us tired of adding the parameter

3ch023 (approved):

npeltier (approved):

PR #71, opened 2024-10-16
mnemonic /multifield + fragment title/desc fields

yesil:
Authoring of multiple mnemonics  with icon, alt text and option link.
Also adds internal fragment title & description.

Resolves: 
[MWPW-160015](https://jira.corp.adobe.com/browse/MWPW-160015)
[MWPW-160079](https://jira.corp.adobe.com/browse/MWPW-160079)
[MWPW-160507](https://jira.corp.adobe.com/browse/MWPW-160507)

Screenshots:
image

image

**Test URLs:**
- After: https://mwpw-160015--mas--adobecom.hlx.live/studio.html?milolibs=mwpw-160015--milo--yesil


3ch023:
Hi @yesil , once small thing i noticed:
I have a card with no icon (there is such req)
when i open edit, it looks like i have to provide it. every time i need to remove it again
image

3ch023 (approved):
small nit about UI, but looks great otherwise!

npeltier (approved):

npeltier:
it's not directly obvious those 3 input belong to same field, some ideas, pick the ones you like :-) :
- on hover, change background & border color of the inputs wrapper,
- pad the 3 inputs a bit,
- on the left of inputs, put the selected icon,

npeltier:
also nit: shouldn't fields belong to editors?

yesil:
@npeltier I was also thinking to highlight the grouping of those fields. Will check what Spectrum proposes.

yesil:
> shouldn't fields belong to editors?
I think fields can be rather generic, as much as possible. And editors re-uses exiting fields to create the form for the content type which is only merch card for now.

Axelcureno:
Small suggestion: Make label fields in **strong** to indicate sub-fields belong to a particular field, like the case of Mnemonics which contain sub-fields Icon URL, alt text, link.
Also, should we also add icon size to list or properties? CCD Slice cards uses different size than Milo (30px vs 40px).

afmicka:
@yesil  as a follow-up from yesterday's DM discussion, i leave here the issues that were discussed:

1. When you enter the text for the fragment description field and click Save button, the text gets erased from the field. You need to close and reopen the card to see the text that was entered. And again if you change it and save, that text gets removed from the field.
Screenshot 2024-10-17 at 16 09 14
Screenshot 2024-10-17 at 16 09 24

2. if you enter the URL for icon without `https://`  it will make a relative path for the url (maybe to have validation?)
Screenshot 2024-10-17 at 16 11 16
Screenshot 2024-10-17 at 16 12 39

3. If you click on the icon on the card that has URL, nothing happens. It previews the link, changes the cursor indicating that it is a link, but does not take you to the page. Is it supposed to?

4.  In my tests card, I have a card created with 2 icons, one with the URL and one without it (since it is an optional field). If i change the order with drag and drop and click to Save, the order gets reversed back and when i reopen the card, it shows the old order. Attaching video to the jira (too big for this section)

yesil:
@Axelcureno 
>Make label fields in strong 

Since we are using spectrum field label and it doesn't provide an option to make it strong, I think we should stick to it as it must be following spectrum design system.

there is a size option but it is not implemented by SWC:

https://spectrum.adobe.com/page/field-label/

yesil:
@afmicka reported issues are addressed, thx.

afmicka:
thanks @yesil. I see them fixed now.

PR #78, opened 2024-10-29
MWPW-161117 consumer selection

npeltier:
- cards are displayed from top folder corresponding to shown tabs,
- only CCD is default, more can be displayed if added to  local storage,
- remove of useless breadcrumb,
- didn't fix table view padding as not sure we'll be using it yet
image

Test URLs:
- After: https://MWPW-161117--mas--adobecom.hlx.live/studio.html


3ch023 (changes requested):
both #path (opening a tab) and #query=nala (search) are using the hash, this seems to be the reason this PR is breaking search.
Maybe we switch to query parameters instead of hash?

npeltier:
> Maybe we switch to query parameters instead of hash?
thanks i'll have a look. Either way is good, catalog is using hash with different dimensions, but i agree query parameters looks better to me cc @yesil

npeltier:
> > Maybe we switch to query parameters instead of hash?
> 
> thanks i'll have a look. Either way is good, catalog is using hash with different dimensions, but i agree query parameters looks better to me cc @yesil

discussed it internally with @3ch023 & @yesil and we agreed to keep things this way as we have other stuff to do than reimplementing the wheel atm :)

3ch023:
https://mwpw-161117--mas--adobecom.hlx.live/studio.html#path=%2Fcontent%2Fdam%2Fmas%2Fnala shows no results

yesil:
there is too much white space above the tabs. 
image

I think we need to move filter panel buttons inside content-navigation and have only a single toolbar.

image

studio/src/aem/content-navigation.js
```diff
@@ -69,6 +72,35 @@ class ContentNavigation extends LitElement {
         if (!this.source) return;
         this.source.addEventListener(EVENT_LOAD, this.forceUpdate);
         this.source.addEventListener(EVENT_CHANGE, this.forceUpdate);
+        this.source.getTopFolders().then((folders) => {
+            this.addTabs(folders);
+        });
+    }
+
+    async addTabs(tabs) {
```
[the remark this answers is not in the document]
npeltier: "problem" with using lit-html is that it implies adding tabs to shadow DOM. I'm fine trying that, but we agreed offline to add them to light DOM

npeltier:
> I think we need to move filter panel buttons inside content-navigation and have only a single toolbar.

this would need longer discussion with PM/UX, i propose not to do anything on that topic in that PR

npeltier:
@3ch023 @yesil  please have a second look. I know styling is still not perfect, but as discussed we are still not sure about the overall styling itself.

npeltier:
image
this is easily achievable though moving `<mas-filter-toolbar></mas-filter-toolbar>` to content-navigation if you prefer (i don't have opinions)

3ch023:
seems like we agreed offline to do the dropdown instead of tabs

studio/src/aem/content-navigation.js
```diff
@@ -64,11 +72,63 @@ class ContentNavigation extends LitElement {
...
+        this.source.path = consumer;
+        this.source.openFolder(consumer);
+        this.source.searchFragments();
+    }
+
+    handleConsumerChange(event) {
+        this.selectConsumer(event.target.value);
+    }
+
+    get consumers() {
```
yesil, line 100: name suggestions: `get roots`

studio/test/aem/aem-fragments.test.js
yesil, line 1: 👍

yesil (approved):

studio/src/aem/content-navigation.js
yesil: I found consumer too opinionated, because, in my idea, everything under aem folder should remain generic and reusable.

studio/src/aem/aem-fragments.js
```diff
@@ -179,9 +178,13 @@ class AemFragments extends LitElement {
         const fragment = await this.#aem.sites.cf.fragments.getById(
             this.searchText,
         );
-        this.#searchResult = [fragment];
+        if (!fragment || fragment.path.indexOf(getDamPath(this.path)) < 0) {
+            this.#searchResult = [];
```
3ch023, line 182: `this.#searchResult = [];` already happened in line 173, no?

studio/src/aem/aem-fragments.js
```diff
@@ -179,9 +178,13 @@ class AemFragments extends LitElement {
...
+        } else {
+            this.#searchResult = [fragment];
+            this.addToCache([fragment]);
```
3ch023, line 185: this needs an await, maybe you could add it, otherwise it should come with Ilyases changes

studio/src/aem/aem-fragments.js
```diff
@@ -193,18 +196,22 @@ class AemFragments extends LitElement {
      * @param {string} search.variant - The variant to filter by.
      */
     async searchFragments() {
-        this.#search = {
-            query: this.searchText,
```
3ch023, line 197: this is needed for setting correct hash when you search by fragment id

studio/src/aem/content-navigation.js
npeltier: i can keep top folders

npeltier: but i guess i like the binding of top folders to consumers/surfaces: i agree that nala/docs/drafts surfaces are strange, but we could still consider them this way.
Essentially they are top folders with limited write access, and that is how we designed our consumer/surface storage

studio/src/aem/aem-fragments.js
npeltier: no, this is here in case some fragment is found but you are in wrong folder

3ch023 (approved):

studio/src/aem/mas-filter-toolbar.js
```diff
@@ -19,7 +20,7 @@ class MasFilterToolbar extends LitElement {
     `;
 
     static properties = {
-        searchText: { type: String, state: true },
+        searchText: { type: String, state: true, attribute: 'search-text' },
```
yesil, line 23: do we need it as an attribute?  As you user changes the search text, we would end up having an inconsistency with that is in the attribute and what is in the search. If we use `reflect:true`, that it will cause unnecessary DOM updates.

studio/src/aem/mas-filter-toolbar.js
```diff
@@ -29,6 +30,11 @@ class MasFilterToolbar extends LitElement {
         this.variant = 'all';
     }
 
+    connectedCallback() {
+        super.connectedCallback();
+        this.searchText = document.querySelector('mas-studio').searchText;
```
yesil, line 35: cannot this get the searchText value from deeplink?
also, I propose to drop the `mas-` prefix in the file names. It is already the MAS repository.

studio/src/aem/content-navigation.js
```diff
@@ -64,72 +77,131 @@ class ContentNavigation extends LitElement {
...
     async forceUpdate() {
         this.requestUpdate();
     }
 
     unregisterFromSource() {
-        this.source?.removeEventListener(EVENT_LOAD, this.forceUpdate);
+        if (this.deeplinkDisposer) {
+            this.deeplinkDisposer();
+        }
+        this.source?.removeEventListener(EVENT_LOAD, this.handlerSourceLoad());
```
yesil, line 122: ` this.source?.removeEventListener(EVENT_LOAD, this.handlerSourceLoad());` will not remove the listener.

I recommend to bind the handler in the constructor and then, add and remove it as follows:
```js
this.source?.addEventListener(EVENT_LOAD, this.handlerSourceLoad);
this.source?.removeEventListener(EVENT_LOAD, this.handlerSourceLoad);
```

yesil: so that the same instance is added and removed.

studio/src/aem/content-navigation.js
```diff
@@ -1,6 +1,10 @@
 import { css, html, LitElement, nothing } from 'lit';
 import { styleMap } from 'lit/directives/style-map.js';
 import { EVENT_CHANGE, EVENT_LOAD } from '../events.js';
+import { deeplink, pushState } from '../deeplink.js';
+import { getTopFolder } from './aem-fragments.js';
```
yesil, line 5: I think `getTopFolder` should be in a different file.
since `aem-fragments.js` doesn't belong to content-navigation, it should not be directly imported IMO.

studio/src/aem/mas-filter-toolbar.js
npeltier: this is used for init

studio/src/aem/mas-filter-toolbar.js
npeltier: i could, but this also triggers a event handler creation i'm not a fan of here

npeltier:
@yesil thanks for your comments, let's treat them as part as planned refactor

yesil:
@npeltier I think we should at least correct the even listeners before merging.

yesil (approved):

afmicka:
@npeltier the search does not work for me if i use `aem.env=qa` ... as soon as i press enter, the `aem.env` parameter gets removed from the url (that does not happen if you click Search button) 
1. go to https://mwpw-161117--mas--adobecom.hlx.live/studio.html?aem.env=qa
2. enter text to search
3. click enter

afmicka:
@npeltier it is also not performing search with partial word. If i search with the word 'photo' and not 'photoshop' i get zero results. Shouldn't i get all the cards that have photoshop in it? 
Screenshot 2024-11-06 at 16 57 22

afmicka:
@npeltier there is a regression in the preview on the cards. Any change in the RTE fields is not previewed on the card. Open any card, make a change in eny of the fields (title, badge, description), click outside the field and notice the change is not previewed.

Axelcureno (approved):

Axelcureno:
Looks neat 🚀

afmicka:
@npeltier as agreed, created jira with the made observation for future improvments https://jira.corp.adobe.com/browse/MWPW-161894

the rest looks good to me. Approving the PR

PR #79, opened 2024-10-30
MWPW-159374: Replace tinyMCE with Prosemirror

yesil:
Resolve:
https://jira.corp.adobe.com/browse/MWPW-159374
https://jira.corp.adobe.com/browse/MWPW-159258
https://jira.corp.adobe.com/browse/MWPW-159259

Milo PR: https://github.com/adobecom/milo/pull/3117
OST PR: https://git.corp.adobe.com/wcms/tacocat.js/pull/645

Test URLs:
- Before: https://main--mas--adobecom.hlx.live/studio.html?aem.env=qa#query=acom
- After: https://mwpw-159374--mas--adobecom.hlx.live/studio.html?aem.env=qa&milolibs=mwpw-159374--milo--adobecom#query=acom

PROD: https://mwpw-159374--mas--adobecom.hlx.live/studio.html?milolibs=mwpw-159374--milo--adobecom#query=acom


yesil:
I have some OST changes as well, I'll open a PR tomorrow.

3ch023:
seeing overlap 
image

studio/src/ost.js
```diff
@@ -138,6 +138,7 @@ export function getOffferSelectorTool() {
                 <div id="ost"></div>
             </sp-dialog-wrapper>
         </overlay-trigger>
```
3ch023, line 140: is `</overlay-trigger>` closing tag needed? can't see the opening one

studio/src/rte/ost.js
```diff
@@ -0,0 +1,234 @@
...
+            if (selectedText) return selectedText.id;
+            return ctaLabel || this.getDefaultText();
+        },
+    },
+};
+
+// Function to get the difference between two objects
+function getObjectDifference(values, defaults) {
+    const difference = {};
+
```
3ch023, line 63: extra spaces her and in line 71

studio/src/rte/rte-field.js
```diff
@@ -0,0 +1,746 @@
+import { LitElement, html, nothing, css } from 'lit';
```
npeltier, line 1: shouldn't this file go with ./fields folder?

studio/src/rte/rte-field.js
```diff
@@ -0,0 +1,746 @@
...
+
+    constructor() {
+        super();
+        this.readOnly = false;
+        this.isLinkSelected = false;
+        this.priceSelected = false;
+        this.showLinkEditor = false;
+        this.inline = false;
+        this.link = false;
+        this.#boundHandlers = {
```
npeltier, line 160: is it complicated to abstract plugins? i'm thinking of 2 things:
- complexity & ramp ups for future additions, where we want to add things like see terms or stuff like this, 
- maintenance of rte-field that gathers lot of logicwhere rte-field file will be heavily modified. If your answer is yes i'm fine leaving this like that.

npeltier (approved):
lgtm

studio/src/rte/rte-link-editor.js
```diff
@@ -0,0 +1,293 @@
...
+                detail: data,
+            }),
+        );
+        this.#handleClose();
+    }
+
+    #handleClose() {
+        this.open = false;
+        this.dispatchEvent(
+            new CustomEvent('close', { bubbles: false, composed: true }),
```
3ch023, line 288: even though this event doesn't bubble up it might be 'safer' to use more concrete even names? like 'link-editor-close'? in future if there is some refactoring and you'll search for 'close' events you will not get 5 diff kinds of them

3ch023 (approved):
lgtm, but a few comments.
+ need to switch focus to another field before RTE changes are picked up - as discussed in slack it's better to fix it early for GWP to have a good experience. but could be done also in a follow up pr, up to you

studio/src/rte/rte-link-editor.js
yesil: I try to avoid creating very specific event names, as it makes harder for a listener to listen all of them.
SWC for instance always uses `close`, `change` events, similar to vanilla DOM.
Sometimes it is handy to listen for `close` or `change` event on a container.

studio/src/ost.js
yesil: nope ,thx

yesil: @3ch023  thanks to you I found dead ost code to cleanup.

studio/src/rte/rte-field.js
yesil: I didn't want to have fields/rte.
I find it easier to access from src/rte.
can we discuss it later? :)

studio/src/rte/rte-field.js
yesil: @npeltier my understanding of Prosemirror is that, since it is highly customizable, you create your own RTE with it, but not necessarily a customizable RTE with it, as it would require you to re-expose many low level features of Prosemirror somehow.
The learning curve is IMO high and I wanted to have something that works for us in this first iteration.
I already have some ideas how to approach placeholders implementation, but who ever works on it, will need to spend some time to learn about Prosemirror.

afmicka:
@yesil I am not able to add the price to the slice wide card that has text in it. Maybe it is something with the card. I can add it in the first row, but not in the second. I also get the 'fragment could not be saved message'
Screenshot 2024-11-04 at 13 31 45

yesil:
@afmicka updates in the editor should be fixed now.

studio/src/rte/ost.js
```diff
@@ -0,0 +1,234 @@
...
+    workflow: 'UCv3',
+    workflowStep: 'email',
+};
+
+const OST_VALUE_MAPPING = {
+    true: true,
+    false: false,
+};
+
+export function onSelect(offerSelectorId, type, offer, options, promoOverride) {
```
Axelcureno, line 129: `offer` doesn't seem to be used, can we remove it?

studio/src/rte/rte-field.js
```diff
@@ -0,0 +1,779 @@
+import { LitElement, html, nothing, css } from 'lit';
+import { EditorState, NodeSelection, TextSelection } from 'prosemirror-state';
```
Axelcureno, line 2: `TextSelection` import is not being used. Can we remove it?

Axelcureno (approved):
A couple suggestions to keep the code clean but overall lgtm! 🚀

studio/src/rte/ost.js
yesil: @Axelcureno we cannot remove, as it is the signature of the callback expected on the OST side.

afmicka:
@yesil Leaving here the list of issues still present:
1. Double click on configured CTA does not open the OST
2. Strikethrough price on slice cards - not able to apply bold style and font colour should be lighter grey as on suggested card
Screenshot 2024-11-05 at 10 16 43

expected: 
Screenshot 2024-11-05 at 10 22 42

3. When 
Screenshot 2024-11-05 at 10 40 11
CTA variant is applied in studio, `data-extra-option` attribute without value is added, giving placeholder-failed

4. Issue #2 from the previous list

afmicka:
@yesil studio looks good to me now. Will create separate jira for the strikethrough price style

The remaining issues i see (but are no regression) are:
1. When you remove the text from the badge field, it does not get removed in the preview. Card needs to be saved and page refreshed in order to see the badge gone

2. When OST is opened for the checkout link, the window has a scroll bar, buttons at the bottom are not visible.

Let me know if these issues need a separate jiras or will be addressed here

afmicka:
Issue no 1 fixed, the second created a follow-up jira https://jira.corp.adobe.com/browse/MWPW-161588. 
Looks good to me

studio/src/studio.js
```diff
@@ -335,7 +332,7 @@ class MasStudio extends LitElement {
                 ? html`<mas-filter-panel></mas-filter-panel>`
                 : nothing}
             ${this.content} ${this.fragmentEditor} ${this.selectFragmentDialog}
-            ${this.toast} ${this.loadingIndicator} ${getOffferSelectorTool()}
+            ${this.toast} ${this.loadingIndicator}}
```
Axelcureno, line 335: There's an extra closing curly bracket at eol that should be removed.

yesil: Good catch, thanks @Axelcureno

PR #83, opened 2024-11-06
refresh published fragment

3ch023:

MWPW-161451

Test URLs:
- After: https://MWPW-161451--mas--adobecom.hlx.live/studio.html

### DIFF
--- a/studio/src/aem/aem-fragments.js
+++ b/studio/src/aem/aem-fragments.js
@@ -227,6 +227,13 @@ class AemFragments extends LitElement {
 
     async publishFragment() {
         await this.#aem.sites.cf.fragments.publish(this.fragment);
+        //this.fragment.status = 'PUBLISHED';
+        setTimeout(async () => {
+            const latest = await this.#aem.sites.cf.fragments.getById(
+                this.fragment.id,
+            );
+            aemFragmentCache.get(this.fragment.id)?.refreshFrom(latest);
+        }, 10000);
     }
 
     async deleteFragment() {


studio/src/aem/aem-fragments.js
```diff
@@ -227,6 +227,12 @@ class AemFragments extends LitElement {
 
     async publishFragment() {
         await this.#aem.sites.cf.fragments.publish(this.fragment);
+        setTimeout(async () => {
+            const latest = await this.#aem.sites.cf.fragments.getById(
+                this.fragment.id,
+            );
+            aemFragmentCache.get(this.fragment.id)?.refreshFrom(latest);
+        }, 4000);
```
npeltier, line 235: are we sure this will work? worst case we have stale cache?

npeltier (approved):

studio/src/aem/aem-fragments.js
yesil: I'm not a fun of arbitrary `setTimeout`. 
If we want to refresh status, let's support the official APIs: https://developer.adobe.com/experience-cloud/experience-manager-apis/api/stable/sites/

image

yesil (changes requested):

3ch023:
@yesil i am not sure i can setup webhooks stuff in the scope of this ticket, this pr was just a tiny UI fix, although i agree it is not 100% reliable..
imo we should have a separate ticket to look into listening to the created/published/unpublished events due to the effort of setting it all up (I went briefly through https://developer.adobe.com/events/docs/guides/)

in the meantime i can close this PR if we are sure that timeout is not something we want to have.
But I am not sure if its that bad, since getting the status updated after 4sec for a user is better than getting nothing.

PR #92, opened 2024-11-14
MWPW-161897: tagging support

yesil:
Introduces `aem-tag-picker-field` custom element.
Adds tags field in the merch card editor
Updates the filter panel to filter based on tags. 

Resolve: MWPW-161897

Test URLs:
- After: https://mwpw-161897--mas--adobecom.hlx.live/studio.html


studio/src/aem/aem.js
```diff
@@ -167,20 +176,70 @@ class AEM {
...
             throw new Error(
                 `Failed to save fragment: ${response.status} ${response.statusText}`,
             );
         }
-        return await this.getFragment(response);
+
+        await this.saveTags(fragment);
+
+        const newFragment = await this.sites.cf.fragments.getById(fragment.id);
+        return newFragment;
```
npeltier, line 201: is newFragment here for debugging?

studio/src/aem/aem-tag-picker-field.js
```diff
@@ -0,0 +1,293 @@
...
+    get tags() {
+        if (this.ready === false) return nothing;
+        if (this.tagsInHierarchy.length === 0) return nothing;
+        return html` ${repeat(
+            this.tagsInHierarchy,
+            (path) => path,
+            (path) =>
+                html`<sp-tag
+                    deletable
+                    @delete=${this.#deleteTag}
```
npeltier, line 234: do you think we can add `protected-namespaces` property that contains all subname spaces for which we should _not_ be able to remove tags from UI?

studio/src/aem/aem-tag-picker-field.js
```diff
@@ -0,0 +1,293 @@
...
+        this.hierarchicalTags = new Map();
+        this.value = [];
+        this.#aem = null;
+        this.ready = false;
+    }
+
+    connectedCallback() {
+        super.connectedCallback();
+        this.#aem = new AEM(this.bucket, this.baseUrl);
+        this.loadTags();
```
npeltier, line 105: await?

npeltier (approved):

package.json
```diff
@@ -48,7 +49,7 @@
         "@spectrum-web-components/split-button": "^0.47.2",
         "@spectrum-web-components/status-light": "^0.47.2",
         "@spectrum-web-components/table": "^0.47.2",
-        "@spectrum-web-components/tabs": "^0.47.2",
+        "@spectrum-web-components/tags": "^0.47.2",
```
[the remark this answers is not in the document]
yesil: oups, it wasn't intentional, but I guess we don't need tabs just yet.

studio/src/aem/aem-tag-picker-field.js
yesil: for a tag picker, we can restrict the exact tree, the user can add / remove tags from.
And here, only those tags will be shown.
To me it makes sense that the user can remove tags that he might have added.
We can have multiple tag pickers storing tags in the same `tags` field.

studio/src/aem/aem-tag-picker-field.js
yesil: connectedCallback is not async, cannot do await but, I don't need to await neither here.

3ch023 (approved):

afmicka:
@yesil Few observations. Let me know what is intended behaviour:
1. Filter pop-up does not close once the selection is made (the last hierarchy value) 
2. Filters are not preserved on page refresh (they are lost if page is reloaded)
3. If you perform combination of search and filter, it does not give correct result in case there is no a match and takes only the value from the search field. (Here i have searched an id card and filtered by creative all apps. This card does not have that tag, i would expect empty list with this combination) 
Screenshot 2024-11-19 at 15 03 40

yesil:
@afmicka this is expected, both contains links to **https://www.adobe.com/creativecloud/plans.html**

image

afmicka:
@yesil on smaller screens (Macbook 13"), if you scroll the edit form almost to the bottom and open the tag menu, the content gets out of the popup. 
Screenshot 2024-11-20 at 10 10 09

yesil:
@afmicka the item overflow on small screen is fixed, checking CTAs.

afmicka:
@yesil @3ch023 @npeltier @Roycethan @Axelcureno 
I keep seeing the same issues with CTA and price edits mentioned above. Given that no relation to the code changes could be identified and the PR is blocking further work, have created separate jira for tracking if this start happening on main as well after merging this. https://jira.corp.adobe.com/browse/MWPW-162870 

If this is ok with everyone, please feel free to merge this PR to unblock.

PR #95, opened 2024-11-15
MWPW-161845: add link analytics id's

3ch023:
Update Link Dialog with a dropdown of Link analytics Ids.
For Checkout CTAs id's can be derived from the CTA text selected in OST.
For rest of links there is a list of valid values in dropdown.

The first 3 cards in the gallery already have the id's set

Resolve: https://jira.corp.adobe.com/browse/MWPW-161845

Test URLs:
- After: https://mwpw-161845--mas--adobecom.hlx.live/studio.html#path=drafts

Gallery: 
https://mwpw-161845--mas--adobecom.hlx.live/gallery/ccd.html?milolibs=mwpw-161845--milo--adobecom

### DIFF
--- a/studio/src/constants.js
+++ b/studio/src/constants.js
@@ -1,0 +1,22 @@
+export const CHECKOUT_CTA_TEXTS = {
+    'buy-now': 'Buy now',
+    'free-trial': 'Free trial',
+    'start-free-trial': 'Start free trial',
+    'save-now': 'Save now',
+    'get-started': 'Get started',
+    'choose-a-plan': 'Choose a plan',
+    'learn-more': 'Learn more',
+    'change-plan-team-plans': 'Change Plan Team Plans',
+    upgrade: 'Upgrade',
+    'change-plan-team-payment': 'Change Plan Team Payment',
+    'take-the-quiz': 'Take the quiz',
+    'see-more': 'See more',
+    'upgrade-now': 'Upgrade now',
+    'get-offer': 'Get offer',
+};
+
+export const ANALYTICS_LINK_IDS = [
+    'learn-more',
+    'see-terms',
+    'what-is-included',
+];
--- a/studio/src/rte/rte-link-editor.js
+++ b/studio/src/rte/rte-link-editor.js
@@ -1,5 +1,6 @@
 import { LitElement, css, html, nothing } from 'lit';
 import { classMap } from 'lit/directives/class-map.js';
+import { CHECKOUT_CTA_TEXTS, ANALYTICS_LINK_IDS } from '../constants.js';
 
 export class RteLinkEditor extends LitElement {
     static properties = {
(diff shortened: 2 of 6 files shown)


studio/src/constants.js
```diff
@@ -0,0 +1,22 @@
+export const CHECKOUT_CTA_TEXTS = {
```
yesil, line 1: how about getting these values from tagging, and using it in both OST and rte field?

studio/src/rte/rte-link-editor.js
```diff
@@ -118,6 +126,34 @@ export class RteLinkEditor extends LitElement {
         return this.shadowRoot.querySelector('#linkHref');
     }
 
+    get #checkoutLink() {
```
yesil, line 129: since this is a boolean, can we rename as `#isCheckoutLink `

studio/src/rte/rte-link-editor.js
```diff
@@ -90,7 +98,7 @@ export class RteLinkEditor extends LitElement {
             >
             <sp-textfield
                 id="checkoutParameters"
-                placeholder="Exrta checkout parameters: e.g: promoid=12345&mv=1"
+                placeholder="Extra checkout parameters: e.g: promoid=12345&mv=1"
```
yesil, line 101: 😃

studio/src/constants.js
npeltier: may be it can still wait a bit here, no ?

npeltier (approved):
as discussed yesterday in the call, for now dropdown is good and needed, but we may want to remove it somehow later

studio/src/constants.js
3ch023: yeah, i'd like to discuss that separately before we do. having a tag for each possible buy now link text might be overwhelming, we need to think it through in peace (but CCD needs some labels now)

studio/src/rte/rte-link-editor.js
3ch023: will do, thx

Axelcureno (approved):

studio/src/constants.js
yesil: 👍

studio/src/rte/rte-link-editor.js
yesil: any update?

yesil (approved):
I prefer to rename `#checkoutLink` to `#isCheckoutLink` but not a blocker

studio/src/rte/rte-link-editor.js
3ch023: sorry, renamed now :)

3ch023:
> @3ch023 you have removed the spacing between cards (though this might be related to the milo PR). https://mwpw-161845--mas--adobecom.hlx.live/gallery/ccd.html?milolibs=mwpw-161845--milo--adobecom Screenshot 2024-11-19 at 13 50 39

confirming that it must be in milo PR, i will fix that

PR #97, opened 2024-11-18
MWPW-159191: freyja support

yesil:
Test URLs:
- After: https://mwpw-159191--mas--adobecom.hlx.live/studio.html?milolibs=mwpw-159191--milo--adobecom


studio/src/aem/render-view.js
```diff
@@ -53,7 +53,11 @@ class RenderView extends LitElement {
                 @mouseleave="${this.handleMouseLeave}"
                 @dblclick="${(e) => this.handleDoubleClick(e, fragment)}"
             >
-                <aem-fragment fragment="${fragment.id}" ims></aem-fragment>
+                <aem-fragment
+                    fragment="${fragment.id}"
+                    ims
```
3ch023, line 58: if i remember correct we wanted to remove 'ims' here? for another ticket but i though we could already remove it now

yesil: it was me who mistakenly said it. `ims` here to tell to use access token from IMS for the fetch call. Needed in authoring but not in publish.  If studio works correctly and fragment is retrieved via aem.js all the time, this attribute might indeed not be needed but for now I'm keeping it. I can remove it in a future PR.

3ch023: ahh okie okie

PR #101, opened 2024-11-20
MWPW-162805 initial load tests scripts 

npeltier:
- one for odin, one for wcs,
- rotation on ids,
- readme

in case one macbook is not enough as an injector, this has been chosen as it seems we could use an ethos infra to start them

Screenshot 2024-11-20 at 18 32 57
Screenshot 2024-11-20 at 18 21 34


yesil (approved):
🤌

npeltier:
cc @afmicka @Roycethan no need for QA here (load test scripts)

PR #102, opened 2024-11-21
MWPW-162740: Fix unresponsive CTA RTE field

Axelcureno:
Fixes unresponsive changes in RTE field when editing a link/cta in footer.

Resolve [MWPW-162740](https://jira.corp.adobe.com/browse/MWPW-162740)

Test URLs:
- After: https://MWPW-162740--mas--adobecom.hlx.live/studio.html


studio/src/rte/rte-field.js
```diff
@@ -232,11 +287,21 @@ class RteField extends LitElement {
                     'data-perpetual': { default: null },
                     'data-promotion-code': { default: null },
                     'data-wcs-osi': { default: null },
+                    'data-template': { default: null },
                     title: { default: null },
                     target: { default: null },
                     'data-analytics-id': { default: null },
+                    text: { default: '' },
```
yesil, line 294: @Axelcureno in my version, I didn't need to add the text attribute. Since it is already it is a child node of type text. I would keep it this way.

studio/src/rte/rte-field.js
```diff
@@ -23,6 +23,61 @@ const isNodeCheckoutLink = (node) => {
     return node.type.name === 'link' && node.attrs['data-wcs-osi'] !== null;
 };
 
+class LinkNodeView {
```
yesil, line 26: This is an interesting concept and can be powerful but I'm not sure if we need for links.

Axelcureno: The reason I added this is because we are using `atom: true` in our link nodes, and that attribute makes the node an indivisible unit meaning it does not have a built-in way to update the DOM. With this class we are overriding that behavior to update the DOM of the nodes (in this case update the `textContent` of the selection).

studio/src/rte/rte-field.js
Axelcureno: True, I have removed text attribute.

studio/src/rte/rte-field.js
Axelcureno: https://discuss.prosemirror.net/t/nodespec-with-atom-true-and-content-inline-and-focus/6040

yesil (approved):

npeltier (approved):
looks good to me, i would even consider putting it in a separate file, as rte-field gets bigger & bigger

PR #100, opened 2024-11-20
MWPW-162614: Normalize UX for field updates

yesil:
Refactored the editor panel in own custom element
added unit tests.

Resolve: 
* https://jira.corp.adobe.com/browse/MWPW-162614
* https://jira.corp.adobe.com/browse/MWPW-161187

Test URLs:
- Before: https://main--mas--adobecom.hlx.page/studio.html
- After: https://mwpw-162614--mas--adobecom.hlx.page/studio.html

PR #103, opened 2024-11-22
New state management

st-angelo-adobe:
Resolves: [MWPW-162507](https://jira.corp.adobe.com/browse/MWPW-162507)

Test URLs:
- After: https://MWPW-162507--mas--adobecom.hlx.live/studio.html


npeltier (commented):
looks like this would need to wait for #100

PR #105, opened 2024-12-03
MWPW-163214: M@S Studio Splash Screen

Axelcureno:
Adds splash screen experience to M@S Studio.

Resolve [MWPW-163214](https://jira.corp.adobe.com/browse/MWPW-163214)

Test URLs:
- Before: https://main--mas--adobecom.hlx.page/studio.html
- After: https://MWPW-163214--mas--adobecom.hlx.page/studio.html


studio/src/reactivity/store-controller.js
```diff
@@ -0,0 +1,34 @@
...
+    value;
+
+    /**
+     *
+     * @param {LitElement} host
+     * @param {ReactiveStore} store
+     */
+    constructor(host, store) {
+        this.store = store;
+        this.value = store.get();
```
yesil, line 17: why the initial value is get from store here and later at the line 31, the value in the store is not updated?

studio/src/mas-hash-manager.js
```diff
@@ -0,0 +1,62 @@
+import { LitElement, nothing } from 'lit';
+import StoreController from './reactivity/store-controller.js';
+import Store from './store.js';
+import { setHashParams } from './utils.js';
+import MasSearch from './entities/search.js';
+import MasFilters from './entities/filters.js';
+
+class MasHashManager extends LitElement {
```
yesil, line 8: why this need to be a web component?
why deeplink.js not sufficient?

studio/src/reactivity/store-controller.js
st-angelo-adobe: In the controller I simply initialize the local state (`this.value`). At line 31, the value in the story is already updated, `updateValue` reacts to that, updating the local state.

studio/src/mas-hash-manager.js
st-angelo-adobe: Hm, actually looking over it, it doesn't have to be a component. It can just be a simple function.
Deeplink does what store subscription does, so no point to have both and make things confusing.

In a simple function we can subscribe to the `search` and `filters` stores, and to the page hash (`hashchange`), and make them update each other.

I would prefer to do this separately, after merge if that's ok.

yesil: To update the state, we can either update hash, react to hash changes and update the store with these values
or update the store and the hash but not react to hash changes, after the first initialization.
both are possible with deeplink.js.

st-angelo-adobe: We can have something like:

image
And make sure they don't enter an infinite update loop.

Maybe we would even want to merge the search and filters stores, to be honest.

My point is, we don't really need the `deeplink.js` anymore, which is supposed to be used in `connectedCallback` and has a disposer, when we can just call a function once that subscribes to stuff, and that's it. The state is outside the components now.

studio/src/mas-fragment.js
```diff
@@ -0,0 +1,89 @@
...
+        store: { type: Object, attribute: false },
+        view: { type: String, attribute: true }, // 'render' | 'table'
+    };
+
+    createRenderRoot() {
+        return this;
+    }
+
+    selecting = new StoreController(this, Store.selecting);
+    selection = new StoreController(this, Store.selection);
```
3ch023, line 22: dupe

studio/src/mas-hash-manager.js
3ch023: agree it could be a function, but yeah, could be done in next PR

studio/src/mas-fragment.js
st-angelo-adobe: It's not, I'm subscribing to 2 different stores there 😄 One that specifies whether we're in selection mode, one that has the actual selection (an array of fragment ids)
To be honest, it might be better to set selection at fragment level (and I noticed on the main branch that there was something like that), but again, this is something we can come back to.

3ch023 (approved):

studio/src/mas-fragment.js
npeltier: that confusion is talkative: we should may be find another name that is less confusing for `selecting`

studio/src/editors/merch-card-editor.js
```diff
@@ -181,33 +302,271 @@ class MerchCardEditor extends LitElement {
                 value="${this.fragment.tags.map((tag) => tag.id).join(',')}"
                 @change=${this.#handeTagsChange}
             ></aem-tag-picker-field>
-        `;
+            <p>Fragment details (not shown on the card)</p>
```
yesil, line 305: This is specific to merch cards but common to any fragment, hence it was put in `editor-panel.js`.
Please move any form logic that is not specific to merch-card, back to editor-panel.js

studio/src/editors/merch-card-editor.js
```diff
@@ -181,33 +302,271 @@ class MerchCardEditor extends LitElement {
...
+            Events.showToast.emit({
+                variant: 'negative',
+                content: 'Failed to copy code to clipboard',
+            });
+        }
+    }
+
+    get toolbar() {
+        return html`<div id="editor-toolbar">
+            <sp-action-group
```
yesil, line 408: The action buttons should not be in merch-card-editor, please bring back editor-panel. :-)

studio/src/editors/merch-card-editor.js
```diff
@@ -181,33 +302,271 @@ class MerchCardEditor extends LitElement {
...
+                aria-label="Fragment actions"
+                role="group"
+                size="l"
+                compact
+                emphasized
+                quiet
+            >
+                <sp-action-button
+                    label="Move left"
```
yesil, line 417: I like this 👍

yesil (dismissed):

npeltier (changes requested):
Hi @st-angelo-adobe 

i tested your change and thanks a lot for this and all the efforts! However i personally  think we need another round of polish before getting it in. I understand you developed from quite a wild initial state and this is painful, but i do think your PR should be at a stage where we can - quickly - then iterate to something good. Typically i don't like current file structure, but i really don't consider blocking. I'll try to illustrate what i dislike and seems structural to me with following fictional scenario: let's imagine mas studio toolbar has no search, but already in another place stuff that do search. so related store exists.
As a UI developer, i should probably only know about some hook to central state manager about 'search' that i can update with my UI element, sending there keys & values. Ideally because we know it's lit, we could even have some lit hooks that takes fields name & directly update the stores.

here i have to add to declare a controller
`search = new StoreController(this, Store.search);`

that i use in 
```    get searchResultsLabel() {
        if (this.loading.value || !this.search.value.query) return nothing;
        return html`<span id="search-results-label"
            >Search results for "${this.search.value.query}"</span
        >`;
    }
```

that i *don't* use in 

```<sp-search
                placeholder="Search"
                @change="${extractValue(this.updateQuery)}"
                @submit="${preventDefault(extractValue(this.updateQuery))}"
                value=${this.search.value.query}
                size="m"
            ></sp-search>
```

that is using 
```    updateQuery(value) {
        Store.search.update((prev) => ({ ...prev, query: value }));
    }
```
and to me that's the problem. I guess i'm fine declaring a controller, but:
- i don't understand why we have "value" and "query" to handle. to me, there is a superflous level of structure here,
- i don't understand why i do have to declare a change event handler for something we do know how it works

here i would see something like
```   if (this.loading.value || !this.search.value) return nothing;
        return html`<span id="search-results-label"
            >Search results for "${this.search.value}"</span
        >`;
    }```

and just

```
```<sp-search
                placeholder="Search"
                @change="${this.search.changeHandler}"
                @submit="${preventDefault(extractValue(this.updateQuery))}"
                value=${this.search.value.query}
                size="m"
            ></sp-search>
```

npeltier:
having written it, i guess we could make those changes afterward. Will update to comment

npeltier (approved):

Roycethan:
@yesil  @Axelcureno here are feedbacks w.r.t ticket AC's and issues:

1) Status( Edits/Publish) are not updated at https://mwpw-163214--mas--adobecom.hlx.page/studio.html at first instance when user makes edits/publish in their content space and click back on "Home" sidenav option - for me i needed a browser refresh to see the recent update

2) 'Support' menu item should be at the bottom with Slack icon that takes users to '#merch-at-scale' Slack channel when clicked
Current:
image

Expectation:
image

3)Welcome message - i believe should  display  user's name fetched from IMS profile - currnetly its hardcoded
image

studio/src/editors/merch-card-editor.js
Axelcureno: Done. Added `FragmentEditorBase` class to reuse common methods between `merch-card-editor.js` and `editor-panel.js`

studio/src/editors/merch-card-editor.js
Axelcureno: Moved

studio/src/fragment-editor-base.js
```diff
@@ -0,0 +1,136 @@
...
+
+    disconnectedCallback() {
+        super.disconnectedCallback();
+        document.removeEventListener('keydown', this.handleKeyDown);
+    }
+
+    /**
+     * @returns {boolean} Whether or not the editor was closed
+     */
+    close() {
```
yesil, line 51: the `close` method is duplicated.

studio/src/editors/merch-card-editor.js
```diff
@@ -50,76 +62,81 @@ class MerchCardEditor extends LitElement {
...
         if (this.fragment.model.path !== MODEL_PATH) return nothing;
+
         const form = Object.fromEntries([
             ...this.fragment.fields.map((f) => [f.name, f]),
         ]);
-        return html`
-            <p>${this.fragment.path}</p>
+
+        return html`<div id="editor">
+            <editor-panel
```
yesil, line 80: this should be the other way around, editor-panel contains merch-card-editor.
also,`editor-panel` and  fragment-editor-base.js should be combined together.

e.g: https://github.com/adobecom/mas/blob/main/studio/src/editor-panel.js#L279

studio/src/mas-toolbar.js
```diff
@@ -0,0 +1,206 @@
...
+                placeholder="Search"
+                @change="${extractValue(this.updateQuery)}"
+                @submit="${preventDefault(extractValue(this.updateQuery))}"
+                value=${this.search.value.query}
+                size="m"
+            ></sp-search>
+        </div>`;
+    }
+
+    get writeActions() {
```
yesil, line 153: I find the name confusing.

studio/src/mas-toolbar.js
```diff
@@ -0,0 +1,206 @@
...
+    handleRenderModeChange(ev) {
+        localStorage.setItem('mas-render-mode', ev.target.value);
+        Store.renderMode.set(ev.target.value);
+    }
+
+    updateQuery(value) {
+        Store.search.update((prev) => ({ ...prev, query: value }));
+    }
+
+    get readActions() {
```
yesil, line 127: I find the name confusing.

studio/test/utils.js
```diff
@@ -152,3 +152,57 @@ export function initElementFromTemplate(templateId, title) {
     }
     return root;
 }
+
+export const waitForElement = (
```
yesil, line 156: this doesn't seem to be used.

studio/test/merch-card-editor.test.html
yesil, line 1: please revert this test back to editor-panel, this is testing mainly the generic editor-panel related features.

yesil (changes requested):
editor-panel changes are not correctly reverted yet.

studio/src/fragment-editor-base.js
Axelcureno: removed

Axelcureno: removed

studio/src/mas-toolbar.js
Axelcureno: renamed to something more descriptive

studio/src/mas-toolbar.js
Axelcureno: renamed to something more descriptive

studio/test/utils.js
Axelcureno: removed

studio/test/merch-card-editor.test.html
Axelcureno: done

studio/src/editors/merch-card-editor.js
```diff
@@ -1,37 +1,37 @@
...
 import '../fields/multifield.js';
 import '../fields/mnemonic-field.js';
+import '../aem/aem-tag-picker-field.js';
+import './variant-picker.js';
+import EditorPanel from '../editor-panel.js';
 
 const MODEL_PATH = '/conf/mas/settings/dam/cfm/models/card';
 
-class MerchCardEditor extends LitElement {
+class MerchCardEditor extends EditorPanel {
```
yesil, line 11: I don't see why extending from EditorPanel is needed.

I think this also resulted in the following being displayed at the top of the editor instead of the bottom.
image

yesil: editor-panel renders merch-card-editor, so the latter should not extends the former.

yesil (changes requested):

studio/src/editors/merch-card-editor.js
Axelcureno: Removed class inheritance

yesil (approved):

Roycethan:
> @yesil @Axelcureno here are feedbacks w.r.t ticket AC's and issues:
> 
> 1. Status( Edits/Publish) are not updated at https://mwpw-163214--mas--adobecom.hlx.page/studio.html at first instance when user makes edits/publish in their content space and click back on "Home" sidenav option - for me i needed a browser refresh to see the recent update
> 2. 'Support' menu item should be at the bottom with Slack icon that takes users to '#merch-at-scale' Slack channel when clicked
>    Current:
> 
> image
> Expectation: image
> 
> 3)Welcome message - i believe should display user's name fetched from IMS profile - currnetly its hardcoded image
> 
> 4. I was not able to copy to new card/delete after selecting it (disabled select options) , am i lack any permissions or something else ?
> 
> image
> cc @npeltier @3ch023

#1,2,3 are fixed. Since slack ison is not available @Axelcureno left default, i'll ask that to Nick in the ticket

PR #108, opened 2024-12-06
MWPW-158966: add healthcheck action

3ch023:
I/O runtime action to test wcs & odin endpoints returning valid json.
https://14257-mashealthcheck.adobeioruntime.net/api/v1/web/MasHealthCheck/health-check

Resolve MWPW-158966

Test URLs:
- After: https://MWPW-158966--mas--adobecom.hlx.live/studio.html


io-actions/app.config.yaml
```diff
@@ -0,0 +1,20 @@
+application:
+  actions: actions
+  web: web-src
```
yesil, line 3: this doesn't exist

yesil (approved):
Looks nice 👍

PR #113, opened 2024-12-11
State management pre cleanup

st-angelo-adobe:
Resolves: [MWPW-162507](https://jira.corp.adobe.com/browse/MWPW-162507), [MWPW-163155](https://jira.corp.adobe.com/browse/MWPW-163155)

Test URLs:
- After: https://mwpw-163155--mas--adobecom.hlx.live/studio.html


studio/src/aem/aem.js
```diff
@@ -248,6 +254,16 @@ class AEM {
         }
     }
 
+    async pollUpdatedFragment(oldFragment) {
```
yesil, line 257: 👍

yesil: I wonder if there would be a case where etag will not change because nothing really has changed. In this case It may block the UI indefinitely. Should we limit to first 10 attempt. and wait 500ms between every attempt.
Values can be changed as you see fit.

studio/src/mas-content.js
```diff
@@ -0,0 +1,119 @@
...
+    createRenderRoot() {
+        return this;
+    }
+
+    constructor() {
+        super();
+        this.goToFragment = this.goToFragment.bind(this);
+    }
+
+    loading = new StoreController(this, Store.fragments.loading);
```
yesil, line 18: having a controller per responsive property looks overkill to me. I find it too verbose and as the codebase grows, it can become hard to wrap the head around.

studio/src/reactivity/reactiveStore.js
yesil, line 1: naming should be in line with the result of the files

studio/src/reactivity/reactiveStore.js
```diff
@@ -0,0 +1,87 @@
...
+            subscriber(this.value);
+        }
+    }
+}
+
+export function reactiveStore(initialValue) {
+    return new ReactiveStore(initialValue);
+}
+
+export class FragmentStore extends ReactiveStore {
```
yesil, line 47: maybe this can be in own file?

3ch023:
looks good to me, except search by fragmentId is not working anymore:
https://mwpw-163155--mas--adobecom.hlx.live/studio.html#query=0ef2a804-e788-4959-abb8-b4d96a18b0ef&path=nala

also, if you pull latests from main the unit tests should be green again

studio/src/entities/filters.js
```diff
@@ -0,0 +1,9 @@
+export default class MasFilters {
```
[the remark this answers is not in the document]
st-angelo-adobe: Right now there is no purpose, because filters are not implemented. But when they are, this can be a place where everything about them can be handled, like `MasSearch` below.

studio/src/reactivity/masEvent.js
```diff
@@ -0,0 +1,19 @@
+export default class MasEvent {
```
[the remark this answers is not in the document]
st-angelo-adobe: It's the same exact concept as reactive stores, but no state. Just a one shot event that you can fire from anywhere and catch anywhere. 
It's basically `dispatch` a `CustomEvent`, but I guess without the downsides of needing an actual element to dispatch, and thinking about bubbling and stuff like that.

PR #116, opened 2024-12-13
MWPW-160756: Setup Nala tests for studio

afmicka:
Adding setup for Nala tests, github workflow to run on PRs and the very first test

Resolve https://jira.corp.adobe.com/browse/MWPW-160756

Test URLs:

Before: https://main--mas--adobecom.aem.live/studio.html
After: https://MWPW-160756--mas--adobecom.aem.live/studio.html


afmicka:
@3ch023 @npeltier @yesil 
The setup here will trigger the tests on every PR. At the moment, logging-in is not happening on github actions (unknown reasons) and the tests would be failing if this PR gets merged.

If [MWPW-164257](https://jira.corp.adobe.com/browse/MWPW-164257) gets implemented in the meantime, you can apply changes from the branch on my fork: [afmicka:MWPW-160756-token](https://github.com/afmicka/mas/tree/MWPW-160756-token) that is ready for that implementation (no changes needed assuming the implementation is the same as for the OST in milo).

Otherwise, you can disable the run on PRs (remove workflow file from this PR), merge this as is and be able to run tests locally (that works) and keep adding new ones.

playwright.config.js
```diff
@@ -0,0 +1,71 @@
...
+  projects: [
+    {
+      name: 'mas-live-chromium',
+      use: { ...devices['Desktop Chrome'] },
+      bypassCSP: true,
+      launchOptions: { args: ['--disable-web-security', '--disable-gpu'] },
+    },
+
+    {
+      name: 'mas-live-firefox',
```
3ch023, line 56: if we don't plan to run studio tests on FF and webkit in nearest future (which i assume we shouldn't) lets just remove them - no need to keep extra code if its unused

nala/utils/pr.run.sh
```diff
@@ -0,0 +1,84 @@
...
+
+# Navigate to the GitHub Action path and install dependencies
+cd "$GITHUB_ACTION_PATH" || exit
+npm ci
+npx playwright install --with-deps
+
+# Run Playwright tests on the specific projects using root-level playwright.config.js
+# This will be changed later
+echo "*** Running tests on specific projects ***"
+npx playwright test --config=./playwright.config.js ${TAGS} ${EXCLUDE_TAGS} --project=mas-live-chromium --project=mas-live-firefox --project=mas-live-webkit ${REPORTER} || EXIT_STATUS=$?
```
3ch023, line 76: i would remove `--project=mas-live-firefox --project=mas-live-webkit ` since we skip them later in test anyway.
nala installs infra each time for them, it makes tests longer without any gain

nala/studio/studio.test.js
```diff
@@ -0,0 +1,114 @@
...
+import ims from '../libs/imslogin.js';
+
+const { features } = StudioSpec;
+const miloLibs = process.env.MILO_LIBS || '';
+
+let studio;
+
+test.beforeEach(async ({ page, browserName, baseURL }) => {
+    test.slow();
+    test.skip(
```
3ch023, line 13: this can be removed if we adjust npx command to only use chromium project

3ch023 (changes requested):
minor comments/suggestion
+ test account needs to be verified

nala/utils/nala.run.js
```diff
@@ -0,0 +1,206 @@
+#!/usr/bin/env node
```
3ch023, line 1: we don't have to change it in this PR but i find having both `pr.run.sh` and `nala.run.js` doing similar thing confusing, are we sure we need them both? or could be 1 script?

afmicka:
putting back the "don not merge" label until the regressions in the new studio are resolved.

afmicka:
@3ch023 removed the "do not merge" label as the tests are adjusted for the new studio (including regressions, marked the sections to be removed once regressions are fixed). 

there are 3 tests running and 1 being skipped completely due to the studio regressions

3ch023 (approved):

PR #117, opened 2024-12-16
MWPW-164093: Adobe Home 'Pricing widget' merch card

Axelcureno:
This PR introduces Adobe Home 'Pricing Widget' merch card variant in M@S Studio.

Milo PR: https://github.com/adobecom/milo/pull/3384

Resolves: https://jira.corp.adobe.com/browse/MWPW-164093

Test URLs:
- Before: https://main--mas--adobecom.hlx.page/studio.html?milolibs=MWPW-164093--milo--adobecom#path=adobe-home
- After: https://mwpw-164093--mas--adobecom.hlx.page/studio.html?milolibs=MWPW-164093--milo--adobecom#path=adobe-home


studio/src/rte/rte-field.js
```diff
@@ -141,6 +142,11 @@ class RteField extends LitElement {
                     color: var(--spectrum-global-color-gray-50);
                 }
 
+                a.primary {
```
yesil, line 145: why this change is needed?

studio/src/rte/rte-link-editor.js
```diff
@@ -170,6 +170,14 @@ export class RteLinkEditor extends LitElement {
                     variant="accent"
                     >Accent</sp-button
                 >
+                <sp-button
```
yesil, line 173: what is the difference between primary or accent? Cannot you use accent ?

studio/src/rte/rte-field.js
Axelcureno: Primary button is not added as an option in RTE (only primary outline) even though we can render with spectrum.css
Screenshot 2025-01-27 at 12 11 37 PM

Current options:
Screenshot 2025-01-27 at 12 11 49 PM

studio/src/rte/rte-link-editor.js
Axelcureno: Primary is black by default whereas accent is blue.

yesil (approved):

PR #118, opened 2024-12-17
MWPW-162452 OST promocode fix

3ch023:
Resolve https://jira.corp.adobe.com/browse/MWPW-162452

Fix OST opening without promo override value.
Remove Gallery from mas - we should use the Milo one.
move `mas-commerce-service` into a Studio.
add commerce env switch (needed to test promocode in action)

For OST, we don't support WCS Stage env, but to allow GWP author DRAFT offers, I check if switch is Stage commerce - and open ost with landscape=DRAFT. This for now will do..

Test here:
https://mwpw-162452--mas--adobecom.hlx.live/studio.html#path=drafts&commerce.env=stage&query=e4fefecb-fba8-4e88-a352-89314ec01df4 
(Promo code: mariiapromo)

Test URLs:
- After: https://MWPW-162452--mas--adobecom.hlx.live/studio.html


yesil (approved):

studio/src/store.js
```diff
@@ -4,6 +4,9 @@ import MasFilters from './entities/filters.js';
 import MasSearch from './entities/search.js';
 import { reactiveStore } from './reactivity/reactive-store.js';
 
+const params = Object.fromEntries(
+    new URLSearchParams(window.location.hash.slice(1)),
```
st-angelo-adobe, line 8: Maybe we could have a hash util, so that anywhere we want a param we can just access it like a normal object. Not saying it's needed here, since I'll work on the hash manager task I can think of something.

studio/src/studio.js
```diff
@@ -13,6 +13,7 @@ import './mas-hash-manager.js';
 import './mas-splash-screen.js';
 import StoreController from './reactivity/store-controller.js';
 import Store from './store.js';
+import Events from './events.js';
```
st-angelo-adobe, line 16: This is unused

st-angelo-adobe (changes requested):

studio/src/studio.js
```diff
@@ -58,7 +77,14 @@ class MasStudio extends LitElement {
 
     render() {
         return html`
-            <mas-top-nav env="${this.env}"></mas-top-nav>
+            ${this.commerceEnv?.value == 'stage'
```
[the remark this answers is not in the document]
3ch023: >Also I think the commerceEnv store, and stores in general should have correct values, and be validated. Right now you're setting the value directly from the param, and that could be anything. I see prod is the default, so when setting the commerceEnv store value, you can just validate there. It's better to always trust the store imo.

good point, addressing

>And then you can just write <mas-commerce-service env=${this.commerceEnv.value}></mas-commerce-service>

Unfortunately in `mas-commerce-service` env attribute is not observable so its not enough to toggle its value. to trigger commerce env swithc i need to readd to dom the whole `mas-commerce-service`. but let me try to fix it up a bit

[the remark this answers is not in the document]
3ch023: to add a bit more explanation if i remove if condition and just pass by the attribute value like this
<mas-commerce-service env=${this.commerceEnv.value}></mas-commerce-service>

lit will not recreate mas-commerce-service element in dom, it will only update the attribute. and i need to trigger element recreation

studio/src/studio.js
3ch023: addressed

studio/src/store.js
3ch023: agreed :)
yeah it fealt not good to me as well when i was adding it

studio/src/studio.js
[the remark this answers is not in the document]
st-angelo-adobe: Ah, I see. That's ok then. 👍

st-angelo-adobe (approved):

yesil (approved):

PR #123, opened 2024-12-24
MWPW-161354: filter unused fields in the editor

yesil:
Resolve:
- https://jira.corp.adobe.com/browse/MWPW-161354
- https://jira.corp.adobe.com/browse/MWPW-162449
- https://jira.corp.adobe.com/browse/MWPW-166250

Test URLs:
- After: https://mwpw-161354--mas--adobecom.hlx.live/studio.html?milolibs=stage--milo--adobecom


3ch023 (approved):

npeltier (approved):

studio/src/editors/merch-card-editor.js
```diff
@@ -167,6 +171,28 @@ class MerchCardEditor extends LitElement {
...
+        });
+        const variant = merchCardCustomElement.getFragmentMapping(
+            this.fragment.variant,
+        );
+        if (!variant) return;
+        Object.keys(variant).forEach((key) => {
+            const field = this.querySelector(`sp-field-group.toggle#${key}`);
+            if (field) field.style.display = 'block';
+        });
+        this.superWide = variant.size?.includes('super-wide');
```
Axelcureno, line 193: might be trivial but some variants have other sizes like `wide`

Axelcureno (approved):

studio/src/editors/merch-card-editor.js
yesil: @Axelcureno wide is supported by default. this flag is to show hide the extra-wide option when supported by the variant.

afmicka:
@yesil  seems like you are missing the latest (Mariia's fix is either missing or not workin on this branch https://github.com/adobecom/mas/pull/118). Could you please update?

afmicka:
@yesil Double-click on the CTA from the RTE opens OST on price offers instead of checkout.
Screenshot 2025-01-21 at 14 32 14

afmicka:
@yesil Could there be more clear separation between the fields (either bigger gap or maybe keep just an outline line for the field)? This "description" and "Footer" titles are barely noticeable (looks like they belong to the previous fields).
Screenshot 2025-01-21 at 14 59 57

afmicka (changes requested):
@yesil changes in any of the fields is not previewed anymore on the cards. Please check.

studio/src/editors/merch-card-editor.js
yesil: I also added flag for wide cards.

studio/src/editor-panel.js
```diff
@@ -94,20 +92,12 @@ export default class EditorPanel extends LitElement {
...
-
-    aemAction(action, reset = false) {
-        return async function () {
-            this.disabled = true;
-            const ok = await action();
-            if (ok && reset) this.hasChanges = false;
-            this.disabled = false;
-        };
+        Store.fragments.discard.set(true);
+        await this.updateComplete;
```
yesil, line 99: this is causing render nothing / then render everything forcing fields to refresh their values from the fragment.

studio/src/store.js
```diff
@@ -1,9 +1,9 @@
 import { Fragment } from './aem/fragment.js';
-import { getEditorPanel } from './editor-panel.js';
```
yesil, line 2: editor-panel already imports the store api.
transitive updates causing problems.

afmicka:
@yesil Duplicate and Publish fragment have the same issue that 'Save' had. Can't perform the action. 
Screenshot 2025-01-27 at 14 23 13

studio/src/editor-panel.js
```diff
@@ -35,79 +35,51 @@ export default class EditorPanel extends LitElement {
...
         return document.querySelector('mas-repository');
     }
 
-    fragmentStoreController = new StoreController(this, Store.fragments.inEdit);
-
     /** @type {FragmentStore | null} */
     get fragmentStore() {
-        if (!this.fragmentStoreController.value) return null;
-        return this.fragmentStoreController.value;
+        if (this.discarded) return null;
```
yesil, line 74: forces merch-card-editor to not render so that the field states are reset.

afmicka:
@yesil what is this bracket after the tags? :) 
Screenshot 2025-01-27 at 17 04 57

yesil:
@afmicka, oh, it is gone now.
image

afmicka:
@yesil Leaving here as a reminder. When you open a slice card, change the variant to suggested, and then back again to slice, the size dropdown field is empty. Though card has turned back to wide slice by default. 
Screenshot 2025-01-27 at 17 10 08

afmicka:
@yesil when you click clone on MAIN, it greys out the fields (fields and actions on top) indicating that the operations is happening (screenshot). On your branch, nothing happens in the editor, user is not sure if anything is going on. The fragment gets cloned at the end but is missing this visual experience.
Screenshot 2025-01-29 at 11 35 59

yesil:
@afmicka  the placeholder behavior is valid w.rt. SWC. I don't know where to fix it. Otherwise I addressed all the issues you reported.
Added also a dialog for the deletion
image

afmicka:
@yesil I believe on main is also SWC, right? The changes are breaking Nala. The field value update is present on main at the moment. Lets discuss on a call either the fix or the test change.

3ch023:
I can still edit the card while spinner is running.
Imo greyed out fields option was more clear, lets ask Nick&Vetri to chose

3ch023:
@yesil discussed offline with @afmicka, we just need to change here
https://github.com/adobecom/mas/blob/830993e0d9f6c4011ce3c782b36f29743eedfaf9/nala/studio/studio.test.js#L154

instead of checking attribute, 
```await expect(
                await studio.editorPanel.locator(studio.editorTitle),
            ).toHaveAttribute('value', `${data.title}`);
```
we can check element value property and should work good
`await studio.editorPanel.locator(studio.editorTitle).value`

probably it makes sense to fix it in this PR

afmicka:
@yesil @3ch023 there is another issue with these changes. As a consequence of not updating the value, automation does not see the field changed in preview either. I can not reproduce manually but this is the image from github actions (and i get the same when i run test locally) - title is changed in the editor but not on the card
test-failed-1

studio/src/store.js
```diff
@@ -1,16 +1,37 @@
...
+import { FragmentStore } from './reactivity/fragment-store.js';
+import Events from './events.js';
 
 const params = Object.fromEntries(
     new URLSearchParams(window.location.hash.slice(1)),
 );
 const initialSearch = MasSearch.fromHash();
 const initialFilters = MasFilters.fromHash();
 
+function discardChanges() {
```
st-angelo-adobe, line 15: Personally I don't think this should be here. 
This file should be kept as simple as possible, with just stores.
This is editor logic, and should belong in the editor files imo. Otherwise this file will look very weird in a few months.

studio/src/editor-panel.js
```diff
@@ -33,24 +31,44 @@ export default class EditorPanel extends LitElement {
...
+        /* Optional: Styles for the dialog */
+        sp-dialog {
+            max-width: 500px;
+        }
     `;
 
-    createRenderRoot() {
-        return this;
-    }
+    inEdit = Store.fragments.inEdit;
```
st-angelo-adobe, line 41: This isn't how we're supposed to use stores. Especially since you're subscribing to them reactively below anyway, with a new controller `ReactiveController`, which to me seems redundant.
If you want to react to `inEdit`, you just use a `StoreController` for it. Same for `operation`. This looks like `StoreController` with extra steps 😄 

Again, we can talk tomorrow in the dev call, but most of the time we should just use `StoreControllers`. That is their purpose, to give you access to the store value, reactively, which is what you want in 95%+ cases.

yesil: Actually I want to remove the controller logic except for the reactivity to store changes.
Currently there are two ways of doing the same thing.
All the data read/update should be done via the reactiveStores.
we don't really need more.
I could remove the controller totally it would still work.

yesil: you will have to convince me why controllers are needed to access store data.

afmicka:
@yesil there is a regression with discard changes. When you make a change to the card and close without saving, the preview of the card stays in the changed state instead of getting back to the original state.

studio/src/studio.js
```diff
@@ -60,6 +60,7 @@ class MasStudio extends LitElement {
 
     currentPage = new StoreController(this, Store.currentPage);
     commerceEnv = new StoreController(this, Store.commerceEnv);
+    currentPage = new StoreController(this, Store.currentPage);
```
st-angelo-adobe, line 63: duplicate assignment

afmicka (approved):

PR #125, opened 2025-01-02
MWPW-164165: fix unit text and old price toggles

yesil:
Resolve:
- https://jira.corp.adobe.com/browse/MWPW-164165
- https://jira.corp.adobe.com/browse/MWPW-164166

Test URLs:
- Before: https://main--mas--adobecom.hlx.live/studio.html
- After: https://mwpw-164165--mas--adobecom.hlx.live/studio.html

PR #126, opened 2025-01-03
MWPW-162020:  Add char counter to RTE field

yesil:
This PR is to be merged after #125 

image

Resolve: https://jira.corp.adobe.com/browse/MWPW-162020

Test URLs:
- After: https://mwpw-162020--mas--adobecom.hlx.live/studio.html


studio/src/utils/throttle.js
```diff
@@ -0,0 +1,24 @@
+function throttle(func, limit) {
```
npeltier, line 1: small introduction comment?

npeltier (approved):

PR #129, opened 2025-01-11
MWPW-165152: Splash screen / content nav URL handling

Axelcureno:
This PR improves URL handling and navigation between splash screen and content navigation, specifically introduces the following criteria:

- When URL has no hashes or searchParams, user is taken to Splash screen (assumes there is no data query).
- When a hash is part of the URL, user is taken to content navigation (for querying or searching content directly).
- When user clicks on "Home" in side nav, user is taken to splash screen and `page=welcome` searchParam is added to URL  (query and path are kept for context).
- when user clicks on "Go To Content" in Splash screen, user is taken to content navigation screen and `page=welcome` is removed from URL.
- When Studio URL has `?page=welcome`, user is taken directly to Splash screen (assuming it's intended).

This allows querying data within studio with URL params bypassing splash screen while letting the user navigate to splash screen when intended.

Resolves [MWPW-165152](https://jira.corp.adobe.com/browse/MWPW-165152)

Test URLs:
- After: https://mwpw-165152--mas--adobecom.hlx.live/studio.html

### DIFF
--- a/studio/src/mas-side-nav.js
+++ b/studio/src/mas-side-nav.js
@@ -4,6 +4,7 @@ import { navigateToPage } from './store.js';
 class MasSideNav extends LitElement {
     static styles = css`
         side-nav {
+            height: 100%;
             grid-column: 1 / 2;
             display: flex;
             flex-direction: column;
@@ -68,6 +69,13 @@ class MasSideNav extends LitElement {
         }
     `;
 
+    navigateToSplash() {
+        const url = new URL(window.location.href);
+        url.searchParams.set('page', 'welcome');
+        history.pushState({}, '', url.toString());
+        navigateToPage('splash')();
+    }
+
     render() {
         return html`<side-nav>
             <div class="dropdown-container">
@@ -77,7 +85,7 @@ class MasSideNav extends LitElement {
                 <sp-sidenav-item
                     label="Home"
                     value="home"
-                    @click="${navigateToPage('splash')}"
+                    @click="${this.navigateToSplash}"
                     selected
                 >
                     <sp-icon-home slot="icon"></sp-icon-home>
--- a/studio/src/store.js
+++ b/studio/src/store.js
@@ -7,6 +7,16 @@ import { reactiveStore } from './reactivity/reactive-store.js';
 const initialSearch = MasSearch.fromHash();
 const initialFilters = MasFilters.fromHash();
 
+const url = new URL(window.location.href);
+const pageParam = url.searchParams.get('page');
+
+const initialPage =
+    pageParam === 'welcome'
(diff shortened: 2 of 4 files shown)


studio/src/studio.js
```diff
@@ -56,6 +56,25 @@ class MasStudio extends LitElement {
         ></mas-splash-screen>`;
     }
 
+    connectedCallback() {
+        super.connectedCallback();
+    }
+
+    handleInitialPageLoad() {
```
yesil, line 63: `handleInitialPageLoad` doesn't seem to be used.

studio/src/mas-side-nav.js
```diff
@@ -68,6 +69,13 @@ class MasSideNav extends LitElement {
         }
     `;
 
+    navigateToSplash() {
+        const url = new URL(window.location.href);
```
yesil, line 73: with the state manager, the history/url management should be consolidated into a single location as a side-effect or reaction to state update.
I think it should be in `navigateToPage`.

studio/src/store.js
```diff
@@ -7,6 +7,16 @@ import { reactiveStore } from './reactivity/reactive-store.js';
 const initialSearch = MasSearch.fromHash();
 const initialFilters = MasFilters.fromHash();
 
+const url = new URL(window.location.href);
+const pageParam = url.searchParams.get('page');
+
+const initialPage =
+    pageParam === 'welcome'
```
yesil, line 14: in the absence of search query in the URL, the pageParam should be set to welcome so that the welcome page has always the same url.

yesil (changes requested):

studio/src/store.js
```diff
@@ -31,7 +41,7 @@ const Store = {
     ), // 'render' | 'table'
     selecting: reactiveStore(false),
     selection: reactiveStore([]),
-    currentPage: reactiveStore(initialSearch.query ? 'content' : 'splash'), // 'splash' | 'content'
+    currentPage: reactiveStore(initialPage), // 'splash' | 'content'
```
st-angelo-adobe, line 44: I think since we already have a store for the current page, we can just link it to the hash, just as `search` and `filters` are, and rename `splash` to `welcome`. The only downside will be that we always have the `page` hash param in the url, but on the other hand it will be more consistent and easier to handle.

Right now we have the `currentPage` store that handles page navigation, but it's not directly linked to the query params, we have an intermediary who checks for `welcome` value and sets `splash` in the store. I say we just "cut the middle man".

This will also remove the need to handle history manually. @yesil mentioned history can be managed in `navigateToPage`, and I agree, but I think if we link the hash to the store this is no longer needed, the browser will do it for you.

PR #130, opened 2025-01-14
MWPW-164253 non VPN OST

3ch023:
* Change IO Actions project to 'MerchAtScale'
* switch OST to https://14257-merchatscale.adobeioruntime.net/api/v1/web/MerchAtScale/ost-products-read
* IO action to generate https://14257-merchatscale.adobeioruntime.net/api/v1/web/MerchAtScale/ost-products
* GH trigger to refresh cache (24 h)

Resolve https://jira.corp.adobe.com/browse/MWPW-164253

Added datetime to see when the list was generated last time:
image
 to ease debugging.

unfortunately on Adobe network the request still takes around 1sec, which is slower than we used to have before

Test URLs:
- After: https://MWPW-164253--mas--adobecom.hlx.live/studio.html


npeltier (approved):
lgtm

yesil (approved):

3ch023:
will hold on with merge till i am more sure runtime if runtime cache works

Axelcureno (approved):

3ch023:
this is ready for QA now

npeltier (approved):
lgtm

npeltier:
2 things though @3ch023: 
1. should we consider having 2 different workspaces one for "live", and one for authoring, as we'll probably not have same scalability / constraints / shared actions and all?
2. have you considered using File rather than State  that seems better suited for "large" files. Now large limit is set at 1MB so we are still not large with 500kb payload but still. Also looks like state is better deployed, however i suspect you need more calls before all regions are warmed. (https://developer.adobe.com/app-builder/docs/guides/application_state/#feature-matrix)

3ch023:
> 1. should we consider having 2 different workspaces one for "live", and one for authoring, as we'll probably not have same scalability / constraints / shared actions and all?

makes sense. I think not in this PR, but we will need the separation in future

io-actions/app.config.yaml
```diff
@@ -27,4 +27,33 @@ application:
             annotations:
               require-adobe-auth: false
               final: true
+          translate:
```
npeltier: nope

io-actions/app.config.yaml
```diff
@@ -27,4 +27,33 @@ application:
...
+              OST_WRITE_API_KEY: $OST_WRITE_API_KEY
+            annotations:
+              require-adobe-auth: false
+          ost-products-read:
+            function: src/ost-products/read.js
+            web: 'yes'
+            runtime: nodejs:22
+            annotations:
+              require-adobe-auth: false
+        sequences:
```
npeltier: nope

npeltier (approved):
lgtm

io-actions/src/ost-products/README.md
```diff
@@ -0,0 +1,18 @@
+# ost-products-write
+
+This action generate will:
```
mirafedas, line 3: Should it be `This action will:` instead of `This action generate will:`?

io-actions/src/ost-products/README.md
```diff
@@ -0,0 +1,18 @@
...
+* query AOS API 
+* generate a light-weight list of products
+This list is used in the initial Product List screen of OST.
+The list will be stored in action worspace Store for a year.
```
mirafedas, line 7: worspace > workspace ;)

io-actions/src/ost-products/write.js
```diff
@@ -0,0 +1,166 @@
...
+
+const paginatedOffers = (allProducts, landscape, locale, params, page = 0) => {
+    const { AOS_URL, AOS_API_KEY } = params;
+    const [, country] = locale.split('_');
+    const offersEndpoint = `${AOS_URL}?country=${country}&merchant=ADOBE&service_providers=MERCHANDISING&locale=${locale}&api_key=${AOS_API_KEY}&landscape=${landscape}&page_size=100&page=${page}`;
+    return fetch(offersEndpoint)
+        .then((response) => response.json())
+        .then((offers) => {
+            console.log(`received ${landscape} - ${page}}`);
+            if (offers && offers.length > 0) {
```
mirafedas, line 46: This can be shortened to `if (offers?.length > 0) {`

mirafedas (approved):

Axelcureno (approved):

PR #131, opened 2025-01-16
Implemented HashLinkedStore and removed hash-manager component

st-angelo-adobe:
Resolves: [MWPW-165445](https://jira.corp.adobe.com/browse/MWPW-165445), [MWPW-165152](https://jira.corp.adobe.com/browse/MWPW-165152), [MWPW-165149](https://jira.corp.adobe.com/browse/MWPW-165149)

Test URLs:
- After: https://mwpw-165445--mas--adobecom.hlx.live/studio.html


studio/src/modifiers.js
```diff
@@ -0,0 +1,4 @@
+export function pageModifier(value) {
```
3ch023, line 1: imo right now there is no benefit to put this 2 lines of code in the reusable separate file, I would keep it in store.js

markovicpredrag1209:
Hey @st-angelo-adobe, because of the URL changes our tests will need an update also. 
One of the possible solutions would be to change [these](https://github.com/adobecom/mas/blob/d137cff39d25765578ba5a8dc1e4c1a181869818/nala/libs/imslogin.js#L27) two lines to something like this:
```
  let formattedUrl = props.url.toLowerCase();
  expect(page.url()).toContain(formattedUrl);

```
In this case we are using toContain() - which is case sensitive, toHaveURL is not
Also when on content page( https://mwpw-165445--mas--adobecom.hlx.live/studio.html#page=content&path=acom ), refresh doesnt stay on same page, it changes back to a welcome page ( https://mwpw-165445--mas--adobecom.hlx.live/studio.html#page=welcome&path=acom )

studio/src/modifiers.js
st-angelo-adobe: I wrote 2 more validators. I put them in the same file as the stores, but maybe we will want to actually move them to a `validators.js` file some time.

st-angelo-adobe:
@markovicpredrag1209 Hey. It seems like the tests are ok now, but I didn't change anything. Not sure if anything was changed on `main` and I got those changes when I merged.

The other issue you mentioned should be fixed as well.

studio/test/store/hash-linking.test.js
```diff
@@ -0,0 +1,50 @@
+import { expect } from '@esm-bundle/chai';
+import Store, { linkStoreToHash, unlinkStoreFromHash } from '../../src/store';
```
yesil, line 2: pleas use `store.js` extension, in unit tests WTR adds it automatically but such imports do not work in runtime.
e.g: 
image

studio/src/aem/mas-filter-panel.js
```diff
@@ -23,6 +25,18 @@ class MasFilterPanel extends LitElement {
         }
     `;
 
+    filters = new StoreController(this, Store.filters);
+
+    #updateFilter(property) {
```
yesil, line 30: maybe we can use name more in line with what it does: #getFilterHandler

studio/src/studio.js
```diff
@@ -58,19 +62,19 @@ class MasStudio extends LitElement {
         return BUCKET_TO_ENV[this.bucket] || BUCKET_TO_ENV.e59433;
     }
 
-    currentPage = new StoreController(this, Store.currentPage);
+    page = new StoreController(this, Store.page);
```
yesil, line 65: `pageController`?

studio/src/studio.js
```diff
@@ -58,19 +62,19 @@ class MasStudio extends LitElement {
...
     commerceEnv = new StoreController(this, Store.commerceEnv);
 
     get content() {
-        if (this.currentPage.value !== 'content') return nothing;
+        if (this.page.value !== 'content') return nothing;
```
yesil, line 69: I wonder if we should use `controller.value` or `Store.page.value`.
The reason is, we can use `Store.*.*` notation anywhere, and use controller only to react changes when needed.
In order to establish a best practice for reading values from the store, I find `Store.page.value`.
another advantage, is that we can use `Store.page.subscribe`/`Store.page.unsubscribe` if needed.

wdyt @npeltier @3ch023 @Axelcureno ?

studio/test/store/hash-linking.test.js
yesil, line 1: I would give a more functional name, `hash-linking` sounds to me like we are manipulating a linked hash map :)

studio/src/studio.js
st-angelo-adobe: I agree with you about reading values directly from the store where reactivity isn't needed, but this is not a good example. The reactivity is needed here, because `studio.js` needs to react to the page change. So if we are going to use a controller, then we might as well use the reactive value.

I would only use the value directly from the store when we DON'T necessarily need the reactivity, and we just want the value at some point in time, like in a handler or something.

I also think `Store.page.value` is ... convenient, but I was thinking maybe it's actually dangerous to expose the value. I even left a comment where we probably should only interact with the value through the getters and setters, and of course controllers, which should still be the main way to access values. Maybe we should discuss that more.

So imo, we should use:
`pageController = new StoreController(this, Store.page);`
in most places, and:
`const page = Store.page.get();`
where convenient.

studio/src/studio.js
yesil: since these are lit controller, I think we should follow it, but let's discuss it in the dev call.

studio/src/studio.js
yesil: when there are two ways of doing the same thing it starts confusing people.
I think we will have several places where you need to read a value from the Store or set it, but not react to it.
you aready using it in `navigateToPage`  with `Store.currentPage.set(value);` without needing a controller.
my bottom line suggestion is to reduce the role of controller to enable reactivity in the page
and use Store.*.* for any other neeeds.
let's discuss this in the dev call as well.

studio/src/mas-repository.js
```diff
@@ -182,23 +187,22 @@ export class MasRepository extends LitElement {
...
+        if (this.page.value !== 'welcome') return;
         if (this.#abortControllers.recentlyUpdated)
             this.#abortControllers.recentlyUpdated.abort();
         this.#abortControllers.recentlyUpdated = new AbortController();
 
         Store.fragments.recentlyUpdated.loading.set(true);
 
         const dataStore = Store.fragments.recentlyUpdated.data;
-        const path = this.search.value.path + Store.locale.path;
+        const path = `${this.search.value.path}/${this.filters.value.locale}`;
```
3ch023, line 205: can filter.value be set to null? if yes maybe `?.locale`

studio/src/studio.js
3ch023: my vote goes for page, doesn't look too confusing atm

studio/test/store/hash-linking.test.js
3ch023: haha true, but @yesil what name do you have in mind?

3ch023 (approved):

studio/src/studio.js
3ch023: i've checked the thread and i think i understand what @yesil idea is, but i am not sure i have a preferrence here.
imo both ways are more or less clear as long as we have a few examples in existing WC.
lets decide today pls and proceed with this PR, it's blocking other things

studio/test/store/hash-linking.test.js
yesil: Router
RouteManager

studio/src/mas-repository.js
yesil: `locale` should never be null though, not anymore.

studio/test/store/hash-linking.test.js
yesil: I just need the file extension, then I'm good, will approve.

studio/test/store/hash-linking.test.js
st-angelo-adobe: I think it should stay like this, because that's what it's called everywhere (`linkHashToStore`, `hashLink`, etc.) If we want to change we should probably rename everything.

studio/src/mas-repository.js
st-angelo-adobe: It can be whatever we put in there to be honest, but I don't like being afraid that x or y value can be null. It makes no sense for the filter object to ever be null, so we should make sure it's not, and trust 100% that the store value is valid. We can use validators for that btw.

st-angelo-adobe: I added the following logic in the validator, just in case:
`if (!value) return { locale: 'en_US' };`

yesil: I also don't like adding useless null checks, it makes the code illegible.

yesil (approved):

npeltier (approved):

Axelcureno (approved):

afmicka:
@st-angelo-adobe When using 'query' parameter and the fragment is found, should it also append the path to the URL and/or update the folder? 
cc: @3ch023 @npeltier @yesil 
 
Screenshot 2025-02-10 at 12 07 45

PR #133, opened 2025-01-16
MWPW-164401 unoptimized translation action

npeltier:
- fetch that is simple freyja fetch,
- translate get whatever is retrieved from management API (for now, waiting for ODIN-260 / ODIN-26000),
- ~~convertSchema converts mgt schema to freyja one~~
- fragment as fetch,translate ~~,convertSchema~~ sequence

this can be tested with manually localized fragemnt (for now):  https://14257-merchatscale-npeltier.adobeioruntime.net/api/v1/web/MerchAtScale/fragment?id=830f76be-0e83-4faf-9051-3dbb1a1dff04&locale=fr_FR


io-actions/src/fragment/translate.js
```diff
@@ -0,0 +1,44 @@
...
+async function main({ status, body, locale }) {
+    if (!status || status != 200 || !body || !locale) {
+        return {
+            status: 400,
+        };
+    }
+    const match = body.path?.match(PATH_TOKENS);
+    if (!match) {
+        return {
+            status: 400,
```
3ch023, line 18: lets add the error reason? smth like
`reason: 'Fragment Path ${body.path} doesn't match the regex',`

io-actions/app.config.yaml
```diff
@@ -18,3 +24,17 @@ application:
...
               final: true
+          translate:
+            function: src/fragment/translate.js
+            runtime: nodejs:18
+            inputs:
+              annotations:
+                require-adobe-auth: false
+        sequences:
+          fragment:
+            actions: 'fetch,translate'
```
3ch023, line 35: we have a choice here between piping several mini actions or having 1 action executing all steps.
And I am not sure what is better. For backend services piped actions give the possibility to reuse code.

In our use case, 1 action might execute faster. It's an assumption, but transferring context from one action to another must be costing time, no?

And even if we decide to have 1 action to get translated fragment, 1 to do placeholders, 1 to combine collection, having an action doing a fetch feels still way too granular.

io-actions/src/fragment/translate.js
```diff
@@ -0,0 +1,44 @@
...
+        );
+        const resp = await response.json();
+        if (resp?.items?.length == 1) {
+            return {
+                status: 200,
+                body: resp.items[0],
+            };
+        } else {
+            return {
+                status: 400,
```
3ch023, line 34: 404?

io-actions/src/fragment/translate.js
```diff
@@ -0,0 +1,44 @@
...
+        return {
+            status: 400,
+        };
+    }
+    const { surface, parsedLocale, fragmentPath } = match.groups;
+    if (parsedLocale !== locale) {
+        const response = await fetch(
+            `https://odin.adobe.com/adobe/sites/fragments?path=/content/dam/mas/${surface}/${locale}/${fragmentPath}`,
+        );
+        const resp = await response.json();
```
3ch023, line 26: nit: rename 'resp' to 'json' or 'content' or anything. Having 'response' and 'resp' together is risky

3ch023 (approved):

3ch023:
approved, with some comments for consideration.
+ pls add a test page for nala tests :)

io-actions/test/fragment/translate.test.js
```diff
@@ -0,0 +1,115 @@
+jest.mock('node-fetch');
```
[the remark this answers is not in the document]
npeltier: sure let's change before it's too late @yesil ? mocha? something else?

io-actions/src/fragment/fetch.js
```diff
@@ -0,0 +1,20 @@
+const fetch = require('node-fetch');
+
+async function main({ id, locale }) {
+    if (id) {
+        const response = await fetch(
```
yesil, line 5: we need to handle possible errors here

io-actions/test/fragment/translate.test.js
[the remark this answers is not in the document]
yesil: mocha with chai please :)

io-actions/app.config.yaml
npeltier: we will have different types of fragment (at least individual fragment, and collection for now), 
some common actions (translate, replace placeholders, more?), 
principle of serveless micro action is to have small actions, with one purpose, would be antipattern to have growing processes (fetch & translate is just a beginning)

io-actions/test/fragment/translate.test.js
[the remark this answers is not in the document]
npeltier: sure, sir :) 
ok, will update your UT too, @3ch023

io-actions/src/fragment/fetch.js
npeltier: true, will move some of the translate cases there

io-actions/test/fragment/translate.test.mjs
yesil, line 1: any reason for the `mjs` extension?

io-actions/test/fragment/translate.test.mjs
```diff
@@ -0,0 +1,104 @@
+import chai from 'chai';
+import nock from 'nock';
+import action from '../../src/fragment/translate.js';
+
+const expect = chai.expect;
+
+describe('translate typical cases', () => {
+    it('should return translated fragment', async () => {
+        nock('https://odin.adobe.com')
```
yesil, line 9: nice  ❤️

yesil (approved):

io-actions/test/fragment/translate.test.mjs
npeltier: at some point of my testing odyssey, mocha was ranting at my files not being modules, may be i can remove this

io-actions/app.config.yaml
honstar: One thing to consider is anticipated traffic and the volume of requests. If these responses cannot be cached on CDN level, having a single action would be better in terms of performance.
OTOH, having microservices mapped to individual runtime actions makes sense in terms of modularity and reuse. I can't really tell how much the expected cost is by chaining actions in a sequence. This would be a good question asked on the app-builder slack channel.

io-actions/test/fragment/translate.test.mjs
3ch023: Jason and js frameworks

io-actions/app.config.yaml
npeltier: discussed on #app-builder and it indeed seems that it's not always a good idea, especially that all "sub actions" are not really idempotent as they depend on each other in good order

PR #137, opened 2025-01-21
MWPW-165966: Add locale picker feature

yesil:
- Updated studio.html
- Modified mas-filter-panel.js to support new locale picker
- Added locale-picker.js for locale selection functionality
- Updated mas-repository.js to handle locale data
- Modified store.js to store locale information

⚠️  deeplink support is not added yet. I'll do it in a separate PR once I know more about he future implementation.

**Screenshots**
image

Resolve MWPW-165966

Test URLs:
- After: https://mwpw-165966--mas--adobecom.hlx.live/studio.html


npeltier (approved):

afmicka:
@yesil Any thoughts of ordering the county list in an alphabetical order maybe? (or something else that will not require random scrolling) 
cc: @3ch023 @npeltier 
Screenshot 2025-01-21 at 17 21 30

yesil:
@afmicka maybe I can put the most active countries at the top and separate the rest with a divider as follows?
image

afmicka:
@yesil well, whatever ends up in a longer (not so active) list, should probably be alphabetical. On the top we can have tier1 countries or most active ones but finding a country should be logical.

studio/src/store.js
```diff
@@ -28,6 +28,40 @@ const Store = {
         loaded: reactiveStore(false),
         data: reactiveStore([]),
     },
+    locale: {
+        current: reactiveStore('en_US'),
+        get path() {
+            return '/' + Store.locale.current.get();
+        },
+        data: [
```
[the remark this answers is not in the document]
3ch023: depends on how hard dynamic part is, we could just hardcode it per surface

3ch023 (approved):
Approving, but I had hard times to find locale switch.
IMO filters should be expanded by default, not sure if there was a req from Nick to hide them?

studio.html
```diff
@@ -52,7 +52,6 @@
         </script>
         <script src="studio/src/maslib.js" type="module"></script>
         <script src="studio/libs/swc.js" type="module"></script>
-        <script src="studio/src/aem/index.js" type="module"></script>
```
Axelcureno, line 55: Just curious, why is this no longer needed?

Axelcureno (approved):

studio/src/filters/locale-picker.js
```diff
@@ -0,0 +1,47 @@
...
+
+    get value() {
+        return Store.locale.current.get();
+    }
+
+    #handleChange(event) {
+        Store.locale.current.set(event.target.value);
+        this.dispatchEvent(
+            new CustomEvent('change', {
+                detail: { value: this.value },
```
Axelcureno, line 41: no `bubbles`?

yesil:
@afmicka finally I decided to sort alphabetically for now.

studio/src/filters/locale-picker.js
yesil: good point, actually this is not used for now. and  when used, if we need to catch the event from the ancestor elements, we can add it. For now, this keeps the noise low.

afmicka:
@yesil  you forgot Brazil somewhere down :) If not a trouble, move it up. The rest looks good, approving.

Screenshot 2025-01-22 at 14 29 10

PR #142, opened 2025-01-23
MWPW-165538 placeholder API

npeltier:
- rewritten as fragment based actions are chained & tested https://github.com/adobecom/mas/commit/b867cf3ee3b3e479f8d77fed30cc6e7c03afd6d8 within "pipeline" single action mapped to as "/fragment" api, with all "subactions" being separate functions (cf. https://github.com/adobecom/mas/pull/133/files#r1920418858)
- added replace action together with test on all different functionality 

image

```
  replace
    ✔ returns 200 & no placeholders
    ✔ returns 200 & replaced entries keys with text
    ✔ returns 200 & replace empty (but present) placeholders
    ✔ returns 200 & manages nested placeholders
    ✔ returns 200 & manages circular references
    ✔ returns 200 & leaves non existing keys
    ✔ returns 200 & manages rich text

  pipeline full use case
    ✔ should return fully baked /content/dam/mas/drafts/fr_FR/someFragment

  pipeline corner cases
    ✔ main should be defined
    ✔ no arguments should return 400

  translate typical cases
    ✔ should return fr fragment (us fragment, fr locale)
    ✔ should return fr fragment (fr fragment, no locale)

  translate corner cases
    ✔ no path should return 400
    ✔ bad path should return 400
    ✔ same locale should return same body

  health-check
    ✔ main should be defined
    ✔ return 200 & success status if odin & wcs are valid
    ✔ return 500 & error if odin json is not valid (no fields object)
    ✔ return 500 & error if odin cdn is 404
    ✔ return 500 & error if WCS is 503


  20 passing (103ms)

```
you can see full pipeline in action there: https://14257-merchatscale-npeltier.adobeioruntime.net/api/v1/web/MerchAtScale/fragment?id=7c87d1c4-d7fd-4370-b318-4cb2ebb4dd13&locale=fr_FR


io-actions/src/fragment/fetch.js
```diff
@@ -1,11 +1,9 @@
 const fetch = require('node-fetch');
-
-async function main({ id, locale }) {
+const { odinId } = require('./paths.js');
```
yesil, line 2: I wonder as a general rule we should be using `mjs` format and use `import` instead of `require`

io-actions/src/fragment/pipeline.js
```diff
@@ -0,0 +1,22 @@
+const fetchFragment = require('./fetch.js').fetchFragment;
+const translate = require('./translate.js').translate;
+const replace = require('./replace.js').replace;
+
+async function main(params) {
+    let context = { ...params, status: 200 };
+    for (const worker of [fetchFragment, translate, replace]) {
```
yesil, line 7: nit: suggesting `transformer`

io-actions/src/fragment/replace.js
```diff
@@ -0,0 +1,105 @@
...
+    return null;
+};
+
+const extractValue = (ref) => {
+    const value = ref.value || ref?.richTextValue?.value || '';
+    // Escape control characters before parsing
+    return value.replace(/[\u0000-\u001F\u007F-\u009F]/g, '');
+};
+
+const getDictionary = async (context) => {
```
yesil, line 28: I recommend to use function instead of arrow functions like `const getDictionary`, the reason is that, afaik in case of error, the function name is not logged in the console, but rather logged as an anonymous function.

io-actions/src/fragment/replace.js
```diff
@@ -0,0 +1,105 @@
...
+            });
+            return dictionary;
+        }
+    } catch (e) {
+        console.error(e);
+    }
+    return null;
+};
+
+replaceValues = (input, dictionary, calls) => {
```
yesil, line 55: declared without `const` or `let` ?

io-actions/src/fragment/fetch.js
npeltier: i switched every test back from mjs :) i like that it reminds we are working on node.js code.

io-actions/src/fragment/replace.js
npeltier: will be a function anyway :)

io-actions/src/fragment/fetch.js
yesil: we could run & debug these tests in WTR with import, but not with require.
Not sure if that makes sense to test backend APIs though.

npeltier: in a browser? :) not really no

3ch023: i'd keep require for now

io-actions/README.md
```diff
@@ -2,12 +2,12 @@
 
 ## Setup / First Steps
 
-- install aio cli
-- Request access to I/O Runtime in Adobe Corp org
+- install aio cli running `npm install -g @adobe/aio-cli`
```
3ch023, line 5: <3

io-actions/src/fragment/pipeline.js
```diff
@@ -0,0 +1,22 @@
...
+const replace = require('./replace.js').replace;
+
+async function main(params) {
+    let context = { ...params, status: 200 };
+    for (const transformer of [fetchFragment, translate, replace]) {
+        if (context.status != 200) break;
+        context = await transformer(context);
+    }
+    returnValue = {
+        status: context.status,
```
3ch023, line 12: we should return reason as well, it will speed up debugging of issues

3ch023 (approved):
If GitHub had a like I'd put it on this PR

3ch023:
putting trivial since there is no impact to Studio as of now. Nala tests failing because of using fork

PR #148, opened 2025-01-30
MWPW-165459 collection transformer

npeltier:
- collection transformer that injects cards & categories fields to the body if fragment belongs to /collections/ folder,
- unit test

Resolve [MWPW-165459](https://jira.corp.adobe.com/browse/MWPW-165459)

it can be seen in action here: https://www.stage.adobe.com/mas/io/fragment?id=09634c2e-1a8f-49c1-936e-8540cac715b7&api_key=nico&locale=fr_FR


3ch023 (changes requested):
failing nala tests

afmicka:
@3ch023 fails because it is fork... @npeltier Can you raise this from upstream?

npeltier:
@afmicka @3ch023 considering it's only io-actions and there is for now no possible impact on any nala test whatsoever, should i really do that? (closing this PR and raising a new one?)

3ch023 (approved):

io-actions/src/fragment/collection.js
```diff
@@ -0,0 +1,77 @@
...
+    const { body } = context;
+    const { id } = body;
+    if (isCollection(body)) {
+        const referencesPath = odinReferences(id, true);
+        try {
+            const response = await fetch(referencesPath);
+            if (response.status === 200) {
+                const root = await response.json();
+                const { references } = root;
+                const { cards, categories } = parseReferences(
```
yesil, line 55: if this is card collection endpoint, let's call it card collection, if it is a generic collection api, maybe we should avoid hardcoding card logic?

io-actions/src/fragment/paths.js
```diff
@@ -19,8 +19,8 @@ function odinId(id) {
  * @param {*} id id of the fragment,
  * @returns full fetchable path to the fragment references
  */
-function odinReferences(id) {
-    return `${odinId(id)}/variations/master/references`;
+function odinReferences(id, allHydrated = false) {
```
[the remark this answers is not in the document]
npeltier: this is only the publish pipeline for now, the author one will need another one (using same functions)

io-actions/src/fragment/collection.js
npeltier: well for now it's card, i don't see other collections to be coming soon. Maybe we can make it generic when that makes sense?

yesil: don't you want to call it card-collection api ?

io-actions/src/fragment/paths.js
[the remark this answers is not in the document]
yesil: what about QA env?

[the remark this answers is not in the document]
npeltier: and stage! later story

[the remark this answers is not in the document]
npeltier: (and it's not specific to that transformer)

io-actions/src/fragment/collection.js
npeltier: the API is still the same, it's fragment, i called it collection api but it's not, the content fragment model is called collection too, as for now collection is pretty clear for everyone. I can rename everything, but i'm not sure it's really clearer?

npeltier: renamed PR "collection transformer" :)

yesil (approved):

PR #149, opened 2025-01-31
MWPW-166929: Added logic to close popover if folder picker loses focus

markovicpredrag1209:
Fix for popover staying open when folder picker loses focus

Resolve: https://jira.corp.adobe.com/browse/MWPW-166929

Test URLs:
- After: https://MWPW-166929--mas--adobecom.aem.live/studio.html


3ch023:
tried the feature branch link and for me popover is still open when focus is gone

3ch023 (approved):

st-angelo-adobe (approved):

Roycethan (changes requested):
@markovicpredrag1209  Although clicking outside, closing the popover, it would be ideal  for better usability and accessibility to close the pop over when user clicks Escape or any other Tabbing keys. Requesting change as part of this to cover this use case Thanks

Roycethan:
@markovicpredrag1209 Escaping closing the pop after your latest commit, the other use case of tabbing still persists... example keep tab using key and once focus reach inside popover dropbox , again start tabbing , in this case also popover should close and user should go to the next element and able to interact with it..

https://github.com/user-attachments/assets/2063809e-95ac-42d5-92f3-bb10bf18194e

Roycethan (approved):

PR #150, opened 2025-01-31
MWPW-165818 add loc ready toggle

npeltier:

PARTIALLY Resolve [MWPW-165818](https://jira.corp.adobe.com/browse/MWPW-165818)

Test URLs:
- Before: https://main--mas--adobecom.hlx.live
- After: https://MWPW-165818--mas--adobecom.hlx.live


studio/src/editor-panel.js
```diff
@@ -212,6 +212,12 @@ export default class EditorPanel extends LitElement {
         this.hasChanges = true;
     }
 
+    #handleLocReady() {
```
3ch023, line 215: had similar change in my ticket for a checkbox and we could result the usual handle method 
but this way is also ok

3ch023 (approved):

mirafedas (approved):

Axelcureno (approved):

PR #151, opened 2025-01-31
MWPW-166928: update to check for fragment id before removing from results

markovicpredrag1209:
Fixes the bug where UI shows wrong fragment removed

Resolve: https://jira.corp.adobe.com/browse/MWPW-166928

Test URLs:
- After: https://MWPW-166928--mas--adobecom.aem.live/studio.html


studio/src/mas-repository.js
```diff
@@ -349,7 +349,9 @@ export class MasRepository extends LitElement {
 
             Store.fragments.list.data.update((prev) => {
                 var result = [...prev];
-                const index = result.indexOf(fragment);
+                const index = result.findIndex(
```
st-angelo-adobe, line 352: It's weird to me that the `inEdit` fragment can't be found in the fragment list, like it should.

Something changed regarding this? @yesil @3ch023 @Axelcureno

3ch023: Hi @st-angelo-adobe my guess is it was an array of fragments before so it used to work. now that its an array of fragmentStores it doesn't find the fragment anymore 
image

@markovicpredrag1209 probably could rename `item` in `fragmentStore` if you want to be a bit more declarative. but `item` is good too

3ch023 (approved):

3ch023:
@markovicpredrag1209 maybe you could extend a nala test also to verify that after fragment deletion the fragment is no longer reflected in studio?

markovicpredrag1209:
@3ch023 right now our test would search for a cloned fragment based on id, then it deletes the fragment and verifies that its not visible on the results page. Should i create a new test that would check deletion from go-to-content also or should i update the test we already have?

afmicka:
> @3ch023 right now our test would search for a cloned fragment based on id, then it deletes the fragment and verifies that its not visible on the results page. Should i create a new test that would check deletion from go-to-content also or should i update the test we already have?

yes! :) We either exclude the search for that fragment from the test and delete it directly or have a new test. But we need to be clever. The fragment we want to delete must NOT be the last on the screen, otherwise we would have false positive. Right now whatever you delete, it removes the last fragment from the view.

markovicpredrag1209:
> > @3ch023 right now our test would search for a cloned fragment based on id, then it deletes the fragment and verifies that its not visible on the results page. Should i create a new test that would check deletion from go-to-content also or should i update the test we already have?
> 
> yes! :) We either exclude the search for that fragment from the test and delete it directly or have a new test, does not really matter. But we need to be clever. The fragment we want to delete must NOT be the last on the screen, otherwise we would have false positive. Right now whatever you delete, it removes the last fragment from the view.
> 
> Also, there are test changes coming in #131. Maybe wait for those as they might get merged today

thats true, since we would delete after cloning, theres a pretty big chance that our cloned card would be the last on screen.

3ch023:
I am also happy to merge it as is and do that test later, @afmicka leaving it to you to decide :p

Axelcureno (approved):

PR #155, opened 2025-02-05
MWPW-165449: Tag picker in the filter-panel

yesil:
Also, slight improvements in the locale/country picker.

Resolve https://jira.corp.adobe.com/browse/MWPW-165449

Test URLs:
- After: https://mwpw-165449--mas--adobecom.aem.live/studio.html


Axelcureno (approved):

studio/src/aem/mas-filter-panel.js
```diff
@@ -1,72 +1,119 @@
...
+        const tags = [
+            ...this.shadowRoot.querySelectorAll('aem-tag-picker-field'),
+        ]
+            .flatMap((tagPicker) => tagPicker.getAttribute('value').split(','))
+            .filter((tag) => tag);
+        Store.search.set({ ...Store.search.get(), tags });
+    }
+
+    #handleRefresh() {
+        Store.search.set({ ...Store.search.get(), tags: [] });
```
st-angelo-adobe, line 30: I prefer to use `update` to change properties: `Store.search.update(prev => ({ ...prev, tags: [] }));`. And `set` for a full overwrite. As you prefer.

mirafedas (approved):

afmicka:
@yesil Couple of things in filters:
1. The dropdown arrows are on the left side. They are on the right side in figma, and honestly, for me is confusing and unnatural to be on the left side. Feels strange, on the first try i filtered the "wrong" thing (field next to the one i wanted). Is this intentional or can be moved to the right?
Screenshot 2025-02-10 at 14 39 11

2. Filtering by country: 
- there is not pill when the country is chosen (there is one in figma)
- as a consequence, it is a bit hard to determine what is the selection. The check icon next to the country easily blends in with the flag icons (maybe would help if it does not shift the selected country and stays "alone" in the column in front of all flags). Do we have figma for this dropdown? 
- There is also no indication that anything is selected for the country giving the impression that is shows all the cards, even though USA is selected by default. 

3. the tags parameter in the url stays empty once the filters are reset or canceled. Should it be removed? 
Screenshot 2025-02-10 at 14 42 18

4. There is no 'Published" status in the drop down for 'Status'. Intentional? 

Screenshot 2025-02-10 at 15 04 26

afmicka:
@yesil there is a regression. Not able to save fragment if you edit the checkout link via 'Edit link' 
Screenshot 2025-02-10 at 17 54 55

studio/src/aem/mas-filter-panel.js
st-angelo-adobe: Just wanted to mention that I changed the store a bit in here: https://github.com/adobecom/mas/pull/131
I removed the `update` method, and just made `set` handle both cases: overriding value, and using an updating function. Just for simplicity. So it could be `Store.search.set(prev => ({ ...prev, tags: [] }));` now.

yesil: does this mean we can update the store value only with a callback?

st-angelo-adobe: No, you can also just say `Store.page.set('content')`. It works both ways. 

I guess one scenario where it fails is if the store value is a function itself. But we don't yet have this scenario, and it's an easy fix anywhere if needed.

afmicka:
@yesil status picker has multi choice (as i believe it should). When you select 2 (draft and published) statuses, i get one "random" card as a result. Shouldn't there be all cards with published or draft status? 
Screenshot 2025-02-25 at 14 09 08

yesil:
@afmicka I implemented AND and OR logic.
There is one exception: **published**, although it loooks like a tag, it is not, it is the actual published status of the fragment.
Basically, when selected, the search result will contain only published fragments.
I'll leave it as is and see if it is confusing people or not.

image

afmicka:
@yesil the logic looks good except with status. It does not combine with it.
If i have selection like this for example, it gives me various cards.
Screenshot 2025-02-26 at 09 57 54

If i add status 'draft' to it (notice that all cards above have draft status, should show the same cards), it has empty search. 
Screenshot 2025-02-26 at 09 58 06

afmicka:
@yesil as discussed offline, created a follow-up jira to tackle filtering by draft status issue: https://jira.corp.adobe.com/browse/MWPW-168561
This PR is ready as it is.

PR #161, opened 2025-02-11
MWPW-167305 Callout Info Icon

bozojovicic:
In RTE field for callout in card editor, add the option to append the info icon with tooltip.
Make Badge and Quantity Selector authorable.

Resolve [MWPW-167305](https://jira.corp.adobe.com/browse/MWPW-167305)
Resolve [MWPW-167306](https://jira.corp.adobe.com/browse/MWPW-167306)

Test url : mwpw-167305--mas--adobecom.hlx.live/studio.html?milolibs=mwpw167306--milo--bozojovicic

Test URLs:
- After: https://mwpw-167305--mas--adobecom.aem.live


3ch023:
@bozojovicic i have a feeling that something went wrong with merge either of milo branch or mas branch because i still can't see in your test page card variant 'plans'
image

3ch023 (approved):

st-angelo-adobe (approved):

afmicka:
@bozojovicic i was waiting for @3ch023's PR to get merged to report this for you, but since @st-angelo-adobe already did, please keep it in mind. The fix will be needed once Mariia's PR gets in
https://github.com/adobecom/milo/pull/3649#issuecomment-2671414057

Axelcureno (approved):
As a note, I do see some subtle differences by loading milo styles in the spacing.

afmicka:
@bozojovicic When the badge is removed from the card, it also removes the border from the card preview. 
When the page is refreshed, the border on the card is back.
Screenshot 2025-03-05 at 14 20 06

afmicka:
@bozojovicic could you please change this label inside the RTE to "Enter the tooltip text" or something like that as in other fields? This way it looks like the text is already entered, confusing.

Screenshot 2025-03-05 at 14 24 52

bozojovicic:
@afmicka card border fixed (in milo code), tooltip text field label also renamed to "Enter the tooltip text".
I discussed with @3ch023 about editing of tooltip text and we decided to leave it like this. Implementation of editing is not easy at all and in this case editors would have no benefits from it. Now they need to remove the icon and insert new one with tooltip text. With editing they would need to select the icon, click on info icon in toolbar, change the text, click Save. Doesn't sound easier than delete/insert. I will document all this in Jira ticket.

afmicka:
@bozojovicic For quantity selector, if you uncheck and check the box again WITHOUT saving in the meantime, the quantity selector does not take the set values.

1. Open the card (notice the label and start value)
Screenshot 2025-03-06 at 10 15 25

2. Uncheck the quantity selector box and check it again
3. The values are the same as they were in editor but unset on the card

Screenshot 2025-03-06 at 10 15 39

bozojovicic:
@afmicka problems with quantity selector check/uncheck - fixed

PR #162, opened 2025-02-11
MWPW-167038: OSI filed in Card Editor

markovicpredrag1209:
Adding OSI field in merch card editor

- Opening pr for easier communication and suggestions
- Still need to add logic to post osi-field value in odin and to read from it when card editor opens
- Right now tags added by selecting an offer will overwrite any previously added tags on card, how to approach this?

Resolve: https://jira.corp.adobe.com/browse/MWPW-167038

Test URLs:
- After: https://MWPW-167038--mas--adobecom.aem.live/studio.html

### DIFF
--- a/studio/src/editors/merch-card-editor.js
+++ b/studio/src/editors/merch-card-editor.js
@@ -4,6 +4,7 @@ import '../fields/multifield.js';
 import '../fields/mnemonic-field.js';
 import '../aem/aem-tag-picker-field.js';
 import './variant-picker.js';
+import '../rte/osi-field.js'
 
 const MODEL_PATH = '/conf/mas/settings/dam/cfm/models/card';
 
@@ -218,6 +219,13 @@ class MerchCardEditor extends LitElement {
                     @change="${this.updateFragment}"
                     >${unsafeHTML(form.ctas.values[0])}</rte-field
                 >
+            </sp-field-group>
+            <sp-field-group>
+                <sp-field-label for="osi-field">OSI</sp-field-label>
+                <osi-field
+                    id="osi-field"
+                    data-field="osi-field"
+                ></osi-field>
             </sp-field-group>
             <aem-tag-picker-field
                 label="Tags"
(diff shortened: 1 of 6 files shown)


3ch023:
Hi @markovicpredrag1209 
tried it out today, looks good, but I think 'offer_type', 'plan_type' and 'market_segments' tags should be always overwritten with a new selection.
since there can only be on offer defining tags per card.
right now they are added to existing tags:
image


and another issue noticed, osi is not saved in odin it seems - when i reopen the card the field is not reflecting the value set before

studio/src/editors/merch-card-editor.js
```diff
@@ -252,6 +253,16 @@ class MerchCardEditor extends LitElement {
                     >${unsafeHTML(form.ctas.values[0])}</rte-field
                 >
             </sp-field-group>
+            <sp-field-group>
+                <sp-field-label for="osi">OSI</sp-field-label>
```
Axelcureno, line 257: Just "OSI" is not descriptive, I suggest adding "search" i.e. "OSI Search". When I first opened the page, I didn't notice that there is a text field at first.

studio/src/rte/ost.js
```diff
@@ -174,7 +190,7 @@ export function getOffferSelectorTool() {
     `;
 }
 
-export function openOfferSelectorTool(offerElement) {
+export function openOfferSelectorTool(triggerElement, offerElement) {
     const landscape =
```
Axelcureno, line 194: we should probably add a `try/catch` in open and close methods as they becomes more complex.

Axelcureno (changes requested):

Axelcureno (approved):
lgtm, thank you!

3ch023 (approved):

afmicka:
@markovicpredrag1209 Lookin info fragment "bd6de070-ec6c-4c11-9f64-2c571bc7f195" in sandbox, when you change the OSI to the offer id "yIcVsmjmQCHKQ-TvUJxH3-kop4ifvwoMBBzVg3qfaTg" it changes correctly the tag to M2M offer, but after saving and reopening the card, the tag is back to ABM
Screenshot 2025-03-03 at 13 14 26

afmicka:
thanks @markovicpredrag1209 . Fixed!

PR #164, opened 2025-02-12
MWPW-167647: Automatic folder selection for single fragment search

st-angelo-adobe:
Resolves: [MWPW-167647](https://jira.corp.adobe.com/browse/MWPW-167647)

Test URLs:
- After: https://mwpw-167647--mas--adobecom.aem.live/studio.html


bozojovicic (approved):

3ch023 (approved):

afmicka:
@st-angelo-adobe when i am already in studio and paste the url with the query parameter, it appends only 'path' to the url. From there if you cancel the search (x in the search field), you stay in the content page but not content (page) param in the url. 
Screenshot 2025-02-19 at 12 48 16

If i open the new tab and paste the link, it will append both 'path' and 'page' parameters and behave as expected from there
Screenshot 2025-02-19 at 12 48 24

afmicka:
> @afmicka Were you already on the content page when you pasted it the first time?

Yes, that happens from the content page

st-angelo-adobe:
@afmicka Yeah, so because you were already on the content page, the store value doesn't change and so there's no update to the hash. If this is an issue I guess I'll have to look over the mechanism and make it so that when ANY hash value changes, ALL values are updated. This is not a small change though.

afmicka:
@3ch023 @npeltier @yesil any thought of the above?

afmicka:
@st-angelo-adobe i have added the check for this to Nala tests.

PR #171, opened 2025-02-19
clone test updated

markovicpredrag1209:
Update cloning test to verify the correct card is removed from content page

https://jira.corp.adobe.com/browse/MWPW-167925

Test URLs:
- After: https://MWPW-167925--mas--adobecom.aem.live/studio.html

### DIFF
--- a/nala/studio/ccd/suggested/specs/suggested_save.spec.js
+++ b/nala/studio/ccd/suggested/specs/suggested_save.spec.js
@@ -5,6 +5,7 @@ export default {
             tcid: '0',
             name: '@studio-suggested-clone-edit-save-delete',
             path: '/studio.html',
+            contentPath: '/studio.html#path=nala&page=content',
             data: {
                 cardid: 'cc85b026-240a-4280-ab41-7618e65daac4',
                 title: 'Field Edit & Save',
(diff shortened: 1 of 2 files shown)


npeltier (approved):
looks good!

i wonder if we should not make all those methods for cards & editor synchronous (remove async) as it seems they are always used in synchronous way (always "await ...")

afmicka:
@markovicpredrag1209 lets discuss splitting this in a separate tests and executing in a single view. (we'll do it over a call)

nala/studio/ccd/suggested/specs/suggested_save.spec.js
```diff
@@ -21,5 +21,16 @@ export default {
             browserParams: '#query=',
             tags: '@mas-studio @ccd @ccd-suggested @ccd-suggested-save',
         },
+
+        {
+            tcid: '1',
+            name: '@studio-suggested-clone-twice-delete',
```
afmicka, line 27: could we give a name that is more informative of the check? Like '@studio-suggested-remove-correct-fragment' or something like that?

nala/studio/studio.page.js
```diff
@@ -71,7 +71,7 @@ export default class StudioPage {
         this.linkSave = page.locator('#saveButton');
     }
 
-    async getCard(id, cardType, cloned) {
+    async getCard(id, cardType, cloned, clonedAgain) {
```
afmicka, line 74: rename clonedAgain to give info that it is also fragment id in question- maybe secondID or excludeID?

afmicka (approved):

PR #173, opened 2025-02-19
MWPW-165605: Adding profile menu to the nav for signing out

seanchoi-dev:
Resolve: [MWPW-165605](https://jira.corp.adobe.com/browse/MWPW-165605)

Test URLs:
- After: https://MWPW-165605--mas--adobecom.aem.live/studio.html


studio/src/mas-top-nav.js
```diff
@@ -21,6 +64,10 @@ class MasTopNav extends LitElement {
         return EnvColorCode[this.aemEnv];
     }
 
+    get profile() {
```
st-angelo-adobe, line 67: Is this property being used somewhere?

seanchoi-dev: That was left over from initial code of mine. I removed it. Thanks.ᐟ

yesil:
@seanchoi-dev @3ch023 

why don't we use universal nav, same as adobe home?

image

npeltier:
> @seanchoi-dev @3ch023
> 
> why don't we use universal nav, same as adobe home?

@yesil  i think this is more of a product management decision to be taking. In the meanwhile having current IMS status & log out capability is a must

Axelcureno:
@seanchoi-dev plz resolve conflicts

3ch023 (approved):

Roycethan:
@seanchoi-dev Conflicts at studio/src/constants.js need to be resolved...

seanchoi-dev:
I will fix the conflict and the bug. This will be a temporal solution and there is another ticket to set universal navigation on m@s studio.

Roycethan:
> @seanchoi-dev
> 
> 1. Conflicts at studio/src/constants.js need to be resolved...
> 2. Thre's a Bug here: when clicked on Profile icon the Drpdown box is hidden behind the panel
>    https://mwpw-165605--mas--adobecom.aem.live/studio.html#path=sandbox&page=content
> 
> image 3) Yes i would agree on @yesil to have Universal Nav since almost evrything is migrated to U-nav might be a good time to normalize this as well. I'll ask Nick in ticket.
> cc @3ch023 @yesil @Axelcureno @npeltier @afmicka

Issue Fixed , Thanks

PR #179, opened 2025-02-28
MWPW-168870: CTA class fix

st-angelo-adobe:
Resolves: [MWPW-168870](https://jira.corp.adobe.com/browse/MWPW-168870)

Test URLs:

Before: https://main--mas--adobecom.aem.live/studio.html
After: https://mwpw-168870--mas--adobecom.aem.live/studio.html


yesil (approved):
A test would be nice to cover this use case.

yesil (approved):

Axelcureno (approved):

afmicka:
@st-angelo-adobe I am trying to understand what classes should remain. If i change the variant of the upt-link, it does change in the RTE, does mnot change the visual on the card but does change the classes. 
Should for upt-link always be primary link or not? 
Screenshot 2025-03-12 at 13 28 58

PR #189, opened 2025-03-10
Mwpw 168091

markovicpredrag1209:

Resolve: https://jira.corp.adobe.com/browse/MWPW-168091

Test URLs:
- After: https://MWPW-168091--mas--adobecom.aem.live/studio.html

Created two subtasks to deal with open issues:
https://jira.corp.adobe.com/browse/MWPW-169508 - fix unit tests
https://jira.corp.adobe.com/browse/MWPW-169509 - fix escape literals for rich text values


io/studio/package.json
```diff
@@ -0,0 +1,32 @@
+{
+  "name": "MerchAtScaleStudio",
+  "version": "0.1.0",
```
npeltier, line 3: 0.0.1 :)

npeltier (approved):
thanks a lot @markovicpredrag1209! beside healthcheck removal (it's meant to be HC for card live), i'm all good.

3ch023 (changes requested):
@markovicpredrag1209 pls remove duplicate 'ost-products' under www

npeltier:
@3ch023 shouldn't we remove it _after_ we switch consumption of the service? i know it's deployed, but in case a bug arrives?

3ch023:
> @3ch023 shouldn't we remove it _after_ we switch consumption of the service? i know it's deployed, but in case a bug arrives?

now or later we anyway have to move, i think we can move now, worst case we can always redeploy to old location 'manually'

PR #191, opened 2025-03-12
MWPW-168729: [3-in-1] Audit to find all the iframes used in Milo

mirafedas:
I slightly refactored the audit script and added an option to set the audit target using the -t parameter.

By default, the target is OST, so you don’t need to specify -t. But if you set -t modal, the script will look for links where the href contains either mini-plans or modals-content-rich. When using this target, searching by a specific phrase isn’t supported (it is not needed for now)

Resolves [MWPW-168729](https://jira.corp.adobe.com/browse/MWPW-168729)

Test URLs:
- After: https://mwpw-168729--mas--adobecom.aem.live/studio.html


npeltier (approved):
lgtm maybe you could also:
- move /scripts/gen-locale.mjs -> /scripts/content/gen-locale.mjs
- move /ost-audit -> /scripts/audit
?

PR #193, opened 2025-03-12
MWPW-169619: fix stale IMS session issue

yesil:
With this PR, we force IMS to refresh token, if the IMS session is stale, it will error out and the user will be redirected to IMS sign-in flow.

Resolve: MWPW-169619

Test URLs:
- After: https://mwpw-169619--mas--adobecom.aem.live/studio.html


Axelcureno (approved):

npeltier (approved):
lgtm

PR #195, opened 2025-03-13
MWPW-164491 [Plans] Updates to Individuals wide key apps merch-card

bozojovicic:
Authoring of plans wide and super-wide cards and authoring of included components.
Ideally this should be tested together with Angelo's changes for card collections MWPW-165461
since the (super-)wide card will be properly displayed only inside of the card collection.

Resolve [MWPW-164491](https://jira.corp.adobe.com/browse/MWPW-164491)

Test url : mwpw-164491--mas--adobecom.hlx.live/studio.html?milolibs=mwpw164491wide--milo--bozojovicic#path=sandbox&page=content&query=167a3618-d5fd-42c3-a2ff-9cb7bc601638

Test URLs:
- After: https://mwpw-164491--mas--adobecom.aem.live


yesil:
mnemonics  should come before what is included in the form.

yesil:
tests are failing

bozojovicic:
> mnemonics should come before what is included in the form.

@yesil right! fixed. unit tests are also green now.

studio/style.css
```diff
@@ -273,6 +273,26 @@ merch-card .overlay {
     height: 100%;
 }
 
+merch-card[variant='plans'][size='wide'] {
```
yesil, line 276: can you move merch-card styles in merch-card.css?  I think it is time to do it.

studio/src/editors/merch-card-editor.js
```diff
@@ -51,6 +53,37 @@ class MerchCardEditor extends LitElement {
...
+
+        const parser = new DOMParser();
+        const doc = parser.parseFromString(whatsIncludedHtml, 'text/html');
+        return doc.querySelector('merch-whats-included');
+    }
+
+    get whatsIncluded() {
+        const label = this.whatsIncludedElement?.querySelector('[slot="heading"]')?.textContent || '';
+        const values = [];
+        this.whatsIncludedElement?.querySelectorAll('merch-mnemonic-list').forEach((listEl) => {
```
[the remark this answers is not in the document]
bozojovicic: @yesil Unfortunately merch-mnemonic-list receives the whole content as child elements like this 
```
<merch-mnemonic-list>
    <div slot="icon">
        <merch-icon size="s" src="/content/dam/shared/images/product-icons/svg/photoshop.svg" alt="Photoshop"></merch-icon>
    </div>
    <p slot="description">
        <strong>Photoshop</strong>
    </p>
</merch-mnemonic-list>
```
 so I have to use querySelector() to get what I need. There is no data model that I can expose. 
I don't know why it is done like that and not with properties like
`<merch-mnemonic-list title="Photoshop" icon="/content/dam/shared/images/product-icons/svg/photoshop.svg" />`

studio/style.css
bozojovicic: I added `merch-card {width: px}` for wide and super-wide cards. I had to do this here because in mas studio we don't use grids so the wide cards cannot span across multiple columns since there are no columns. So I had to simulate wide and super-wide cards with fixed widths. Since we need this only in mas studio I still don't think we should move it to merch-card.css
We discussed about this in fishbags and I think @npeltier said that all this will be redesigned in mas studio and for now I can simply set these fixed widths.

yesil: what I meant is that, we can have merch-card.css in mas studio, since style.css is growing big and maybe it would be easier to keep merch card stying for MAS studio in a separate file.

studio/src/editors/merch-card-editor.js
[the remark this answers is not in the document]
yesil: Since we will author merch-cards in MAS Studio, we are not subject to the limitation that were set in the Milo block libs/blocks/merch-mnemonic-list/merch-mnemonic-list.js or libs/blocks/merch-whats-included/merch-whats-included.js.
All the data you need is likely available in the fragment JSON.

studio/style.css
bozojovicic: ahaaaa :) done

studio/src/editors/merch-card-editor.js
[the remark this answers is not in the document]
bozojovicic: > All the data you need is likely available in the fragment JSON.

Yes, it is, but as piece of HTML, not as some json string that I can easily parse.

We get that data as HTML (one example above) which I need to transform to DOM element and then use `querySelector()` to find the pieces of info that I need.
We do that with all web components but other web components are much better designed, all values are in attributes so I just take the value from the right attribute. In this case, instead of attributes, `merch-whats-included` and `merch-mnemonic-list` get the values as a chunk of nested HTML elements which I need to parse with `querySelector()`

yesil (approved):

3ch023 (approved):

afmicka:
@3ch023 we probably want milo [PR](https://github.com/adobecom/milo/pull/3805) to be merged first. Lets not merge this one before even though verified

afmicka:
milo PR has been merged to main. Will merge this one after the tests

PR #201, opened 2025-03-17
MWPW-169150 [Nala]: Add tests for checkout parameters

afmicka:
Add Nala tests to validate editing checkout parameters

Resolve [MWPW-169150](https://jira.corp.adobe.com/browse/MWPW-169150)

Test URLs:
- After: https://MWPW-169150--mas--adobecom.aem.live


Roycethan (changes requested):
@afmicka Can you plz split and compare as list items for each parameter ? Since parameter list supported is more and we may need to add more as when needed in future and they can break at times. Rest look good, Thank you

afmicka:
> @afmicka Can you plz split and compare as list items for each parameter ? Since parameter list supported is more and we may need to add more as when needed in future and they can break at times. Rest look good, Thank you

@Roycethan Sure. Done. We can add more any time

Roycethan (approved):

markovicpredrag1209 (approved):

PR #215, opened 2025-03-31
MWPW-170385: io endpoint updates

yesil:
collection transformer removed.

Resolve [MWPW-170385](https://jira.corp.adobe.com/browse/MWPW-170385)

Test URLs:
- After: https://mwpw-170385--mas--adobecom.aem.live/studio.html


io/www/src/fragment/replace.js
```diff
@@ -76,14 +76,14 @@ function replaceValues(input, dictionary, calls) {
...
+    let body = context.body;
+    let bodyString = JSON.stringify(body);
+    if (bodyString.match(PH_REGEXP)) {
         const dictionary = await getDictionary(context);
         if (dictionary && Object.keys(dictionary).length > 0) {
-            fieldsString = replaceValues(fieldsString, dictionary, []);
+            bodyString = replaceValues(bodyString, dictionary, []);
             try {
-                body.fields = JSON.parse(fieldsString);
+                body = JSON.parse(bodyString);
```
npeltier, line 86: Why changing this here? as you already know i'm not a huge fan in mixing up fixes for history management. And typically here i'm not sure: do we need PH management outside of fields inputs? what is the use case?

package.json
```diff
@@ -21,7 +21,7 @@
     },
     "dependencies": {
         "@adobecom/milo": "github:adobecom/milo#stage",
-        "axios": "^1.7.4",
```
st-angelo-adobe, line 24: Looks like the nala tests are failing because they can't find `axios`

io/www/src/fragment/paths.js
```diff
@@ -20,8 +20,7 @@ function odinId(id) {
  * @returns full fetchable path to the fragment references
  */
 function odinReferences(id, allHydrated = false) {
-    /* istanbul ignore next */
-    return `${odinId(id)}/variations/master/references${allHydrated ? '?references=all-hydrated' : ''}`;
+    return `${odinId(id)}${allHydrated ? '?references=all-hydrated' : ''}`;
```
npeltier, line 23: for a start, just checking: have you tested this (removal of master variation specification) works with actual usages of it: 
- placeholders/index request,
- collection request (i guess yes :))

if yes, i think it's better now, as we may never use CF variation mechanism anyhow

npeltier (changes requested):
please also remove studio duplication of it as well (no unit test to manage)

io/www/src/fragment/replace.js
yesil: yes, in the references, there are underneath fields object.

io/www/src/fragment/paths.js
yesil: I didn't specifically test removing master.  But it works with the collection request. Happy to discuss the variation use case if you think we might use it at all.

npeltier: i'm good in the scope of that request

package.json
yesil: I reverted it.

yesil:
>please also remove studio duplication of it as well (no unit test to manage)
@npeltier  I removed collection.js from studio, but did not update everything. Before doing this, I would like confirm if we can or not move placeholder processing to client side/studio in which case we could remove the io/studio completely.

3ch023 (approved):

st-angelo-adobe (approved):

PR #218, opened 2025-04-01
MWPW-170690: Adding sr-only css for rte-field.

seanchoi-dev:
Resolve: [MWPW-170690](https://jira.corp.adobe.com/browse/MWPW-170690)

Test URLs:
- After: https://MWPW-170690--mas--adobecom.aem.live

### DIFF
--- a/io/www/src/fragment/common.js
+++ b/io/www/src/fragment/common.js
@@ -12,6 +12,12 @@ function logError(message, context) {
     console.error(`${logPrefix(context, 'error')} ${message}`);
 }
 
+function logDebug(getMessage, context) {
+    if (context.debugLogs) {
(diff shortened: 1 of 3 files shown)


Axelcureno (approved):
lgtm, I actually wasn't aware this is good practice for styling sr-only. Nice!

seanchoi-dev:
> lgtm, I actually wasn't aware this is good practice for styling sr-only. Nice!

Since this is just for editor, we may just simply do `display: none` but generally, sr-only is meant to be readable from screen reader, so that is kinda standard css.

3ch023 (approved):

3ch023 (approved):

PR #220, opened 2025-04-02
MWPW-169084: Part 2 & 3: Update the USE text and same text to be card title on editor.

seanchoi-dev:
Part 2 & 3: Update the USE text and same text to be card title on editor.

Resolve: [MWPW-169084](https://jira.corp.adobe.com/browse/MWPW-169084)

Test URLs:
- After: https://MWPW-169084--mas--adobecom.aem.live


3ch023 (changes requested):
hi @seanchoi-dev looks good, could you please add a nala test for it?

studio/src/editor-panel.js
```diff
@@ -660,7 +668,13 @@ export default class EditorPanel extends LitElement {
         }
         return html`
             <div id="editor">
-                ${this.fragmentEditorToolbar} ${editor}
+                ${this.fragmentEditorToolbar}
+                <sp-divider size="s"></sp-divider>
+                <div>
+                    <p>${this.generateCodeToUse()[2]}</p>
```
yesil, line 674: this expression qualifies to be in a getter.

studio/src/editor-panel.js
```diff
@@ -183,20 +183,27 @@ export default class EditorPanel extends LitElement {
     getFragmentPartsToUse() {
         let fragmentParts = '';
         let title = '';
-        const surface = Store.search.value.path.toUpperCase();
+        const surface = Store.search.value.path?.toUpperCase();
```
yesil, line 186: I propose to add a getter `Store.surface` with the above expression.

yesil (commented):
Also please add a test to validate your changes in studio/test/editor-panel.test.html

npeltier (approved):
please add  product: this.fragment?.getTagTitle('promo') at the end of the string, that can obviously be empty (this tag namespace does not yet exist, but we'll need it). If too complicated, or not working without it, it's fine.

studio/src/editor-panel.js
seanchoi-dev: Store isn't a class but just an object. I am not sure if setting a getter to Store is a good idea.

afmicka:
@seanchoi-dev @3ch023 @npeltier @yesil these changes are not incorporating locale info. I see discussions around it in jira description but could someone tell me what was the conclusion at the end?

afmicka:
We can have a follow-up jira if there are further requirements...

PR #222, opened 2025-04-04
MWPW-170897: Unordered list action for rte-field

st-angelo-adobe:
Resolves: [MWPW-170897](https://jira.corp.adobe.com/browse/MWPW-170897)

Test URLs:
- After: https://mwpw-170897--mas--adobecom.aem.live


3ch023 (approved):

3ch023:
@st-angelo-adobe you can provide https://mwpw-170897--mas--adobecom.aem.live/ instead of https://mwpw-170897--mas--adobecom.aem.live/studio.html in Before/After PR description. this will ensure psi green

bozojovicic (approved):

mirafedas (approved):

npeltier (approved):

st-angelo-adobe:
@Roycethan I implemented your suggestion.
I'll just say that there may be some niche scenarios where the behavior of the list action might not be exactly what we want (as in, sometimes you get two consecutive lists instead of merging them). But this is the default behaviors of these actions in `prosemirror`, I've checked with some demo implementations too, and they all behave the same.

If we want some refinements we could do that separately.
cc: @npeltier @3ch023 @yesil

Roycethan:
> MWPW-170897

Thanks, fixed

PR #227, opened 2025-04-08
MWPW-170896: Heading for rte-field

st-angelo-adobe:
Resolves: [MWPW-170896](https://jira.corp.adobe.com/browse/MWPW-170896)

Test URLs:
- After: https://mwpw-170896--mas--adobecom.aem.live/studio.html

### DIFF
--- a/studio/src/rte/rte-field.js
+++ b/studio/src/rte/rte-field.js
@@ -232,7 +232,9 @@ class RteField extends LitElement {
                 div.ProseMirror-focused
                     span[is='inline-price'].ProseMirror-selectednode,
                 div.ProseMirror-focused a.ProseMirror-selectednode,
-                div.ProseMirror-focused a.ProseMirror-selectednode {
+                div.ProseMirror-focused a.ProseMirror-selectednode,
+                div.ProseMirror-focused
+                    p[class^='heading-'].ProseMirror-selectednode {
                     outline: 2px dashed var(--spectrum-global-color-blue-500);
                     outline-offset: 2px;
                     border-radius: 16px;
@@ -247,6 +249,18 @@ class RteField extends LitElement {
                     clip: rect(0, 0, 0, 0);
                     white-space: nowrap;
                     border: 0;
+                }
+
+                .heading-icon {
+                    display: flex;
+                    justify-content: center;
+                    align-items: center;
+                    font-weight: bold;
+                    font-size: 20px;
+                }
+
+                p[class^='heading-'] {
+                    font-weight: bold;
                 }
             `,
             prosemirrorStyles,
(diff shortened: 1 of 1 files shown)


bozojovicic (approved):

yesil (commented):
Cloud you please add a test case to studio/test/rte/rte-field.test.html ?

studio/src/rte/rte-field.js
```diff
@@ -323,7 +335,22 @@ class RteField extends LitElement {
...
-        let nodes = basicNodes.addToStart('inlinePrice', {
+
+        let nodes = basicNodes
+            .remove('heading')
+            .addBefore('code_block', 'heading', {
+                content: 'inline*',
+                group: 'block',
+                defining: true,
+                attrs: { class: { default: null } },
+                parseDOM: [{ tag: 'p[class^="heading-"]' }],
```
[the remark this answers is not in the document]
yesil: actually, maybe not, if it works as a node, let's keep it :)

studio/src/rte/rte-field.js
```diff
@@ -326,10 +359,25 @@ class RteField extends LitElement {
...
 
     #initEditorSchema() {
-        const basicNodes = this.list
+        let nodes = this.list
             ? addListNodes(schema.spec.nodes, 'paragraph block*', 'block')
             : schema.spec.nodes;
-        let nodes = basicNodes.addToStart('inlinePrice', {
+
+        if (this.styling) {
+            nodes = nodes.remove('heading').addBefore('code_block', 'heading', {
```
yesil, line 367: do you need to remove headings? I believe this is the true headings (h1-h6) and we don't need to really remove tem. Although we do not let them to be authorable via RTE.
I would call your node type something different than `heading` to avoid collision. 
isn't this supported to be MAS Styling extension?

st-angelo-adobe: Actually I didn't have to yes, especially since I repurposed the node to be more generic, `styling`. So I changed this.

Also, you were right about the `p` tags 😄  I noticed after saving, when I would re-open the editor, the styled tags would just turn into normal `p` tags. That's because I added my custom node AFTER the `paragraph` block, and the simple `p` selector would just catch all my styled elements. 
So I just changed the tag to `span` => this means in Milo I'll have to add a `display: block`, but I think that's not a big deal.

I also merged the [promo-text](https://github.com/adobecom/mas/pull/234) PR into this, just to have all the changes here.

st-angelo-adobe: [Here](https://mwpw-164498--milo--adobecom.aem.page/libs/features/mas/docs/plans.html) is an example of a link styled with `heading-xs`. The last card on the page, the phone link.

yesil: >hey just apply to the whole paragrapth they're part of
I agree, it makes sense, thanks.

yesil (approved):

st-angelo-adobe:
@yesil @bozojovicic I made some changes, may be worth to take another look. Basically instead of a node, there's a `styling` mark, and in the case of headings and promo text, it's being simply rendered with `display: block`. They also override eachother, so you can't stack them. For example, when applying `heading-s`, all the other styling(!) marks in the selection are removed first.
Obviously in the future there might be `styling` marks that aren't blocks, and can be stacked, but for now, this is the logic.

@afmicka unless other comments pop up this can be tested

afmicka:
@st-angelo-adobe If you apply the style to the text, and lets say i clicked the wrong one, so when i apply again the new style to the same text, It duplicates all the fields on the card. Could you look into that?

Screenshot 2025-05-05 at 14 27 42

PR #273, opened 2025-05-27
MWPW-172123 card titles must use H3 tag, for a11y

bozojovicic:
For a11y reasons card titles must be `<h3>` instead of `<P>`.
This is done with Milo PR https://github.com/adobecom/milo/pull/4247
For students cards the secondary headers must use `<h4>`.
These headers are authored in MAS so with this change RTE will generate headers with ARIA attributes like
`<p><span class="heading-xs" role="heading" aria-level="4">Heading XS</span></p>`
For a11y this is the same as `<h4>Heading XS</h4>` with no need to adjust the CSS.

Screenshot 2025-05-27 at 09 23 13

Screenshot 2025-05-27 at 09 20 39

Resolves https://jira.corp.adobe.com/browse/MWPW-172123

Test page `mwpw-172123--mas--adobecom.aem.live/studio.html?milolibs=mwpw172123--milo--bozojovicic#page=content&path=sandbox&query=799924c4-5eca-4c09-a5e7-d58a76a41add`

Test URLs:
- After: https://mwpw-172123--mas--adobecom.aem.live/


Axelcureno (approved):

Roycethan:
@bozojovicic if a perticular style is selected using brush icon in the next visit editor doesn't highlight which style is selected...
for us we can inspect and find out but for GWPs it might be difficult to see:
image
https://mwpw-172123--mas--adobecom.aem.live/studio.html?milolibs=mwpw172123--milo--bozojovicic#page=content&path=sandbox&query=799924c4-5eca-4c09-a5e7-d58a76a41add

Also a if possible a tip saying what style is correponds to H1...H6 would be helpful - Nice to have but not in scope of this PR ..

bozojovicic:
@Roycethan highlights added
Screenshot 2025-06-09 at 12 56 00
Even if you select multiple lines of text where different styles are applied, all of them will be highlighted.
You don't need to select the text. It's enough to position the cursor to the styled text.

PR #289, opened 2025-06-09
MWPW-173772 enable collection placeholders

npeltier:
- add placeholder mapping for collection in settings (for now only plans, but later we'll have values depending on variant / tags),
- replace now consider a context dictionary to be merged with a general one,
- unit tests

Resolves (partially) https://jira.corp.adobe.com/browse/MWPW-173772

Test URLs:
- After: https://MWPW-173772--mas--adobecom.aem.live/


3ch023 (approved):

io/www/src/fragment/settings.js
```diff
@@ -1,3 +1,34 @@
+const COLLECTION_MODEL_ID =
+    'L2NvbmYvbWFzL3NldHRpbmdzL2RhbS9jZm0vbW9kZWxzL2NvbGxlY3Rpb24';
+
+function applyCollectionSettings(context) {
+    context.body.placeholders = {
```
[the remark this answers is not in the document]
npeltier: at this stage it does, yes (the only cases where body does not exist are errors that would have cut the pipeline before)

io/www/src/fragment/settings.js
```diff
@@ -1,3 +1,34 @@
...
+        resultMobileText: '{{coll-result-mobile-text}}',
+        resultsMobileText: '{{coll-results-mobile-text}}',
+        searchResultText: '{{coll-search-result-text}}',
+        searchResultsText: '{{coll-search-results-text}}',
+        searchResultMobileText: '{{coll-search-result-mobile-text}}',
+        searchResultsMobileText: '{{coll-search-results-mobile-text}}',
+        noSearchResultsText: '{{coll-no-search-results-text}}',
+        showMoreText: '{{coll-show-more-text}}',
+    };
+    context.dictionary = {
```
yesil, line 24: can `dictionary` be baked in frontend code? Are they going to be ever updated via authoring somehow?

yesil (approved):

io/www/src/fragment/settings.js
npeltier: i don't think they should, this is dev to dev kind of thing: tech {{ }} placeholder gives tech <span> placeholder that our own code uses.

yesil: above, are you reference to `dictionary` as tech placeholders?

npeltier: yes, because those entries are in dictionnary _before_ we fetch entries from user (odin)

PR #323, opened 2025-07-15
MWPW-176249: Fix missing addon placeholders in the card editor

st-angelo-adobe:
The problem was only the first 50 placeholders were being loaded (I believe that's the default limit). Meaning newer addon placeholders which were not part of the first 50 would not show up, even after getting published.

Resolves https://jira.corp.adobe.com/browse/MWPW-176249

- [ ] C3. Verify all Checks are green (unit tests, nala tests, psi)

Test URLs:

- After: https://mwpw-176249--mas--adobecom.aem.live/


3ch023 (approved):

3ch023:
> ok, this will turn into performance problem (or pagination ticket) quite soon, but i guess we can live with this for now.

right, pagination in sandbox already becomes a problem..

PR #351, opened 2025-09-02
Mwpw-174771: Mnemonics modal experience in Studio

Axelcureno:
Resolves https://jira.corp.adobe.com/browse/MWPW-174771

Test URLs:

- After: https://mwpw-174771--mas--adobecom.aem.live/


npeltier (changes requested):
Looks overall good to me 

1. can you please factorize nala function to do & check stuff with icon, so we don't have those 50 lines duplicated more or less all over the place?
2. when i go to a card, edit it and add a mnemnonic, i don't see the preview updated. It takes a save, refresh to see this icon. I guess this needs a fix

studio/src/fields/mas-mnemonic-modal.js
```diff
@@ -0,0 +1,407 @@
...
+    { id: 'substance-3d-stager', name: 'Substance 3D Stager' },
+    { id: 'substance-3d-painter', name: 'Substance 3D Painter' },
+    { id: 'substance-3d-sampler', name: 'Substance 3D Sampler' },
+    { id: 'substance-3d-designer', name: 'Substance 3D Designer' },
+    { id: 'substance-3d-assets', name: 'Substance 3D Assets' },
+    { id: 'stock', name: 'Stock' },
+    { id: 'fonts', name: 'Fonts' },
+    { id: 'color', name: 'Color' },
+    { id: 'capture', name: 'Capture' },
+    { id: 'incopy', name: 'InCopy' },
```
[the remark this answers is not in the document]
Axelcureno: I have switched the code to pull from AOS and sync with OST. LMK what you think about aos-products.js in my changes.

nala/studio/mnemonic-modal/tests/mnemonic-modal-simple.test.js
```diff
@@ -0,0 +1,88 @@
+import { expect, test } from '@playwright/test';
```
afmicka, line 1: @Axelcureno what is this test file exactly? It does not have corresponding spec files either but rather has cardid in the test file itself which does not follow our structure.

afmicka (commented):
@Axelcureno you will have to either refactor these Nala tests or remove them completely from this PR. We don't have a separate discard test/spec files anymore but rather add the discard step and validation at the end of each edit test. Please check all other tests, refactoring PR has been merged today. There are also changes in imports, before and after hooks, commons, setting the pages... If this PR is urgent, i advise to rather have Nala added as a separate jira/PR

Roycethan (commented):
@Axelcureno  
1) Is this new experience needed for in line mnemonics  of variants fries or simplified express ? they still have older modal... or do we have a different ticket address ?
2) Nala errors must be resolved.
3) The tooltip box is overlaping over the Edit mnemonic icon window 
https://mwpw-174771--mas--adobecom.aem.live/studio.html#page=content&path=sandbox&query=f6defd08-ed85-49fb-83ec-a84ad7298ffa
image
4) URL if its long its cutting off at right browser border- should flow to next line if window width is fixed
image

afmicka (changes requested):
@Axelcureno beside test comments bellow, given there are 2 ways to update the mnemonic on the card, would be good to have a test case for each method to make sure both are working. 

Also, when you choose the icon in the picker, the url in the second tab is not being updated with the new choice. That field is being updated only upon saving. The other way around works like that. If i update the url manually in the second tab, it automatically switches to the picker showing the correct new selection. Btw, should it switch the tab at all? what if the user wants to make additional changes and set the icon manually with all the fields?

Screenshot 2025-10-24 at 13 05 02
Screenshot 2025-10-24 at 13 05 16

cc: @markovicpredrag1209 @3ch023

Axelcureno:
> * https://mwpw-174771--mas--adobecom.aem.live/
I'll take care of all this. 

@afmicka About your last question. I don't see why a user would enter the same URL provided by the icon picker (it's counter intuitive), the whole point of the icon picker is to avoid having to do that, in a way it's a hint for the user saying: "No need to do that anymore, it's way easier now". And what other fields are you referring to? Both tabs only offer alt text and link URL (as current experience), nothing is missing between the two tabs.

nala/studio/commerce/fries/tests/fries_save.test.js
```diff
@@ -74,8 +75,6 @@ test.describe('M@S Studio Commerce Fries card test suite', () => {
 
                 // Validate mnemonic in editor and card
                 test.step('Validation-3: Validate edited mnemonic', async () => {
-                    const iconInput = await editor.iconURL.first();
-                    await expect(iconInput).toHaveValue(data.iconURL.updated);
```
afmicka: please don't just remove the checks in the editor (value check). We need those as well, not just the card preview. Could you please add the corresponding check here with the new mnemonic feature?

afmicka (approved):
thanks

npeltier (approved):

PR #358, opened 2025-09-05
Improve UX for search indicator

yesil:
After the first page is loaded, the main spinner is hidden and another spinner at the bottom of the page appears until the result fully loads.
Users will see the spinner at the bottom only if they scroll down and I think it is a good way to say that more result is coming.

Resolves https://jira.corp.adobe.com/browse/MWPW-180011

- [X] C5: you are ready to do a demo from Test Page in PR (bonus: write a working demo script that you'll use on Thursday, you can eventually put in your PR)
- [X] C.6 read your Jira one more time to validate that you've addressed all AC's and nothing is missing

Test URLs:

- After: https://mwpw-180011--mas--adobecom.aem.live/


st-angelo-adobe (approved):

3ch023 (approved):

yesil:
@Roycethan I think there is a misunderstanding, by first page, I'm referring to the first page of the search result which is paginated. Every time you go to content page, it will perform a search query which will yield a multi page result. As soon as the first page is shown to the user, the spinner will hide, and instead you will see it at the bottom of the result until the rest of the pages are loaded.

PR #362, opened 2025-09-11
MWPW-179803: Content browsing enhancement

st-angelo-adobe:
Resolves https://jira.corp.adobe.com/browse/MWPW-179803

Test URLs:

- After: https://mwpw-179803--mas--adobecom.aem.live/


3ch023:
unfortunately for now postponed

PR #368, opened 2025-09-16
MWPW-177244: Copying a fragment from one surface folder to another

Axelcureno:
Let's the user copy fragments into NALA or Sandbox folders for speeding development.

Resolves https://jira.corp.adobe.com/browse/MWPW-177244

Test URLs:

- After: https://mwpw-177244--mas--adobecom.aem.live/


nala/studio/folder-copy/folder-copy-ui.test.js
```diff
@@ -0,0 +1,126 @@
...
+            expect(nameInputVisible).not.toBeNull();
+
+            // Check that dialog wrapper exists with confirm/cancel buttons
+            const dialogWrapper = await studio.getCopyDialogElement('sp-dialog-wrapper');
+            expect(dialogWrapper).not.toBeNull();
+
+            // Cancel the dialog using the cancel event on the dialog wrapper
+            await page.evaluate(() => {
+                const toolbar = document.querySelector('mas-toolbar');
+                let dialog = null;
```
mirafedas, line 91: Nit: we can shorten this a little:
```
const dialog = toolbar?.shadowRoot?.querySelector('mas-copy-dialog') 
?? document.querySelector('mas-copy-dialog');
if (dialog?.shadowRoot) { ... }
```

mirafedas (approved):

studio/src/mas-copy-dialog.js
```diff
@@ -0,0 +1,425 @@
...
+        );
+
+        if (!copiedFragment) {
+            throw new Error('Copy operation completed but could not retrieve copied fragment');
+        }
+
+        return copiedFragment;
+    }
+
+    async handleSubmit() {
```
yesil: during the copy, can you unset originalId, it might create invalid cross surface references which can be hard to troubleshoot.

yesil (changes requested):

studio/src/mas-copy-dialog.js
```diff
@@ -335,7 +333,7 @@ export class MasCopyDialog extends LitElement {
     close() {
         this.dialog.open = false;
         this.dispatchEvent(
-            new CustomEvent('dialog-closed', {
+            new CustomEvent('cancel', {
```
yesil: why do you need to dispatch this event?

npeltier:
@afmicka @Axelcureno 
1. yes should be the name, not the title as it seems to be now
2. i guess either is fine for now
3. i guess that's fine as well (maybe a problem later with promotions)
4. would be great to have this feature for promotions, yes

npeltier:
checked with @afmicka and the problematic part is the modal when copying, we should not propose another name or name field at all, just title. And probably not show the path either as it's not things we are used to expose to users.
I remember having asking you something like this as a potential workaround with variations, but it appears we don't need it anymore

PR #370, opened 2025-09-18
MWPW-180733 [mas/io] introduce init steps

npeltier:
- migrated from commonJS to ES6 as later was becoming complicated,
- every  now is an object with process & name, and sometimes init method,
- pipeline & fragment client are checking at beginning if they can call init for each transformer,
- replace has an init method that _if_ we have a dictionary id already stocked, triggers dictionary fetch

Resolves https://jira.corp.adobe.com/browse/MWPW-180733 

Test URLs:

-


io/www/src/fragment/replace.js
```diff
@@ -72,19 +74,33 @@ function replaceValues(input, dictionary, calls) {
     return replaced;
 }
 
+async function init(context) {
+    return new Promise(async (resolve) => {
+        //we fetch dictionary at this stage only if id has already been cached
+        //because we can't now surface of fragment *before* first fetch
```
3ch023, line 80: typo: we can't know

io/www/src/fragment-client.js
```diff
@@ -31,13 +33,21 @@ async function previewFragment(id, options) {
         api_key: 'n/a',
         locale,
     };
-    for (const transformer of [fetchFragment, translate, settings, replace, corrector]) {
+
+    for (const transformer of PIPELINE) {
+        if (transformer.init) {
+            context.transformer = `${transformer.name}-init`;
+            context.promises = context.promises || {};
+            context.promises[transformer.name] = transformer.init(context);
```
3ch023, line 41: do we await these promises somewhere?

io/www/src/fragment/replace.js
```diff
@@ -72,19 +74,33 @@ function replaceValues(input, dictionary, calls) {
...
+        const dictionary = context.dictionaryId ? await getDictionary(context) : null;
```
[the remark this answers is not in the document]
npeltier: if you look at getDictionary it checks a dictionary presence from the promises, if not compute it, so in that case it will pull id, then dictionary, in sequence, so that request will be slow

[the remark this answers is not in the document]
npeltier: it will! see l91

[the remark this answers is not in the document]
npeltier: and l95 :-)

io/www/src/fragment/pipeline.js
```diff
@@ -115,9 +108,19 @@ async function mainProcess(context) {
         context = { ...context, translatedId, dictionaryId };
     }
 
-    for (const transformer of [fetchFragment, translate, settings, replace, wcs, corrector]) {
-        /* istanbul ignore next */
+    for (const transformer of PIPELINE) {
+        if (transformer.init) {
+            const initContext = structuredClone(context);
```
3ch023, line 113: we can add a comment, smthg like
'structuredClone  to prevent init from modifying context'

io/www/src/fragment/replace.js
```diff
@@ -72,19 +74,33 @@ function replaceValues(input, dictionary, calls) {
     return replaced;
 }
 
+async function init(context) {
```
3ch023, line 77: seems like it doesnt need to be async

io/www/src/fragment/replace.js
3ch023:
```suggestion
        // if dictionaryId is present in cache - early load dictionary
        // if nothing in cache - disctionaryId and disctionary itself will be loaded later, during process
```

npeltier: typo: dictionaryId :-p

io/www/src/fragment/common.js
```diff
@@ -1,8 +1,8 @@
-const fetch = require('node-fetch');
-const { transformBody } = require('./odinSchemaTransform.js');
+import fetch from 'node-fetch';
```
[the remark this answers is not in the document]
npeltier: well, does not work :-) leaving it, and will figure this out later

PR #383, opened 2025-10-02
MWPW-180989: add Odin author request counter

afmicka:
Resolves https://jira.corp.adobe.com/browse/MWPW-180989

Adds counter for ODIN PROD author requests for the Nala execution (with possibility to add more requests for tracking)

Screenshot 2025-10-02 at 22 30 50

Test URLs:

- After: https://mwpw-990002--mas--adobecom.aem.live/


afmicka:
> looks great to me!
> 
> 2 things:
> 
> * i think we could have a "masTest" object that is an extension of @playwright/test, and in which you could do all those "common" operations like counter init & dump,
> * i would have set aemRequestCounter with a list of urls. I know for now we just care about odin author, but this feature could be helpful for other stuff (typically e.g. wcs requests, or mas/io requests, ...)

@npeltier mas-test took a seriously wrong turn because of the hook handling (should have done it in the refactoring PR :) ) but this should do it now. We now also have possibility to list other request urls too. And the report lists the method's counters too for each service/url tracked
Screenshot 2025-10-02 at 16 40 48

PR #446, opened 2025-11-25
MWPW-172706 - create a CMR for each PR to the main branch

tusc-adbe:
~Do not merge.  Testing functionality.~
~**Before merging:**~
* ~Update `branches` to `main`~ - DONE

**Overview**
* This workflow will create CMRs for any changes to M@S Studio or M@S IO
* The workflow runs when a PR for the main branch is opened or closed, and contains changes under io/ or studio/
* On PR open
  * The workflow determines exactly what changed and create a CMR for M@S Studio, M@S IO, or both.
    * Transaction numbers for those CMRs are written to the PR comments.  The workflow will reference these transaction numbers when closing the CMRs.
  * A notification will be sent to the Slack #merch-at-scale channel
* On PR close
  * The workflow will check PR comments to determine whether any CMRs were opened, and will update and close them
    * If the PR was merged, the CMRs will be closed as Successful; if the PR was closed without merging the CMRs will be Cancelled.
  * A notification will be sent to the Slack #merch-at-scale channel (only if the PR was merged)
* Future enhancements
  * Expand this to include other paths (this may require creating and onboarding new services in the Service Registry
  * Refactor this workflow so all actions take place on PR close.  This will reduce CMR noise and reduce the time between opening a PR and closing it.
  * These will be done in separate JIRA tickets and PRs


.github/workflows/servicenow.yaml
```diff
@@ -0,0 +1,185 @@
...
+      - name: Detect changed paths and set SNOW Instance ID
+        id: set-instance-id
+        run: |
+          # Get the list of changed files in the PR
+          CHANGED_FILES=$(git diff --name-only ${{ github.event.pull_request.base.sha }} ${{ github.event.pull_request.head.sha }})
+          
+          echo "Changed files:"
+          echo "$CHANGED_FILES"
+          
+          # Check if any files match the io/** pattern
```
3ch023: can we add here /io OR /web-components?

.github/workflows/servicenow.yaml
```diff
@@ -0,0 +1,185 @@
+# This workflow will install Python dependencies, run CMR creation in ServiceNow
+# For more information see: https://docs.github.com/en/actions/automating-builds-and-tests/building-and-testing-python
+
+name: Create CMR in ServiceNow
+
+on:
+  pull_request:
+    types:
+      - opened
+      - closed
```
3ch023: shouldn't it trigger on merge?

.github/workflows/servicenow.yaml
```diff
@@ -0,0 +1,185 @@
...
+    branches:
+      - MWPW-172706
```
3ch023: i guess it will be updated later :)

3ch023 (commented):
Hi @tusc-adbe , I checked it a bit and left some minor comments.
could you cleanup PR and make it ready for merging? Nico can give it a look and we merge afterwards

.github/workflows/servicenow.yaml
tusc-adbe: No.  The CMR is opened when the PR is opened.  If we only update and close the CMR on `merge` then we miss the case where a PR is closed without merging, and we'll end up with a lot of orphaned CMRs.  There's logic later in the workflow to determine whether or not it was actually merged.

.github/workflows/servicenow.yaml
tusc-adbe: Right, updating this to `main` is the *absolute* last thing to update before merging this PR.

.github/workflows/servicenow.yaml
tusc-adbe: Yes and no.  :)  This workflow can be expanded to additional paths but unfortunately it's not as simple as dropping in a new path here.  There's more logic below that needs to be updated, and if the new path relates to a separate service in the Service Registry, then that service needs to be created and onboarded first.

.github/workflows/servicenow.yaml
tusc-adbe: No.  CMRs will be created when the PR is opened.  When a PR is closed, we need to close those CMRs.  If we use `merge` here, then we'll miss closing CMRs for PRs that are closed without merging, but `close` will catch both merged PRs and PRs that were closed without merging.  There's logic later in the workflow to determine whether the PR was merged: if it was merged then we close the CMR as successful, if it was closed without merging then we cancel the CMR.

tusc-adbe:
Thanks @3ch023 !

I replied to your questions, added an overview in both the PR and in the main yaml file, and cleaned up all the temp stuff so it's almost ready for a final merge.

The last thing to do before merging is to update `branches` to `main` in servicenow.yaml.

cc: @npeltier

.github/workflows/servicenow.yaml
npeltier: yes, i guess we will need new service registrations for web-component & probably ost too

.github/workflows/servicenow.py
```diff
@@ -0,0 +1,328 @@
...
+      output_file.write(f"change_id={cmr_id}\n")
+      output_file.close()
+
+    print("Setting Actual Maintenance Time Windows for CMR...")
+    actual_start_time = int((datetime.datetime.now() - datetime.timedelta(seconds = 10)).timestamp())
+    actual_end_time = int(datetime.datetime.now().timestamp())
+
+    print("Closing CMR in ServiceNow...")
+
+    pr_merged_str = os.environ.get('PR_MERGED', 'true')
```
npeltier, line 282: why is default value true?

npeltier (approved):
lgtm

.github/workflows/servicenow.py
tusc-adbe: Just to give it a default.  :) I suppose it could be removed if you're concerned about it, but per the code the variable will always be defined when the workflow executes.

npeltier: fine by me!

afmicka (approved):

PR #672, opened 2026-03-11
MWPW-189852: fix locale picker displaying wrong locale for variations

Axelcureno:
When editing a regional variation (e.g. en_GB), the locale picker in the top nav was showing the global store locale instead of the fragment's actual locale. This happened because the picker always read from `Store.localeOrRegion()` and never checked the fragment being edited.

This fix passes `currentFragmentLocale` (derived from the fragment path) to the locale picker, so it correctly reflects the locale of the fragment in view. The picker is also set to disabled when viewing a variation to prevent locale changes that would conflict with the fragment context. A `getLocaleByCode` fallback was added so non-surface locales (like regional variations) can still resolve a display name.

Resolves https://jira.corp.adobe.com/browse/MWPW-189852

- [ ] C5: you are ready to do a demo from Test Page in PR (bonus: write a working demo script that you'll use on Thursday, you can eventually put in your PR

Test URLs:

- Before: https://main--mas--adobecom.aem.live/studio.html#content-type=merch-card&fragmentId=6d885870-6d5a-4fc9-a460-11f808d5b51b&locale=en_IN&page=fragment-editor&path=express
- After: https://mwpw-189852--mas--adobecom.aem.live/studio.html#content-type=merch-card&fragmentId=6d885870-6d5a-4fc9-a460-11f808d5b51b&locale=en_IN&page=fragment-editor&path=express
EOF
)


yesil (changes requested):
The locale picker should reflect `Store.localeOrRegion()` directly — the fix should be in the store, not in the template.

The root cause is the guard in `#updateLocaleIfNeeded` (line 635): it only syncs the region when `Store.filters.value.locale === 'en_US'`, which blocks the update when navigating directly with a locale param like `locale=en_GB`. The `#initializeFromRepository` path (line 781) already does an unconditional region sync for variations — `#initializeFromCachedStore` should do the same.

Fixing the store means:
- The locale picker works via its existing subscription — no template override or `getLocaleByCode` fallback needed
- All other `Store.localeOrRegion()` consumers (commerce service, dictionary path, placeholders, preview) also get the correct locale

yesil (changes requested):
Left some comments.

Andrei4226 (commented):
Nit: the `Store.filters `and `Store.search` subscriptions share identical logic.
Optionally, we could extract a method to avoid the duplication.
It’s not mandatory, so I’m fine with the current implementation.

studio/src/mas-locale-picker.js
```diff
@@ -359,17 +367,19 @@ export class MasLocalePicker extends LitElement {
...
         }
         if (changedProperties.has('tempSelectedLocales')) {
             this.#tempSelectedSet = new Set(this.tempSelectedLocales);
         }
     }
 
-    /** can only be one of default languages, not regional ones */
     get currentLocale() {
-        return this.getLocales().find((l) => getLocaleCode(l) === this.locale) || getDefaultLocale(this.surface, this.locale);
+        const surfaceLocale = this.getLocales().find((l) => getLocaleCode(l) === this.locale);
```
yesil, line 365: while loading the fragment, it shows EN(ID) at the top for a short moment and then switches back to correct EN(US).
I wonder if this change has other impacts as well.
Ideally this regression should be fixed before PR merge.

studio/src/mas-top-nav.js
```diff
@@ -167,6 +168,17 @@ class MasTopNav extends LitElement {
         return this.page.value === PAGE_NAMES.SETTINGS_EDITOR;
     }
 
+    get topNavLocale() {
```
yesil, line 171: is this method still needed?

studio/src/mas-locale-picker.js
Axelcureno: Fixed. Root cause was the two  calls in  — they fired immediately on subscribe (before the parent's attribute binding arrived), seeding the wrong locale. Removed both subscriptions entirely; the locale is now read once at connect time via `Store.localeOrRegion()` with `??=`. All five parent components already pass `locale` as a reactive attribute, so the subscriptions were redundant. 30/30 tests pass.

studio/src/mas-top-nav.js
Axelcureno: No longer needed — removed. The subscriptions that relied on it are gone, so disconnectedCallback is now just the super call.

Axelcureno:
@Andrei4226 the duplication you flagged is gone — both subscriptions were removed entirely. The locale picker now reads from Store.localeOrRegion() once at connect time instead of subscribing reactively. All parent components already pass locale as an attribute, so the subscriptions were doing redundant work and causing the flicker.

@yesil summary of what changed in the latest commit:

1. **Locale picker flicker (EN(ID)→EN(US))**: Removed the two Store.subscribe() calls from connectedCallback. They were firing immediately on subscribe, before the parent attribute binding arrived — seeding the wrong locale. With the subscriptions gone, the picker reads the initial locale once via ??= and stays driven by the locale attribute from the top-nav.

2. **Back-navigation not restoring locale**: Removed the stale locale re-injection block from the router hashchange handler (the block that re-read Store.filters.value.locale when the hash had no locale param). syncStoreFromHash already handles restoring locale from the hash correctly — that block was overriding it with stale Store state on every back-navigation.

3. **Not translated panel flash on back-nav**: Added a guard to missingVariationState — when fragmentId (the target) differs from fragment.id (what is currently loaded), a reload is already in flight, so return null instead of evaluating the locale mismatch. This mirrors what forward navigation already does (it clears inEdit before changing locale, so this.fragment is null and the check short-circuits naturally).

142 tests pass across locale picker, top-nav, and fragment editor.

yesil (approved):

Andrei4226 (approved):

Blainegunn (approved):

Roycethan (approved):

PR #736, opened 2026-04-02
MWPW-188624: support promo override in grouped variations

yesil:
Resolves https://jira.corp.adobe.com/browse/MWPW-188624

Adds grouped-variation preview locale selection in Studio and introduces a per-fragment `compatVersion` gate so the new fragment-level promo-code behavior only activates on cards authored after this change.

### Studio
- Preview panel shows a locale dropdown for grouped variations, populated from `pznTags`.
- Preview card and localized prices re-render when the preview locale changes.
- Preview locale override lives on the merch-card editor instead of global store state.
- Editor applies the card `promoCode` as the default promotion code for price rendering and passes it to OST as a contextual `promotionCode`, separate from placeholder-level `storedPromoOverride`.
- Rendered prices refresh when the `promoCode` field changes.
- RTE external sync fix: restore flows no longer immediately recreate footer overrides.
- On open, the editor stamps `compatVersion = 1` onto the fragment so re-save migrates it forward.

## 🧪 Nala E2E Tests

### To run Nala tests again:

- Remove the `run nala` label

## Test URLs:

- After: https://mwpw-188624--mas--adobecom.aem.live/

## Test fragments:

**parent**: https://mwpw-188624--mas--adobecom.aem.live/studio.html#fragmentId=37620c59-39b7-4340-9042-792aabad8206&page=fragment-editor&path=sandbox

**grouped variation**: https://mwpw-188624--mas--adobecom.aem.live/studio.html#fragmentId=a010c8e9-683c-4c61-ba09-85ad1337c51f&page=fragment-editor&path=sandbox

### compatVersion:1
global promo applied
https://main--da-cc--adobecom.aem.page/drafts/ilyas/grouped-variation?maslibs=mwpw-188624--mas--adobecom&akamaiLocale=IT&mas-geo-detection=on

### compatVersion:0
global promo skipped
https://main--da-cc--adobecom.aem.page/drafts/ilyas/grouped-variation?maslibs=mwpw-188624--mas--adobecom&akamaiLocale=FR&mas-geo-detection=on


honstar:
There are test issues, please check!

honstar:
There are conflicts with files in `web-components/dist`, please resolve.

studio/src/editors/merch-card-editor.js
```diff
@@ -434,17 +519,41 @@ class MerchCardEditor extends LitElement {
...
+    #registerCommercePriceProviders(service = document.querySelector('mas-commerce-service')) {
+        registerGroupedPreviewLocaleProvider(service);
+        registerEditorPromoCodeProvider(service);
+    }
+
+    refreshRenderedPrices() {
+        document.querySelector('mas-commerce-service')?.refreshOffers?.();
+        this.querySelectorAll('rte-field').forEach((field) => {
+            field.shadowRoot?.querySelectorAll(INLINE_PRICE_SELECTOR).forEach((price) => {
+                price.requestUpdate(true);
```
honstar, line 545: As per https://lit.dev/docs/v1/components/lifecycle/#requestupdate, passing a boolean argument is not supported and would be treated as a property name. `true` should be removed here.

studio/src/editors/merch-card-editor.js
```diff
@@ -434,17 +519,41 @@ class MerchCardEditor extends LitElement {
...
     }
 
     disconnectedCallback() {
         super.disconnectedCallback();
+        document.removeEventListener(EVENT_COMMERCE_READY, this.#handleCommerceReady);
         this.lastMnemonicState = null;
     }
 
+    #handleCommerceReady = (event) => {
+        this.#registerCommercePriceProviders(event.detail || event.target);
```
honstar: `event.target` for a `wcms:commerce:ready` event dispatched on `document` is `document`, not the commerce service element. So this fallback would pass document to the registration function, which would silently fail the `service?.providers?.price` guard. The fallback is effectively dead code.
The function already has `document.querySelector('mas-commerce-service')` as its default arg, so the `event.target` path adds noise without benefit.

studio/src/editors/merch-card-editor.js
```diff
@@ -611,6 +720,19 @@ class MerchCardEditor extends LitElement {
 
     async updated(changedProperties) {
         super.updated(changedProperties);
+        if (changedProperties.has('previewLocaleOverride')) {
+            const oldValue = changedProperties.get('previewLocaleOverride');
+            if (oldValue !== this.previewLocaleOverride) {
```
honstar: LitElement only populates `changedProperties` with properties that actually changed, so this check is always true. It's harmless but misleading.

studio/src/editors/merch-card-editor.js
```diff
@@ -434,17 +519,41 @@ class MerchCardEditor extends LitElement {
...
+            });
+        });
+    }
+
     #cachedGlobalDefaults = null;
 
     willUpdate(changedProperties) {
         this.#cachedGlobalDefaults = null;
+        if (this.fragmentStore?.get()) {
+            this.#syncGroupedPreviewLocale();
```
honstar, line 555: This recomputes `groupedPreviewLocales` (iterating + mapping pznTags) on every render cycle. Should at minimum be guarded with a check for whether fragment data actually changed, e.g. only when `fragmentStore` or pznTags-bearing fields are in `changedProperties`.

honstar (changes requested):

studio/src/editors/merch-card-editor.js
yesil: The element being called here is `inline-price`, which is a custom element — not a LitElement. Its `requestUpdate(true)` API accepts a boolean argument as intended, so this is correct usage for that element's implementation.

studio/src/editors/merch-card-editor.js
yesil: The reactive controller calls `requestUpdate()` with no arguments when store data changes, so `changedProperties` is empty in those cases — there's no property name to guard on. Guarding on `changedProperties.has('fragmentStore')` would only fire when the store reference changes, not when the data inside it changes (which is the case we care about). Given this, the computation is lightweight enough that running it every render is the most pragmatic approach. Proposing to keep it as-is.

studio/src/editors/merch-card-editor.js
yesil: done.

studio/src/editors/merch-card-editor.js
yesil: done

mirafedas (approved):

studio/src/editors/merch-card-editor.js
honstar: Ah, ok. Thanks for the explanation!

honstar (approved):

afmicka:
@yesil this is failing docs Nala because the promo is applied to the price in the description field of the fragment which was authored without it originally.
See https://mwpw-188624--mas--adobecom.aem.live/web-components/docs/ccd-mini.html
Screenshot 2026-04-16 at 13 17 44

this field previously is used for UPT links. Are we sure that we don't have cases where prices were authored without the promo in the fragments that have promocode field populated? These changes will affect all such fragments and will add the old price to all the prices in the text

afmicka:
@yesil In the jira it says that the promo should be applied to CTA. Even though when i open the OST i do see the promo code, it is missing from the CTA commerce link. There should be `apc=` param if a valid code is applied. Could you please check on this?

https://mwpw-188624--mas--adobecom.aem.live/studio.html#fragmentId=9713e893-6180-492f-a053-27302d18517c&page=fragment-editor&path=sandbox
Screenshot 2026-04-16 at 13 43 48

On prod, when you author the CTA with promo, it has it
Screenshot 2026-04-16 at 13 44 03

yesil:
@afmicka it was discussed and since checkout links now have 3in1 modals, the promo code is not needed on the CTA anymore apparently.
@npeltier  IIRC, we discussed this and we purposefully decided to limit this to prices and not checkout links, correct?

yesil:
@afmicka 

> this field previously is used for UPT links. Are we sure that we don't have cases where prices were authored without the promo in the fragments that have promocode field populated? These changes will affect all such fragments and will add the old price to all the prices in the text

you are right. With this PR, you can actually go and disable a promo on a price if the promo code is set on the card.
The question is how many fragments will be affected by this change.

I have an idea how to smoothly introduce this feature without breaking existing content, I'll add a tech discussion item.

npeltier:
@afmicka @yesil i think for 3in1 we don't want apc code in CTAs, but this could still be a 3in1 consumer change. I wonder if somewere apc code is still needed

afmicka:
> @afmicka it was discussed and since checkout links now have 3in1 modals, the promo code is not needed on the CTA anymore apparently. @npeltier IIRC, we discussed this and we purposefully decided to limit this to prices and not checkout links, correct?

are we sure that all CTAs are going to be 3in1? What about non-plans fragments (special offers, image, ..) or other consumers? CCD or adobe home?

yesil:
~@afmicka promocode still does not apply to CTAs. I think it can be a fast follow.~

afmicka:
@yesil when i open the card that is used in Nala tests, in the editor and the card preview the promo is applied to the price in the description (on prod there is no promo applied to this price). At the same time, the Save button is disabled so user cannot save this new option if they want to. On the card preview in the render view there is not strikethrough price. This is confusing for the user. They might think that they need to correct something once they open the editor. 

https://mwpw-188624--mas--adobecom.aem.live/studio.html#fragmentId=df357350-95b2-47f2-844f-df2e491eecef&page=fragment-editor&path=nala

Screenshot 2026-04-28 at 12 57 43

render view: 
Screenshot 2026-04-28 at 13 00 37


If the changes have to be applied once the editor is opened, then maybe better to at least have a Save option back?

afmicka:
@yesil saving the fragment does not resolve the issue. The save button is always enabled, no matter how many times i save it or make a real change and save. The next time i open the fragment, save button is again enabled

https://mwpw-188624--mas--adobecom.aem.live/studio.html#fragmentId=6f189be0-d64b-468f-b340-92888206cce8&page=fragment-editor&path=nala

afmicka:
@yesil could you please check one thing. When you apply promo on the price from the promo code field, there is no `data-promotion-code` attribute on the price element. See prod how the price looks when promo is applied

Screenshot 2026-04-29 at 15 17 54

price on this fragment on your branch does not have it but the promo is obviously applied
https://mwpw-188624--mas--adobecom.aem.live/studio.html#fragmentId=9713e893-6180-492f-a053-27302d18517c&page=fragment-editor&path=sandbox
Screenshot 2026-04-29 at 15 18 14

yesil:
> @yesil could you please check one thing. When you apply promo on the price from the promo code field, there is no data-promotion-code attribute on the price element. See prod how the price looks when promo is applied

as discussed with @afmicka on slack, this is intended behavior.

PR #740, opened 2026-04-03
MWPW-190616+MWPW-194710: UI follow-up for grouped variation translation

mirafedas:
Resolves https://jira.corp.adobe.com/browse/MWPW-190616
https://jira.corp.adobe.com/browse/MWPW-194710

AC 1:
1. Fixed dropdown chevron direction
2. Removed grey connector lines
3. Enabled checking the checkbox by row click
4. Grouped variations now show the offer ID. If the offer ID is not set on the variation, the parent offer ID is displayed.
These changes are covered with unit tests.
5. Additionally, fixed a few minor CSS issues caused by the bulk publish PR.
6. Increases z-index of the prevent leave modal to fix the MWPW-194710.

AC 2: variations render properly, so no changes made.

## 🧪 Nala E2E Tests

### To run Nala tests again:

- Remove the `run nala` label

## Test URLs:

- Before: https://main--mas--adobecom.aem.live/studio.html#page=translations&path=sandbox
- After: https://mwpw-190616--mas--adobecom.aem.live/studio.html#page=translations&path=sandbox


io/studio/src/locales.json
```diff
@@ -0,0 +1,277 @@
+{
```
yesil: this should be imported from io/www

yesil (approved):

io/studio/src/cleanup-variations/index.js
```diff
@@ -0,0 +1,243 @@
...
+    if (!authToken) {
+        return { statusCode: 401, body: { error: 'Missing Authorization header' } };
+    }
+
+    const odinEndpoint = params.odinEndpoint;
+    if (!odinEndpoint) {
+        return { statusCode: 500, body: { error: 'Missing odinEndpoint configuration' } };
+    }
+
+    const { surface: targetSurface, locale: targetLocale, dryRun = true } = params;
```
Axelcureno: **SSRF risk — `odinEndpoint` should come from config, not the request body**

Every other action in `io/studio` gets `odinEndpoint` from `$ODIN_ENDPOINT` injected via `app.config.yaml` inputs. Accepting it from the POST body lets any authenticated caller point this action at an arbitrary AEM host.

Fix: add to `app.config.yaml` under `cleanup-variations`:
```yaml
inputs:
    LOG_LEVEL: debug
    odinEndpoint: $ODIN_ENDPOINT
```
The value arrives in `params.odinEndpoint` the same way — no read-side change needed — but the caller can no longer override it.

io/studio/src/common.js
```diff
@@ -392,6 +396,8 @@ async function deleteFragmentById(odinEndpoint, fragmentId, authToken, etag) {
 
 module.exports = {
     DEFAULT_PACKAGE_NAME,
```
Axelcureno, line 431: Unnecessary export — `cleanup-variations/index.js` only imports `localeFromPath` from `common`, never `PATH_TOKENS`. Remove `PATH_TOKENS` from the `module.exports` block.

Axelcureno (changes requested):
Two blockers (SSRF + unnecessary export) and one nit — see inline comments already posted.

Axelcureno (approved):
Both blockers resolved (SSRF + PATH_TOKENS). The console.warn is minor — ship it.

npeltier (changes requested):
sorry Myra, but we shouldn't add io changes (esp. of that magnitude) together with other changes. Can you eventually track what looks like an improvement in a separate ticket, and leave that one with only UI changes?

studio/src/common/utils/items-loader.js
```diff
@@ -14,6 +14,34 @@ import {
...
+    if (repository.resolveHydratedParentFragment && Fragment.isGroupedVariationPath(card.path)) {
+        try {
+            const parentFromAem = await repository.resolveHydratedParentFragment(card.path);
+            fallbackWcsOsi = new Fragment(parentFromAem).getFieldValue('osi');
+        } catch (err) {
+            console.warn(`Failed to load parent fragment for grouped variation ${card.id}:`, err.message);
+        }
+    }
+
+    return loadOfferData(card, { cache, signal, timeoutMs: 10000, fallbackWcsOsi });
```
honstar: Setting `timeoutMs: 1000` is redundant since `loadOfferData` already sets this value as default.

studio/src/common/utils/render-utils.js
```diff
@@ -29,7 +29,7 @@ export function renderFragmentStatusCell(status) {
 export function getItemTypeLabel(item) {
     if (!item) return 'Unknown';
     if (Fragment.isGroupedVariationPath(item.path)) return 'Grouped variation';
-    if (item.model?.path?.includes('/dictionary/')) return 'Placeholder';
+    if (item.model?.path?.includes('/diction')) return 'Placeholder';
```
honstar: What is the purpose of this change? There is no mention of anything related to dictionary issues in the PR description or ticket.

studio/src/common/utils/items-loader.js
```diff
@@ -14,6 +14,34 @@ import {
...
 } from './item-loading.js';
 
+/**
+ * Resolves WCS offer data for a view-only table row: uses the fragment's own OSI, or parent card OSI when the row is a grouped variation without OSI.
+ * @param {Object} card - Fragment payload (may be a parent card or a /pzn/ variation)
+ * @param {Object} repository - MasRepository
+ * @param {AbortSignal} [signal] - Abort signal
+ * @returns {Promise<Object|null>}
+ */
+async function loadOfferDataForViewOnlyCard(card, repository, signal, cache = new Map()) {
```
honstar: This new wrapper function for `getOfferData` drops support for `fallbackWcsOsi` as function argument. This will lead to unconditional loading of an already-loaded parent card's OSI (see line 35 below).

The `loadGroupedVariations` function (https://github.com/adobecom/mas/pull/740/changes#diff-7a9a7fcac46ba2a26c823d61508838413248a69b60c8201552cba03191d3c4f2R201) passes in a parent OSI and should be respected to avoid extra round trips.

honstar (changes requested):

studio/src/common/utils/render-utils.js
mirafedas: It was not requested in this Jira, but I didn't want to create a separate PR for such a small change, thus included it here. The issue is that we have two cf models, one has 'dictionary' with one 'n', and another one has 'dictionnary' with double 'n'. Both are used. I created a Jira to unify this: https://jira.corp.adobe.com/browse/MWPW-194882 , but until it is done, we need to make it work with both models, that's why I truncated it to 'diction'.

honstar: I see, I was actually wondering about that myself but have an inkling where `dictionnary` might have come from ;)

In any case, wouldn't it be better to explicitly check for both `dictionary` and `dictionnary` for the time being?

mirafedas: I think it's better if we remove the `dictionnary` model at some point, because it is misleading, we should have only one model for placeholders. And then we will not need to make any change to this line of code because it's universal :)

honstar: I don't think this is accurate, just checked our CF models on odin prod: while it is true that there are 2 seemingly the same models for placeholders, they are in fact different:
- `dictionnary` ([model](https://author-p22655-e59433.adobeaemcloud.com/mnt/overlay/dam/cfm/models/editor/content/editor.html/conf/mas/settings/dam/cfm/models/dictionnary)): this is an actual placeholder with key/value pair
- `dictionary` ([model](https://author-p22655-e59433.adobeaemcloud.com/mnt/overlay/dam/cfm/models/editor/content/editor.html/conf/mas/settings/dam/cfm/models/dictionary)): this is the index that holds references to above placeholder entries

I don't think the latter - the index fragment - would play a role in the scope of this PR.

honstar (approved):

Andrei4226 (approved):
LGTM:)
Mention: Good point to catch the case with `/dictionnary` and `/dictionary/ `in another ticket
It can create confusion now

mirafedas:
@Andrei4226 thanks for the review! I didn't want to create a separate Jira for a one-word change, and this is related to the functionality I'm fixing in this PR, but in case it was not related or a bit bigger change, I would definitely create a separate Jira 👍

Andrei4226:
> @Andrei4226 thanks for the review! I didn't want to create a separate Jira for a one-word change, and this is related to the functionality I'm fixing in this PR, but in case it was not related or a bit bigger change, I would definitely create a separate Jira 👍

I see now that the ticket for unify is Done:) 
I got your point 👍

npeltier (approved):

afmicka:
@mirafedas is by any chance this "Select all" for grouped variations defined somewhere? I can't find it in figma
Screenshot 2026-05-22 at 16 03 57


figma example i found:
Screenshot 2026-05-22 at 16 03 47

afmicka:
@mirafedas the "click on the row" to select item works only for Fragments table view. Does not work for Collections and Placeholders views. In those 2 only clicking on checkbox will select the items

mirafedas:
@afmicka Thank you for testing. 
The 'Select all' was added in the scope of this Jira: https://jira.corp.adobe.com/browse/MWPW-188626 , I can see it in Figma here: https://www.figma.com/design/GDDRLo3S7fz0SMRefpJOpx/M-Studio?node-id=19732-166728&m=dev . It looks to me that it should have been added for bulk publish only, but as we reuse the same table component there and in the translation editor page, it got populated. If it should be removed, we should do it in a separate Jira.
The collection and placeholder tables are fixed now 👍

afmicka (approved):

PR #741, opened 2026-04-03
MWPW-191059: route Studio preview through MAS gateway

Axelcureno:
Resolves https://jira.corp.adobe.com/browse/MWPW-191059

Routes Studio preview fragments through a new MAS gateway (`preview.mas.corp.adobe.com`) instead of calling Freyja/Odin directly. The gateway injects auth server-side, so the browser no longer holds an IMS token.

## Why

- Freyja stage returns HTTP 500 on CORS preflight when `Authorization` is sent, blocking authenticated calls from `localhost`.
- IMS tokens in the browser are leakable; gateway owns the credential.

## What changed

- `fragment-client.js` — always target the gateway, no `Authorization`, no env/backend toggles.
- `mas-repository.js` — dictionary loads route through `getPreviewContext()` instead of hardcoded Odin URL.
- `mas-fragment-editor.js` — fil_PH/en_US locale discovery uses the gateway's `?path=` query.
- `io/www/paths.js` — adds `GATEWAY_PREVIEW_URL` constant (2-line addition; rest of `io/www/src/` unchanged).
- Removed: `masFreyjaToken` sessionStorage plumbing, `ODIN_PREVIEW_ORIGIN`, `preview.backend=odin` toggle, `freyja-bench.html`.

## Cross-team

- Gateway owned by Ilyas; currently on his personal token, will swap to tech account.
- ODIN-950 closed — MAS Studio tech account verified against Freyja PROD (HTTP 200, full hydration).
- Load test: 100% success, 194 req/s, p95 167ms.

## Test URLs

- After: https://mwpw-191059--mas--adobecom.aem.live/studio.html

## Checklist

- [x] Tests pass locally (io/www 99% coverage gate, 54 studio tests)
- [x] Browser smoke test: all preview calls hit gateway, no CORS errors

## Follow-up

Ilyas swaps gateway from personal token to ODIN-950 tech account before prod.


npeltier (changes requested):
results look promising but:

- i would make authToken part of the preview configuration, as it's only for preview,
- io common fetch should *not* be odin specific (it's used with other hosts), nor preview specific, i would remove the fallback all together, and just make switch from one to the other easy enough,
- please add tests URLs so we can test your change (not sure how easy it is)
- don't forget this should be working on aem.page URLs in milo as well! right now it's only considering studio,
- please move benchmarks to a dedicated place, leaving fragment-client.js to what it does

honstar:
Please address formatting issues and improve code coverage.

docs/superpowers/plans/2026-03-31-freyja-preview-migration.md
```diff
@@ -0,0 +1,703 @@
+# Freyja v2 Preview Migration — Implementation Plan
```
honstar: Just curious: what AI agent/workflow did you use for these markdown files?

io/www/src/fragment/utils/paths.js
```diff
@@ -2,6 +2,10 @@ const MAS_ROOT = '/content/dam/mas';
 
 const FRAGMENT_URL_PREFIX = 'https://odin.adobe.com/adobe/sites/fragments';
 
+const FREYJA_PREVIEW_URL = 'https://preview-p22655-e59433.adobeaemcloud.com/adobe/contentFragments';
```
honstar: This is hardcoded to prod, what about supporting non-prod environments via `aem.env`?

studio/bench/freyja-bench.html
```diff
@@ -0,0 +1,157 @@
+<!doctype html>
```
honstar, line 1: When trying to run this via https://mwpw-191059--mas--adobecom.aem.live/studio/bench/freyja-bench.html, all requests from the feature branch to freyja preview fail with CORS errors.

io/www/src/fragment/utils/paths.js
npeltier: it's not really prod, but "preview" freyja
prod is odin.adobe.com

npeltier (changes requested):
what i meant by aem.page test link @Axelcureno is to have 
https://main--cc--adobecom.aem.page/sa_ar/products/catalog?maslibs=mwpw-191059--mas--adobecom
working (which isn't atm)

io/www/src/fragment/utils/paths.js
honstar: Well, with prod I meant the prod environment in AEMaaCS speak. e59433 is prod, regardless if it's odin or freyja.

npeltier:
also @Axelcureno, as discussed, performance is one part of the fix, we need to know how much time it takes from when content is changed on author to when content is previewed on that freyja server. 
Once we do:
- either it's acceptabe time (let's say < 10s),
- either we need to mitigate with authors,
- either we need to auto-publish to preview freyja when we save

docs/superpowers/plans/2026-03-31-freyja-preview-migration.md
Axelcureno: https://github.com/obra/superpowers
Sorry, I will remove this and add a rule to not commit these files.

Axelcureno:
We are blocked by product team who needs to make more changes to enable Freyja for us, so the PR is not yet ready for review. I will in the meantime address the comments. Once we are unblocked I will request reviews. @honstar @npeltier

honstar (approved):

web-components/src/ims.js
```diff
@@ -8,6 +8,11 @@ export function imsReady({ interval = 200, maxAttempts = 25 } = {}) {
         /* c8 ignore next 10 */
         function poll() {
             if (window.adobeIMS?.initialized) {
+                const token = window.adobeIMS?.getAccessToken?.()?.token;
```
yesil, line 11: ims.js is meant to be used for loggedin users and is not freyja token.

preview on acom stage should work for both logged-out and logged-in users, so we cannot use an access token.

web-components/src/aem-fragment.js
```diff
@@ -479,11 +479,16 @@ export class AemFragment extends HTMLElement {
     async generatePreview() {
         const fragmentClientUrl = this.getFragmentClientUrl();
         const { previewFragment } = await import(fragmentClientUrl);
-        const data = await previewFragment(this.#fragmentId, {
+        const freyjaToken = sessionStorage.getItem('masFreyjaToken');
```
yesil, line 482: Please see ODIN-949, `aem-fragment` should work with the second url in the ticket and we will likely remove transformAuthorData from `aem-fragment` and just use transformPublishData for preview as well, happy to discuss in more detail.

io/www/src/fragment/utils/paths.js
yesil, line 1: in this file,  we need to switch the value of FRAGMENT_URL_PREFIX from https://odin.adobe.com/adobe/sites/fragments to https://odinpreview.corp.adobe.com/adobe/sites/fragments

In the case of preview.

yesil (changes requested):
@Axelcureno please see ODIN-949, let's connect if you need more details about my change request.

yesil:
@Axelcureno I opened #777 as an alternative, I propose to close this one.

PR #742, opened 2026-04-03
MWPW-191570: Studio pagination, search refill, and translation dialog UX

Axelcureno:
## Overview

This PR improves Studio pagination, search behavior, and the translation items dialog UX:

1. **Skeleton shimmer loading** for translations and placeholders admin tables
2. **Pagination plumbing fixes** — sentinel cascade guards, RAF re-observe, eager-load cap, re-entrancy guard for processCardsData
3. **Search refill loop** — auto-fetches additional cursor pages when a narrow filter leaves fewer than 25 visible items, eliminating the "one result, then loading, then restart" UX
4. **Translation dialog UX** — close (X) icon + backdrop dismiss, global search bar above tabs with 300ms debounce, reduced whitespace

## Resolves
https://jira.corp.adobe.com/browse/MWPW-191570

### Search Refill Loop (mas-repository.js)
- `MIN_FILTERED_PAGE_RESULTS = 25` constant with JSDoc
- `#refillBelowThreshold()` — mirrors `#eagerLoadAllPznPages` pattern: fire-and-forget from the non-personalization branch, identity-based abort, AbortError catch
- Wired into `searchFragments` else branch — only runs when personalization is OFF and cursor has more pages
- Renames `MAX_EAGER_PAGES` to `MAX_EAGER_PZN_PAGES` with JSDoc

### Translation Dialog UX (mas-items-selector.js, mas-search-and-filters.js, mas-translation-editor.js)
- `dismissable` attribute on `sp-dialog-wrapper` — adds close icon + backdrop-click-to-close
- Search bar moved from per-tab to global dialog header next to "Select items" title
- 300ms debounce on search input (was per-keystroke)
- Filter reactivity consolidated through `willUpdate` — removed redundant direct `#applyFilters()` calls

### Translation Re-entrancy (translation-items-loader.js)
- `processCardsData` stashes pending payloads instead of silently dropping concurrent invocations
- DRY table-head template extracted in `mas-translation.js`

### Pagination Foundation (earlier commits)
- Sentinel cascade guard on non-empty items list
- RAF-deferred sentinel re-observe to prevent cascade loads
- `MAX_EAGER_PZN_PAGES = 20` cap on eager personalization prefetch
- Skeleton shimmer for translations and placeholders admin tables

## Test URLs

- **Before:** https://main--mas--adobecom.aem.live/
- **After:** https://mwpw-191570--mas--adobecom.aem.live/

## Test Coverage

- 5 new refill-below-threshold test cases (happy path, early termination, cursor exhaustion, abort, pzn skip)
- 3 existing pagination tests updated for new refill behavior
- 25 search-and-filters tests updated for external searchQuery property pattern
- Translation skeleton test aligned with shimmer UI

## QA Checklist

- [x] C1. Unit tests pass locally (1502 passing, 5 pre-existing failures unrelated)
- [ ] C2. NALA test (check with #fishbags if needed)
- [ ] C3. All checks green
- [x] C4. Test URLs in PR description
- [ ] C5. Demo ready
- [x] C6. JIRA ACs validated

## Screenshots

_(Add before/after screenshots of narrow-filter refill and dialog UX changes)_


studio/src/mas-repository.js
```diff
@@ -459,6 +459,9 @@ export class MasRepository extends LitElement {
                     this.#eagerLoadAllPznPages(cursorState, searchController);
                 } else {
                     this.#abortControllers.search = null;
+                    if (cursorState) {
+                        this.#refillBelowThreshold(cursorState, searchController);
```
yesil, line 468: **Bug: loading flag race** — `#refillBelowThreshold` is fire-and-forget here. It sets `loading=true` synchronously (before its first `await`), but then `searchFragments` falls through to `Store.fragments.list.loading.set(false)` a few lines below. That overwrites the refill's loading guard.

At that point `hasMore=true` + `loading=false` — so the sentinel can fire `loadNextPage()` while `#refillBelowThreshold` is still iterating the same cursor.

The PZN path avoids this by setting `hasMore=false` before the fire-and-forget `#eagerLoadAllPznPages` call. The refill path doesn't have that guard.

Suggested fix — skip the outer `loading.set(false)` when refill is about to run and let refill's `finally` block own the loading lifecycle:
```js
let refilling = false;
// …
if (cursorState) {
    refilling = true;
    this.#refillBelowThreshold(cursorState, searchController);
}
// after try block:
if (!refilling) Store.fragments.list.loading.set(false);
```

(The existing test passes because it exercises `#refillBelowThreshold` in isolation, without the outer `searchFragments` clobbering the flag.)

studio/src/translation/mas-items-selector.js
```diff
@@ -89,6 +124,7 @@ class MasItemsSelector extends LitElement {
                                 : html`
                                       <mas-search-and-filters
                                           .type=${tab.value}
+                                          .searchQuery=${this.searchQuery}
```
yesil: Something to watch: `searchQuery` is now pushed into all three tab panels, including the hidden ones. Each `mas-search-and-filters` instance re-filters and rewrites its display store on every change via `willUpdate`, so every search update does work for cards, collections, **and** placeholders even when only one tab is visible. Consider gating the filter pass on whether the tab is actually selected.

studio/src/translation/mas-items-selector.js
```diff
@@ -54,6 +58,24 @@ class MasItemsSelector extends LitElement {
         Store.translationProjects.showSelected.set(!this.showSelected);
     }
 
+    #handleSearchInput({ target: { value } }) {
+        clearTimeout(this.#searchDebounceTimer);
```
yesil: There's already a `debounce` utility in `studio/src/utils.js:105`. This hand-rolled timer duplicates it and adds cleanup burden (`#searchDebounceTimer` field + `clearTimeout` in `disconnectedCallback`).

Using the existing utility:
```js
import { debounce } from '../utils.js';

#handleSearchInput = debounce(({ target: { value } }) => {
    this.searchQuery = value;
}, 300);
```
That eliminates the private field, the manual `clearTimeout`, and the `disconnectedCallback` override.

studio/src/translation/translation-items-loader.js
```diff
@@ -304,6 +307,12 @@ async function processCardsData(allCards, repository, state) {
         Store.translationProjects.cardsByPaths.set(cardsByPaths);
     } finally {
         state.isProcessingCards = false;
+        performance.measure('processCardsData', 'processCardsData:start');
+        if (state.pendingCards && !signal?.aborted) {
+            const next = state.pendingCards;
+            state.pendingCards = null;
+            processCardsData(next, repository, state);
+        }
```
yesil, line 305: Nit: this recursive call is not `await`ed. If it rejects, the error becomes an unhandled promise rejection. Either `await` it (fine inside `finally`) or add a `.catch`:
```js
processCardsData(next, repository, state).catch(() => {});
```

studio/src/mas-repository.js
Axelcureno: Good catch — fixed in fd1b1c5ef. Added `let refilling = false` before the `try` block, set it when `#refillBelowThreshold` is fired, and guarded the trailing `Store.fragments.list.loading.set(false)` with `if (!refilling)`.

Follow-up in 576165eff: a challenge-agent pass flagged a remaining window — if a second `searchFragments` aborts the first *after* `refilling=true` but *before* the scheduled refill took ownership of the loading lifecycle, the `catch` path left `loading` pinned. The catch now also clears loading on `AbortError` when `refilling` was set.

studio/src/translation/mas-items-selector.js
Axelcureno: Fixed. First pass in fd1b1c5ef gated the prop with `tab.value === this.selectedTab ? this.searchQuery : ''`. But that still triggered `willUpdate → #applyFilters()` on hidden tabs (just with empty string), which clobbered their per-tab filter state.

Final fix in 576165eff: only the active tab's `mas-search-and-filters` is rendered. `sp-tab-panel` hides inactive panels anyway, so we made the hiding structural. Tests updated (+5 passing, 148 total).

studio/src/translation/mas-items-selector.js
Axelcureno: Fixed in fd1b1c5ef. Replaced with `debounce` from `../utils.js`. Removed the `#searchDebounceTimer` field, the manual `clearTimeout` in `#handleSearchSubmit`, and the `disconnectedCallback` override (which only existed to clear the timer).

studio/src/translation/translation-items-loader.js
Axelcureno: Fixed in fd1b1c5ef — added `await` to the recursive `processCardsData()` call.

Noted for follow-up: a challenge-agent pass pointed out that awaited recursion inside `finally` can grow stack/closure depth under sustained load. Leaving the `await` for now (correctness first) and filed a cleanup to convert to a `while (state.pendingCards)` loop in a post-merge ticket.

yesil (approved):

mirafedas (approved):

antonio-rmrz (approved):

Roycethan (changes requested):
@Axelcureno Feedbacks:
https://mwpw-191570--mas--adobecom.aem.live/studio.html#page=translation-editor&path=sandbox

1) The 'Cancel' and 'Add Selected Items" are not visible:

Expected:

Image

Actual:
Image

2) On the above screen "Selected items" button needs to be aligned properly inside the container right line and below the searched items

3) In Translations: When i searched fragments with more than 50 by moving down the line say 158 results and close the button and re-open the modal its rendering all the 158 results again and not resetting back to 50 results.
Similarly with filters applied are not reset after closing and re-opening the modal

4) In translations: Collections tab filters out 7 results initially and Placeholders filters all the 77 results - are we not doing the perf enhancement on these tabs ?

Axelcureno:
Observations are now fixed @Roycethan

Roycethan:
> @Axelcureno Feedbacks: https://mwpw-191570--mas--adobecom.aem.live/studio.html#page=translation-editor&path=sandbox
> 
> 1. The 'Cancel' and 'Add Selected Items" are not visible:
> 
> Expected:
> 
> Image
> Actual: Image
> 
> 2. On the above screen "Selected items" button needs to be aligned properly inside the container right line and below the searched items
> 3. In Translations: When i searched fragments with more than 50 by moving down the line say 158 results and close the button and re-open the modal its rendering all the 158 results again and not resetting back to 50 results.
>    Similarly with filters applied are not reset after closing and re-opening the modal
> 4. In translations: Collections tab filters out 7 results initially and Placeholders filters all the 77 results - are we not doing the perf enhancement on these tabs ?

#1 and #2 are fixed now.
#3 - decided to keep as is , since its cached
#4 - Axel should increase the load count for collections tab, per him placeholders is bit risky.
#5 - Close button  is not rendered now post recent commit

Roycethan:
> https://jira.corp.adobe.com/browse/MWPW-191570

#4 and #5 are fixed , introduced this issue: 
#6 : When filter applied on project modal its breaking the view and no results returned:
image

Roycethan:
> > https://jira.corp.adobe.com/browse/MWPW-191570
> 
> #4 and #5 are fixed , introduced this issue: #6 : When filter applied on project modal its breaking the view and no results returned: image

#6 filter applied works now
#7 Auto type search doesn't work

Roycethan:
> > > https://jira.corp.adobe.com/browse/MWPW-191570
> > 
> > 
> > #4 and #5 are fixed , introduced this issue: #6 : When filter applied on project modal its breaking the view and no results returned: image
> 
> #6 filter applied works now #7 Auto type search doesn't work

#7 - fixed

Roycethan (commented):
@Axelcureno  Collections tab is returning 0 results in workspaces like Acom, Nala etc...
ex: https://mwpw-191570--mas--adobecom.aem.live/studio.html?#page=translation-editor&path=acom

Image

nala/docs/ccd-mini/mini.spec.js
```diff
@@ -177,7 +177,7 @@ export const features = [
             planTypeText: 'Annuel, facturé mensuellement',
             recurrenceText: '/mois',
             seeTerms: {
-                text: 'Conditions d’utilisation',
+                text: 'Voir les conditions',
```
afmicka, line 180: please remove this change unless you have a reason why you are changing the placeholder value in the Nala surface. Lets discuss if the change is really needed. For now i am setting the previous value

Image

Roycethan:
when more number of  fragments selected , i see the 'Hide selection' button is bit overlapping over the right pane items
image

Roycethan (approved):

PR #748, opened 2026-04-09
MWPW-189894: Lingo issue with placeholder fix

seanchoi0:
Resolves https://jira.corp.adobe.com/browse/MWPW-189894

## 🧪 Nala E2E Tests

### To run Nala tests again:

- Remove the `run nala` label

## Test URLs:

- After: https://mwpw-189894--mas--adobecom.aem.live/

## More Test URLs:
- https://main--cc--adobecom.aem.live/fr/drafts/gunn/mas-lingo-ksink?akamaiLocale=CH&maslibs=MWPW-189894&mas-io-url=https://14257-merchatscale-seanchoi.adobeioruntime.net/api/v1/web/MerchAtScale
image


io/www/src/fragment/transformers/customize.js
```diff
@@ -304,6 +304,8 @@ async function customize(context) {
         ...context,
         status: 200,
         body: customizedFragment,
+        locale: regionLocale,
+        defaultLocale,
```
npeltier: see below

npeltier (changes requested):

io/www/src/fragment/transformers/replace.js
```diff
@@ -153,14 +153,7 @@ function replaceValues(input, dictionary, calls) {
...
 
-async function init(context) {
-    // we fetch dictionary at this stage only if id has already been cached
-    // because we can't know surface of fragment *before* first fetch
-    // if dictionaryId is present in cache - early load dictionary
-    // if nothing in cache - dictionaryId and dictionary itself will be loaded later,
-    // during process
-    return await getDictionary(context);
-}
+async function init() {}
```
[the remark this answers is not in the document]
seanchoi-dev: I moved default-locale + computeRegionLocale into fetchFragment’s init. replace.init only awaits that promise, merges it onto context, then calls getDictionary so we still start the dictionary in init as soon as the locale is correct that is right after fetchFragment finishes.

npeltier (changes requested):
thanks a lot @seanchoi-dev that looks much better, one additional thing, though:
if you look at getRequestInfos, it is the async promise a lot of initial requests are pending on to start, among which the dictionary id, and settings id requests we want to start ASAP.
"problem" with this is that you add default variation fetch to that processing which may increase overall time of the pipeline when nothing is cached.

i think as soon as you get initial fragment fetch, so called requestsInfos should be available to other init method. So probably you should encapsulate the default fragment fetch in that first promise

seanchoi-dev:
> thanks a lot @seanchoi-dev that looks much better, one additional thing, though: if you look at getRequestInfos, it is the async promise a lot of initial requests are pending on to start, among which the dictionary id, and settings id requests we want to start ASAP. "problem" with this is that you add default variation fetch to that processing which may increase overall time of the pipeline when nothing is cached.
> 
> i think as soon as you get initial fragment fetch, so called requestsInfos should be available to other init method. So probably you should encapsulate the default fragment fetch in that first promise

I added an early `requestInfos` promise that resolves right after the first fragment fetch and path parse, so anything that only needs surface/path/body (e.g. via `getRequestInfos`) no longer waits on the default-locale fetch.

I hope this addresses the concern that dictionary/settings inits were gated on the full fetchFragment timeline when nothing was cached.

io/www/src/fragment/transformers/customize.js
```diff
@@ -13,53 +12,16 @@ function skimFragmentFromReferences(fragment) {
...
-        ({ body } = response);
-    }
-    context.defaultLocale = defaultLocale;
-    return { body, defaultLocale, status: 200 };
-}
-
 async function init(context) {
+    if (context?.promises?.fetchFragment) {
+        return await context.promises.fetchFragment;
+    }
```
npeltier: why is that i don't understand?
Init are here to trigger asap if possible in parallel network dependencies. From what i see, you don't need init in customize anymore, as all has been triggered from fetch fragment

web-components/src/aem-fragment.js
```diff
@@ -358,7 +358,7 @@ export class AemFragment extends HTMLElement {
             this.#rawData = fragment;
             return true;
         }
-        const { masIOUrl, wcsApiKey, country, locale } = this.#service.settings;
+        const { masIOUrl, wcsApiKey, locale, country } = this.#service.settings;
```
npeltier: nit: if you want to change order here, please do it alphabetically

web-components/src/settings.js
```diff
@@ -126,7 +170,7 @@ function getSettings(config = {}, service) {
         `https://www${env === Env.STAGE ? '.stage' : ''}.adobe.com/mas/io`;
     const preselectPlan = getParameter('preselect-plan') ?? undefined;
     return {
-        ...getLocaleSettings(config),
+        ...getLocaleSettings(mergeCommerceCountryFromGeo(config)),
```
npeltier: is that change needed to fix placeholder issue? to me it's interesting but not necessarily related to our problem

io/www/src/fragment/transformers/replace.js
```diff
@@ -154,12 +154,9 @@ function replaceValues(input, dictionary, calls) {
 }
 
 async function init(context) {
-    // we fetch dictionary at this stage only if id has already been cached
-    // because we can't know surface of fragment *before* first fetch
-    // if dictionaryId is present in cache - early load dictionary
-    // if nothing in cache - dictionaryId and dictionary itself will be loaded later,
-    // during process
-    return await getDictionary(context);
+    const fetchResult = await context?.promises?.fetchFragment;
```
npeltier: in the end you still wait for both, no?
i know that it's not easy...my suggestion is:
context.promises.fetchFragment is as before, with encapsulated another promise for customize

npeltier (changes requested):
i'm still a bit confused with how fetchFragment, replace, settings & customize transformers are playing together.

i don't get why we have (WCS) settings managing akamai country in the scope of that PR

i would suggest to create side JIRAs to skim down this PR to its minimum requested changes.

web-components/src/settings.js
seanchoi-dev: Yes, it is needed. This is needed for applying the Akamai/geo merge for the whole MAS settings object.

`mergeCommerceCountryFromGeo` fills in (or overrides) country from `akamaiLocale` or the akamai session blob when it’s safe to treat country as path/geo-derived.

If we kept `getLocaleSettings(config)` only, the merged country would never feed into those derived settings, so client requests could still disagree with Akamai/geo and we wouldn’t fully fix the “Lingo placeholders vs regional locale / country” alignment. So, yeah I think we need this unless you diagree.

io/www/src/fragment/transformers/customize.js
seanchoi-dev: I thought this is better for maintainability and clearer boundary between transformers but I guess I made you confused.

I’ll update them to remove init() from customize.js.

io/www/src/fragment/transformers/replace.js
seanchoi-dev: You’re right that `replace.init` still waits for the full fragment result; it needs merged locale / surface for the dictionary cache key, so it can’t stop at requestInfos alone. The gain is elsewhere: `getRequestInfos` uses requestInfos first, so dictionary id / settings can start after the first fetch without waiting on default-locale work.

I updated some code according to your suggestion, `promises.customize` is now the same promise as `promises.fetchFragment` (set inside `fetchFragment.init`), so customize is explicitly tied to that encapsulated result rather than a separate parallel init.

Here is an AI report on the performance changes in this branch:
```
Yes, but it’s narrow — not a blanket “faster pipeline.”

Where it can be faster
getRequestInfos → requestInfos: Anything that only needs surface / path / body from the first fragment response can proceed without waiting on the default-locale fetch and the rest of fetchFragment.
So inits that call getRequestInfos (e.g. settings, dictionary id resolution that only needs early metadata) can overlap more work with that second phase instead of sitting behind it.

Where it’s not faster (or similar)
replace.init still awaits the full fetchFragment / customize result before getDictionary, because the dictionary cache key needs the merged regional locale. So replace is not “freed” by requestInfos; that part is correctness, not latency win.
End-to-end for a cold request: you still do the same fragment + default-locale work as before; you mainly reorder what can run in parallel after the first fetch.

Summary
Meaningful win: less serial waiting on the full fetchFragment promise for code paths that only need early request metadata.
Not a big global speedup: total network work is largely the same; replace still waits for the full fragment outcome.

File: web-components/src/settings.js [hunk 7]
```diff
@@ -31,6 +31,50 @@ function getLocaleSettings({
...
+    try {
+        const raw = window.sessionStorage?.getItem('akamai');
+        if (!raw) return null;
+        const data = JSON.parse(raw);
+        const c =
+            data?.country ??
+            data?.country_iso ??
+            data?.countryCode ??
+            data?.country_code;
+        return c ? String(c).toUpperCase() : null;
```
Axelcureno: The 4-field fallback chain (`country`, `country_iso`, `countryCode`, `country_code`) suggests uncertainty about the Akamai session blob schema. If these are all documented variants, a brief comment explaining why would help. If only one or two are canonical, trimming to just those reduces surface area for bugs.

Also, the empty `catch {}` here and in `mergeCommerceCountryFromGeo` will silently swallow malformed JSON or other runtime errors — makes debugging in production very difficult. A `console.debug` on the catch path would be enough.

io/www/src/fragment/utils/cache.js
```diff
@@ -1,6 +1,14 @@
 import { getJsonFromState, mark, measureTiming } from './common.js';
 import { log } from './log.js';
-const getRequestMetadataKey = (context) => `req-${context.id}-${context.locale}`;
+function getRequestMetadataKey(context) {
+    const id = context.id;
+    const locale = (context.requestLocale ?? context.locale) || 'no-locale';
+    const raw = context.country;
+    if (raw != null && String(raw).trim() !== '') {
+        return `req-${id}-${locale}-${String(raw).trim().toUpperCase()}`;
```
Axelcureno: The country-aware cache key is the right fix (same locale + different country must not share cached metadata). Worth noting: this multiplies cache entries per fragment by the number of country variants hitting it. Probably fine in practice, but worth monitoring I/O Runtime state usage after deploy if high-traffic surfaces see many country variants for the same locale.

Axelcureno (commented):
Solid restructuring of the locale pipeline — the two-phase init in `fetchFragment` addresses the performance concerns well, and removing `customize.init()` simplifies the transformer coordination. A couple of items to address before this is ready to merge.

io/www/test/fragment/customize.test.js
npeltier: these should be moved to fetchFragment tests

npeltier: this will break

io/www/src/fragment/pipeline.js
npeltier, line 50: please rather add here `requestLocale` parameter

io/www/src/fragment/pipeline.js
```diff
@@ -143,6 +143,7 @@ async function main(params) {
 }
 
 async function mainProcess(context) {
+    context.requestLocale = context.locale;
```
npeltier: see above

io/www/src/fragment/utils/cache.js
```diff
@@ -1,6 +1,14 @@
...
+    }
+    return `req-${id}-${locale}`;
```
npeltier: while agreeing it's a needed fix, i would make cache key format here, that is 
always `req-${id}-${reqLocale}-${reqCountry}` 
reqCountry being `context.country || 'n/a'` or any other string that is not a country code :)

io/www/src/fragment/utils/common.js
```diff
@@ -186,14 +186,16 @@ async function getFragmentId(context, odinUrl, mark) {
...
+ * fetch + path parse), falling back to the full `fetchFragment` promise for backward compatibility.
  * @param {*} context
  * @returns parsedLocale, surface, fragmentPath
  */
 async function getRequestInfos(context) {
     let { body, parsedLocale, surface, fragmentPath } = context;
     if (!parsedLocale || !surface || !fragmentPath || !body) {
-        const fetchResult = await context.promises?.fetchFragment;
+        const earlyPromise = context.promises?.requestInfos ?? context.promises?.fetchFragment;
+        const fetchResult = await earlyPromise;
```
npeltier: not sure which backward compatibility you are referring to? there won't be mixed deployement, so you can go full new solution and just go with 
`const fetchResult = await context.promises?.requestInfos`

web-components/src/settings.js
npeltier: i understand it's related to same _kind_ of issues, but it has absolutetly no relationship with mas/io and mas/io placeholders. Worse, it's named settings like mas/io settings which adds to the confusion. I would really move it off of that PR for sake of clarity. 
for that future other PR, i would still leverage "country" parameter of the current mas-commerce-service that should already be set from akamai detection.

io/www/src/fragment/transformers/replace.js
```diff
@@ -154,12 +154,11 @@ function replaceValues(input, dictionary, calls) {
...
 async function init(context) {
-    // we fetch dictionary at this stage only if id has already been cached
-    // because we can't know surface of fragment *before* first fetch
-    // if dictionaryId is present in cache - early load dictionary
-    // if nothing in cache - dictionaryId and dictionary itself will be loaded later,
-    // during process
-    return await getDictionary(context);
+    // Dictionary cache key needs merged `locale` (region) from fetchFragment; `customize` is the same promise.
+    // Parallelism for dictionary id is via `getRequestInfos` → `requestInfos` inside getDictionaryId, not here.
+    const fetchResult = await (context?.promises?.customize ?? context?.promises?.fetchFragment);
```
npeltier: cool, please just use fetchFragment's and kill customize

io/www/src/fragment/transformers/customize.js
```diff
@@ -13,53 +12,18 @@ function skimFragmentFromReferences(fragment) {
...
-    if (!locale) {
-        context.defaultLocale = parsedLocale;
-        return { body, parsedLocale, status: 200 };
-    }
-    const defaultLocale = getDefaultLocaleCode(surface, locale);
-    if (!defaultLocale) {
-        return { status: 400, message: `Default locale not found for requested locale '${locale}'` };
+/** Fragment + locale variation: in pipeline, same promise as `fetchFragment` (`promises.customize`); tests without pipeline hit the fallback. */
+async function resolveFragmentInit(context) {
+    const fragmentInit = context?.promises?.customize ?? context?.promises?.fetchFragment;
```
npeltier: please just use fetchFragment's

io/www/src/fragment/transformers/customize.js
```diff
@@ -13,53 +12,18 @@ function skimFragmentFromReferences(fragment) {
...
-        ({ body } = response);
-    }
-    context.defaultLocale = defaultLocale;
-    return { body, defaultLocale, status: 200 };
-}
-
-async function init(context) {
     const { body, surface, fragmentPath, parsedLocale } = await getRequestInfos(context);
-    context = { ...context, surface, fragmentPath, parsedLocale, body };
+    const merged = { ...context, surface, fragmentPath, parsedLocale, body };
```
npeltier: i would not use "merge" here as it has special meaning in that transformer's context, and tbh at this stage you should already all all of those informations in the context (or if not please make fetchFragment process add it)

io/www/src/fragment/transformers/customize.js
```diff
@@ -285,48 +249,17 @@ function customizeTree(root, referencesTree = [], customizeContext) {
...
+    const { body, defaultLocale, status, message } = await resolveFragmentInit(context);
     const promos = await context.promises?.promotions;
 
     if (status != 200) {
         return { ...context, status, message };
     }
     const baseFragment = skimFragmentFromReferences(body);
     const { references, referencesTree } = body;
-    const regionLocale = computeRegionLocale({ ...context, defaultLocale, surface });
+    const regionLocale = context.regionLocale ?? computeRegionLocale({ ...context, defaultLocale, surface });
```
npeltier: why would context.regionLocale be unset? we should always have it here, and no mention of computeRegionLocale at all

npeltier (changes requested):
sorry a few more change requested, in addition to 2 "major" ones:
- please move out the mas settings changes,
- i realise looking at your change code would be clearer by splitting fetchFragment into 2: first being requestInfo or fetchFragment, almost untouched, and then inserting before promotions the defaultLanguage one, that is providing that second part. This way we would have init & process concept untouched. Wdyt?

web-components/src/settings.js
seanchoi-dev: This code is removed.

io/www/src/fragment/utils/cache.js
seanchoi-dev: I reverted these code as well.

io/www/src/fragment/transformers/customize.js
seanchoi-dev: I removed `computeRegionLocale` from customize() and only use `context.regionLocale` when the pipeline has already run `fetchFragment` process (so region is always on context there).

The `?? regionLocaleFromInit` is for our unit tests: they call customize.process() with a stub context and no promises.fetchFragment (see `customize.test.js` — the small process() helper only sets `context.promises = {}` and runs `customize.process`). In that path we never run `fetchFragment` process, so `context.regionLocale` is never set. `resolveFragmentInit` then uses the fallback (`getRequestInfos` + `getDefaultLanguageVariation`) and now returns `regionLocale` the same way `fetchFragment` init would, and customize reads it from that result.

Example: a test builds { locale: 'fr_FR', country: 'CA', body, surface, … } and expects the Canadian variation path — regionLocale has to come from the init-shaped result, not from context, because there is no prior fetchFragment step in the test harness.

io/www/src/fragment/transformers/customize.js
```diff
@@ -14,52 +13,25 @@ function skimFragmentFromReferences(fragment) {
...
-async function init(context) {
-    const { body, surface, fragmentPath, parsedLocale } = await getRequestInfos(context);
-    context = { ...context, surface, fragmentPath, parsedLocale, body };
+    const { body, surface, fragmentPath, parsedLocale } = requestInfosCached ?? (await getRequestInfos(context));
+    const variationContext = { ...context, surface, fragmentPath, parsedLocale, body };
     if (!surface || !fragmentPath) {
         return { status: 400, message: 'Missing surface or fragmentPath' };
     }
-    return await getDefaultLanguageVariation(context);
+    const result = await getDefaultLanguageVariation(variationContext);
```
npeltier: you should not call this again, but look at the `promises.defaultLanguage ` that is fed because you called it in the init of the transformer

io/www/src/fragment/transformers/customize.js
```diff
@@ -14,52 +13,25 @@ function skimFragmentFromReferences(fragment) {
...
+    if (result.status !== 200) {
+        return result;
+    }
+    const defaultLocale = variationContext.defaultLocale;
+    const regionLocale = computeRegionLocale({ ...variationContext, defaultLocale });
```
npeltier: this could be an output of defaultLanguage as well. Please move that code there, and just here consume it

io/www/src/fragment/transformers/defaultLanguage.js
```diff
@@ -0,0 +1,137 @@
...
+        surface: response.surface,
+        fragmentPath: response.fragmentPath,
+        defaultLocale: response.defaultLocale,
+        locale: response.locale,
+        regionLocale: response.regionLocale,
+    };
+}
+
+export const transformer = {
+    name: TRANSFORMER_NAME,
```
npeltier, line 134: please call this transformer (and file!) defaultLanguage, and use its init to do the network call, and the process to feed in whatever is needed

npeltier (changes requested):
thanks ! looks better, we are getting close to it. 
- we need defaultLanguage~Variation~ to be a proper transformer (name and init/process),
- there should be no duplicated logic in customize, no need to call default language as it has already been requested

npeltier (approved):
lgtm @seanchoi-dev :-) thanks for the patience

please make sure you get 100% coverage and i'm good

Axelcureno (approved):

npeltier:
Hi @seanchoi-dev , can you add better coverage

```
-------------------------|---------|----------|---------|---------|-------------------
File                     | % Stmts | % Branch | % Funcs | % Lines | Uncovered Line #s 
-------------------------|---------|----------|---------|---------|-------------------
All files                |   99.92 |    99.54 |     100 |   99.92 |                   
 fragment                |     100 |    99.07 |     100 |     100 |                   
  locales.js             |     100 |    98.36 |     100 |     100 | 401               
  pipeline.js            |     100 |      100 |     100 |     100 |                   
 fragment/transformers   |   99.83 |    99.49 |     100 |   99.83 |                   
  corrector.js           |     100 |      100 |     100 |     100 |                   
  customize.js           |     100 |      100 |     100 |     100 |                   
  defaultLanguage.js     |   98.54 |     97.5 |     100 |   98.54 | 119-120           
  fetchFragment.js       |     100 |      100 |     100 |     100 |                   
  promotions.js          |     100 |      100 |     100 |     100 |                   
  replace.js             |     100 |      100 |     100 |     100 |                   
  settings.js            |     100 |    99.05 |     100 |     100 | 255               
  wcs.js                 |     100 |      100 |     100 |     100 |                   
 fragment/utils          |     100 |      100 |     100 |     100 |                   
  cache.js               |     100 |      100 |     100 |     100 |                   
  common.js              |     100 |      100 |     100 |     100 |                   
  log.js                 |     100 |      100 |     100 |     100 |                   
  odinSchemaTransform.js |     100 |      100 |     100 |     100 |                   
  paths.js               |     100 |      100 |     100 |     100 |                   
 health-check            |     100 |      100 |     100 |     100 |                   
  index.js               |     100 |      100 |     100 |     100 |                   
-------------------------|---------|----------|---------|---------|-------------------
```

afmicka:
@seanchoi-dev one of the jira requirements was to fix the placeholder also in studio. 
Screenshot 2026-04-16 at 11 20 32

That part is not fixed. I still see FR placeholder on CH variation
https://mwpw-189894--mas--adobecom.aem.live/studio.html#content-type=merch-card&fragmentId=957fb7e0-eb40-4d62-a32c-59580f798354&locale=fr_FR&page=fragment-editor&path=sandbox
Screenshot 2026-04-16 at 11 18 08

seanchoi-dev:
@npeltier @afmicka 
Code coverage is now 100%
image

For the Studio preview fix, there is a separate PR: https://github.com/adobecom/mas/pull/760.

afmicka (approved):

PR #751, opened 2026-04-09
MWPW-192158: Shared item loading layer for bulk publish

Andrei4226:
Resolves https://jira.corp.adobe.com/browse/MWPW-192158

Extracts the shared item selection UI and data-loading logic from the Translation area into common/, making it reusable across features (Bulk Publish, Translations etc.). Decouples all UI components from Store.translationProjects via a new getItemsSelectionStore() abstraction. 
Adds portable data-loading utils, shared rendering helpers and full unit test coverage.

Please review all functionalities from the "Select Items" section in the Translation area.

## 🧪 Nala E2E Tests

### To run Nala tests again:

- Remove the `run nala` label

## Test URLs:

- After: https://mwpw-192158--mas--adobecom.aem.live/


studio/src/common/utils/item-loading.js
```diff
@@ -0,0 +1,398 @@
+import { Fragment } from '../../aem/fragment.js';
+import { CARD_MODEL_PATH, COLLECTION_MODEL_PATH } from '../../constants.js';
+
+export const OFFER_DATA_CONCURRENCY_LIMIT = 5;
```
honstar, line 4: Are these limits based on actual historical data where a higher concurrency is causing issues? Or any known limits we should be respecting?

studio/src/common/utils/item-loading.js
```diff
@@ -0,0 +1,398 @@
...
+            variation.path,
+            {
+                ...variation,
+                ...(getDisplayName ? { studioPath: getDisplayName(new Fragment(variation)) } : {}),
+                offerData: offerDataResults[index] ?? null,
+            },
+        ]),
+    );
+}
+/**
```
honstar, line 381: nit: please add a blank line ;)

honstar (approved):
LGTM

studio/src/common/utils/item-loading.js
Andrei4226: These values are carried over from the existing `translation-items-loader.js`
I reused here for consistency.
The translation area and the new bulk publish feature will use these files.
If there’s another known limit, I’ll update it accordingly.

honstar (approved):

studio/src/common/utils/translation-items-loader.js
```diff
@@ -1,8 +1,9 @@
...
-import { getService } from '../utils.js';
+import Store from '../../store.js';
+import { getItemsSelectionStore } from '../items-selection-store.js';
+import { CARD_MODEL_PATH, COLLECTION_MODEL_PATH, TABLE_TYPE } from '../../constants.js';
+import { Fragment } from '../../aem/fragment.js';
+import { getFragmentName } from '../../translation/translation-utils.js';
+import { getService } from '../../utils.js';
 
 const OFFER_DATA_CONCURRENCY_LIMIT = 5;
 const VARIATIONS_CONCURRENCY_LIMIT = 5;
```
yesil, line 9: `item-loading.js` exports `processConcurrently` and these same constants, but nothing in `src/` imports from it yet — only its own test does. Is the plan to migrate `translation-items-loader.js` to use `item-loading.js` in a follow-up, or is `item-loading.js` intended for a different consumer?

studio/src/common/utils/translation-items-loader.js
```diff
@@ -1,8 +1,9 @@
...
-import { CARD_MODEL_PATH, COLLECTION_MODEL_PATH, TABLE_TYPE } from '../constants.js';
-import { Fragment } from '../aem/fragment.js';
-import { getFragmentName } from './translation-utils.js';
-import { getService } from '../utils.js';
+import Store from '../../store.js';
+import { getItemsSelectionStore } from '../items-selection-store.js';
+import { CARD_MODEL_PATH, COLLECTION_MODEL_PATH, TABLE_TYPE } from '../../constants.js';
+import { Fragment } from '../../aem/fragment.js';
+import { getFragmentName } from '../../translation/translation-utils.js';
+import { getService } from '../../utils.js';
```
yesil, line 6: The name `translation-items-loader.js` is translation-specific, and this file still imports `getFragmentName` from `../../translation/translation-utils.js`. Moving it to `common/` without generalizing it is the worst of both worlds — it either belongs back in `translation/`, or it needs to be renamed and have the translation-specific dependency inverted (e.g. accept a `getDisplayName` callback, as `item-loading.js` already does).

studio/src/translation/mas-translation-editor.js
```diff
@@ -61,6 +62,7 @@ class MasTranslationEditor extends LitElement {
 
     async connectedCallback() {
         super.connectedCallback();
+        setItemsSelectionStore(Store.translationProjects);
```
yesil, line 66: `setItemsSelectionStore` is called in `connectedCallback` but never reset in `disconnectedCallback`. If a child component renders before this editor mounts (or after it disconnects), it will read from whatever store was last set. Worth either resetting on disconnect or documenting the assumed single-instance constraint.

studio/src/common/utils/translation-items-loader.js
```diff
@@ -58,11 +59,11 @@ async function processConcurrently(items, asyncFn, concurrencyLimit, batchSize =
  * Loads offer data for a fragment using its OSI field
  * @param {Object} fragment - Fragment object with fields
  * @param {AbortSignal} signal - Optional abort signal for cancellation
- * @param {Number} timeoutMs - Timeout in milliseconds (default: 10000)
+ * @param {Number} timeoutMs - Timeout in milliseconds
  * @returns {Promise<Object|null>} Offer data or null if not found/failed
```
yesil, line 63: The `(default: 10000)` annotation was dropped from the JSDoc. The default is still present in the function signature, but the doc hint was useful for callers — worth keeping.

studio/src/common/utils/translation-items-loader.js
Andrei4226: I think the plan is to migrate also `translation-items-loader.js` to use `item-loading.js`.
So it’s better to have it in this PR as well. 
Now, `translation-items-loader.js` imports the shared helpers directly from `item-loading.js`.

studio/src/common/utils/translation-items-loader.js
Andrei4226: Yes, you've right. I think it's better to rename it and have the translation‑specific dependency inverted. 
So the `getFragmentName` import is removed. All functions now accept `getDisplayName` as a callback parameter, following the same pattern as `item-loading.js`.

studio/src/translation/mas-translation-editor.js
Andrei4226: Very good point. Now, the previous store is saved on connect and restored on disconnect via `#itemsSelectionStoreSnapshot`.

studio/src/common/utils/translation-items-loader.js
Andrei4226: Restored the `(default: 10000)` annotation.

studio/src/common/utils/items-loader.js
```diff
@@ -62,7 +19,7 @@ async function processConcurrently(items, asyncFn, concurrencyLimit, batchSize =
  * @returns {Promise<Object|null>} Offer data or null if not found/failed
  */
 async function loadOfferData(fragment, signal, timeoutMs = 10000) {
-    const cache = Store.translationProjects.offerDataCache;
```
yesil, line 65: 🤖 This is almost identical to `loadOfferData` in `item-loading-browser.js` — the only difference is that this version reads the cache from `getItemsSelectionStore().offerDataCache`, while the shared one takes `cache` as a parameter. Could we import the shared one and pass the cache in? That way the new module isn't shadowed by a duplicate.

studio/src/common/utils/translation-items-loader.js
```diff
@@ -201,13 +152,22 @@ export function setCardVariationsByPaths(groupedVariationsByParentValue) {
...
  */
-function parseCardsFromStore(allFragments) {
-    return (allFragments || [])
-        .filter((fragment) => fragment.value.model.path === CARD_MODEL_PATH)
-        .map((fragment) => ({
-            ...fragment.value,
-            studioPath: getFragmentName(fragment.value),
-        }));
+function parseFragmentsFromStore(allFragments, getDisplayName) {
+    return (allFragments || []).reduce(
```
yesil: Same idea as the `loadOfferData` comment above — `item-loading.js` exports a `parseFragmentsFromStore` that does this exact thing (with an optional `getDisplayName`). Calling the shared one keeps everything in one place.

studio/src/common/utils/translation-items-loader.js
```diff
@@ -432,7 +398,7 @@ export async function loadSelectedFragments(selectedPaths, type, repository, opt
  * @param {AbortSignal} signal - Abort signal
  * @returns {Promise<Array<Object>>} Enriched cards
  */
-async function enrichCardsForViewOnly(cards, repository, signal) {
+async function enrichCardsForViewOnly(cards, repository, signal, getDisplayName) {
     const offerDataResults = await processConcurrently(
```
yesil: `enrichCards` in `item-loading.js` covers this case — pass empty maps for `existingOfferDataByPath` and `existingGroupedVariationsByPath` and you get the same behavior. We can delete this helper and call the shared one.

studio/src/common/items-selection-store.js
```diff
@@ -0,0 +1,14 @@
+import Store from '../store.js';
+
+let activeItemsSelectionStore = Store.translationProjects;
+
```
yesil: Small suggestion: defaulting this to `Store.translationProjects` keeps a translation-specific reference inside `common/`, which works against the goal of making this reusable. What if it defaulted to `null` and consumers had to call `setItemsSelectionStore(...)` explicitly? Easier to spot when someone forgets to set it.

Bigger picture (probably a follow-up, not this PR): a module-level singleton means two consumers can clobber each other. Threading the store as a Lit property through `<mas-items-selector>` would let multiple instances coexist — but that's a larger change.

studio/src/common/utils/item-loading.js
```diff
@@ -0,0 +1,399 @@
...
+    }
+
+    const enrichedCards = cards.map((card) => ({
+        ...card,
+        offerData: offerDataByPath.get(card.path) ?? null,
+        groupedVariations: groupedVariationsByPath.get(card.path) ?? [],
+    }));
+
+    if (enrichedCards.length > 50) {
+        await yieldToMain();
```
yesil, line 287: Tiny thing: `50` here is a bit magic. Worth pulling into a named constant at the top of the file (e.g. `LARGE_BATCH_YIELD_THRESHOLD`). The same `50` also appears in `translation-items-loader.js:237` — same constant.

studio/src/common/utils/item-loading.js
```diff
@@ -0,0 +1,399 @@
...
+ * @returns {Promise<Array>}
+ */
+export async function processConcurrently(items, asyncFn, concurrencyLimit, batchSize = 20) {
+    const results = new Array(items.length);
+    const executing = [];
+    let processedCount = 0;
+
+    for (let i = 0; i < items.length; i++) {
+        const promise = Promise.resolve().then(() => asyncFn(items[i], i));
+        results[i] = promise;
```
yesil, line 38: One subtlety worth a JSDoc note: `Promise.resolve().then(...)` schedules every task into the microtask queue immediately, before the concurrency check kicks in. The throttle works for awaited work, but any synchronous work inside `asyncFn` will still run in parallel. Almost certainly fine for our use case — just worth documenting so future readers don't get tripped up.

studio/src/translation/mas-collapsible-table-row.js
```diff
@@ -303,31 +306,35 @@ export class MasCollapsibleTableRow extends LitElement {
```
yesil, line 302: The new `getItemTypeLabel` helper in `render-utils.js` already does this exact mapping. Could we use it here (and in `mas-selected-items.getType`) so the new util actually gets adopted?

yesil (changes requested):

studio/src/common/utils/items-loader.js
Andrei4226: Removed local copy, now importing and using `loadOfferData` from `item-loading-browser.js` and passing the cache explicitly.

studio/src/common/utils/translation-items-loader.js
Andrei4226: Removed local copy, now importing and using `parseFragmentsFromStore` from `item-loading.js`.

studio/src/common/utils/translation-items-loader.js
Andrei4226: Removed `enrichCardsForViewOnly`, replaced with `enrichCards` from `item-loading.js`.

studio/src/common/items-selection-store.js
Andrei4226: Changed `default `to `null`, `getItemsSelectionStore()` now throws if store is not set. 
Added `allowUnset` option for the snapshot use case in `mas-translation-editor`.

studio/src/common/utils/item-loading.js
Andrei4226: Added the `LARGE_BATCH_YIELD_THRESHOLD` constant in `item-loading.js`, used in both files.

studio/src/common/utils/item-loading.js
Andrei4226: Added JSDoc NOTE to `processConcurrently` explaining the microtask queue behavior.

studio/src/translation/mas-collapsible-table-row.js
Andrei4226: `renderItemType` in `mas-collapsible-table-row.js` now delegates to `getItemTypeLabel`. 
Same for `getType` in `mas-selected-items.js`.

studio/src/common/utils/translation-items-loader.js
yesil: @Andrei4226 thanks for inverting the `getDisplayName` dependency — that part is great. The file itself wasn't renamed though, and the content is now fully generic (no translation-specific imports left). Could you rename it to something like `items-loader.js` so the filename matches what it actually does? Otherwise future readers will assume it's still translation-only.

studio/src/common/utils/items-loader.js
```diff
@@ -1,103 +1,18 @@
...
-        }
-    } catch (err) {
-        console.warn(`Failed to load offer data for fragment ${fragment.id}:`, err.message);
-        if (!signal?.aborted) {
-            cache.set(wcsOsi, null);
-        }
-        return null;
-    }
-}
+import Store from '../../store.js';
```
yesil, line 1: @Andrei4226 (re-posting on current HEAD since the original thread is now outdated) — thanks for inverting the `getDisplayName` dependency, that part is great. The file itself wasn't renamed though, and the content is now fully generic (no translation-specific imports left). Could you rename it to something like `items-loader.js` so the filename matches what it actually does? Otherwise future readers will assume it's still translation-only.

Andrei4226: Fixed.
I missed this renaming, now it’s much clearer.

yesil (approved):

npeltier (approved):
looks good to me, 

add following claude's review 
```
  Critical Issues (Must Fix)                                                                                                                                                                                     
                            
  1. Global ambient store — race condition on concurrent instances                                                                                                                                                  
  - studio/src/common/items-selection-store.js                                                                                                                                                                      
  - activeItemsSelectionStore is a module-level singleton. MasTranslationEditor uses #itemsSelectionStoreSnapshot to restore the previous value on disconnect. If a new instance connects before the old one        
  disconnects (route transitions), the restoration will null out the store the new instance already set.                                                                                                            
  - Fix: Use @lit-labs/context instead of the global singleton + snapshot/restore pattern.                                                                                                                          
                                                                                                                                                                                                                    
  2. loadSelectedFragments crashes when getDisplayName is undefined                                                                                                                                                 
  - studio/src/common/utils/items-loader.js                                                                                                                                                                         
  - getDisplayName(fragment) is called unconditionally in the processor loop. A caller that omits it gets TypeError: getDisplayName is not a function at runtime.                                                   
  - Fix: Either enforce it in JSDoc or provide a default: const { signal, onItems, getDisplayName = (f) => f?.path ?? '' } = options;                                                                               
                                                                                                                                                                                                                    
  ---                                                                                                                                                                                                               
  Should Fix                                                                                                                                                                                                        
                                                                                                                                                                                                                    
  3. getItemsSelectionStore() called multiple times per render                                                                                                                                                   
  - Called 2–4× in hot getter paths (selectedCount, showSelected, selectedInTable). Cache with const s = getItemsSelectionStore() locally or in connectedCallback.                                                  
                                                                                                                                                                                                                    
  4. JSDoc dangling comment in loadAllFragments                                                                                                                                                                     
  - A blank line between */ and export function loadAllFragments breaks the JSDoc association. Remove the blank line.                                                                                               
                                                                                                                                                                                                                    
  5. groupedVariationsData sync invariant undocumented                                                                                                                                                              
  - setCardVariationsByPaths implicitly keeps two store properties in sync. The original comment was removed. A future caller setting groupedVariationsByParent directly would silently break the invariant — add a 
  warning comment.                                                                                                                                                                                                  
                                                                                                                                                                                                                    
  6. .get() vs .value inconsistency in loadAllFragments                                                                                                                                                             
  - getItemsSelectionStore()[...].get()?.length uses .get() while the rest of the file uses .value. Inconsistent and fragile for future store consumers. 
```

you can eventually consider. Issue #1 should however not happen much as i don't see several invocation racing here. I'm still interested in mentionned lit context. We could have a look at it separately

afmicka (approved):

PR #760, opened 2026-04-13
MWPW-192450: Fix - Lingo placeholders in Studio card preview

seanchoi0:
Resolves https://jira.corp.adobe.com/browse/MWPW-192450

## 🧪 Nala E2E Tests

### To run Nala tests again:

- Remove the `run nala` label

## Test URLs:

- After: https://mwpw-192450--mas--adobecom.aem.live/

## More Test URL:
- https://mwpw-192450--mas--adobecom.aem.live/studio.html#content-type=merch-card&fragmentId=957fb7e0-eb40-4d62-a32c-59580f798354&locale=fr_FR&page=fragment-editor&path=sandbox


studio/src/mas-repository.js
```diff
@@ -646,7 +646,7 @@ export class MasRepository extends LitElement {
     async loadPreviewPlaceholders() {
         if (!this.search.value.path) return;
 
-        const cacheKey = `${this.filters.value.locale}_${this.search.value.path}`;
+        const cacheKey = `${Store.localeOrRegion()}_${this.search.value.path}`;
```
Axelcureno: ### Core change untested at the repository level

This is the foundational fix — switching from `this.filters.value.locale` to `Store.localeOrRegion()`. But every existing test in `mas-repository.test.js` stubs `loadPreviewPlaceholders` entirely, so this change (and the matching ones at lines 674, 677, 710) has zero direct unit test coverage.

A focused test could verify:
```javascript
it('uses Store.localeOrRegion() for dictionary cache key and fetch', async () => {
    // Set region override to simulate variation navigation
    Store.search.set((prev) => ({ ...prev, region: 'fr_FR' }));
    Store.filters.value = { locale: 'en_US' };
    // ... setup mocks for getDictionary ...
    await repository.loadPreviewPlaceholders();
    // Verify getDictionary was called with locale: 'fr_FR' (not 'en_US')
    // Verify cache key is 'fr_FR_sandbox' (not 'en_US_sandbox')
});
```

Without this, if someone reverts one of the 4 call sites but not the others, no test catches the inconsistency.

studio/src/mas-fragment-editor.js
```diff
@@ -701,6 +701,12 @@ export default class MasFragmentEditor extends LitElement {
 
         if (isVariationAfterContext && !skipVariation) {
             await this.#attachParentToCachedVariation(existingStore, fragmentPath);
+            const fragmentLocale = extractLocaleFromPath(fragmentPath);
+            if (fragmentLocale && fragmentLocale !== Store.localeOrRegion()) {
+                Store.search.set((prev) => ({ ...prev, region: fragmentLocale }));
```
Axelcureno, line 728: ### New cached-store locale reload path has no test coverage

These 6 new lines add locale-specific placeholder loading to `#initializeFromCachedStore` — this is the path Codecov flagged (4 missing lines, 71.42% patch coverage on this file).

The existing test at `mas-fragment-editor.test.js:358` only exercises `#initializeFromRepository` (non-cached). A test for this path would look like:

```javascript
it('reloads locale placeholders for cached variations when locale differs', async () => {
    const fragmentData = createFragmentData({ id: 'cached-var', locale: 'fr_FR', slug: 'var' });
    const fragment = new Fragment(fragmentData);
    const sourceStore = generateFragmentStore(fragment);
    Store.fragments.list.data.set([sourceStore]);
    Store.filters.value = { locale: 'en_US' };

    el.editorContextStore.isVariation.returns(true);
    sandbox.stub(el, 'resolveVariationParentFragment').resolves(null);
    Store.fragmentEditor.fragmentId.value = 'cached-var';

    await el.initFragment();

    expect(mockRepo.loadPreviewPlaceholders.calledOnce).to.be.true;
    expect(Store.search.get().region).to.equal('fr_FR');
});
```

studio/test/mas-fragment-editor.test.js
```diff
@@ -367,8 +366,7 @@ describe('MasFragmentEditor', () => {
 
             await el.initFragment();
 
-            expect(mockRepo.loadPreviewPlaceholders.callCount).to.equal(2);
-            expect(resolvePreviewSpy.calledOnce).to.be.true;
+            expect(mockRepo.loadPreviewPlaceholders.calledOnce).to.be.true;
```
Axelcureno: ### Weakened assertion — preview resolution no longer verified

The old test verified both `loadPreviewPlaceholders` call count AND `resolvePreviewFragment` being called. The new test only checks `calledOnce` and region.

In the new flow, preview resolution happens implicitly via `PreviewFragmentStore`'s constructor subscription to `Store.placeholders.preview`. This works, but is no longer tested for the variation-locale-differs path.

Consider adding a lightweight assertion that `Store.placeholders.preview` was set (which confirms the dictionary loaded), even if you don't spy on `resolvePreviewFragment` directly.

Axelcureno (changes requested):
## Review Summary

The architectural approach is sound — switching `loadPreviewPlaceholders` from `this.filters.value.locale` to `Store.localeOrRegion()` fixes the root cause (dictionary fetched/cached with the wrong locale for variations), and the "set region before load" reordering eliminates the unnecessary double-fetch. Both code paths (`#initializeFromCachedStore` and `#initializeFromRepository`) now handle variation locale detection consistently.

### What works well
- Fix is at the correct layer (repository/source, not consumer)
- Cache key now uses the effective locale — this was the actual root cause
- Region override is set *before* `loadPreviewPlaceholders()`, not after — cleaner and faster
- Remaining `this.filters.value.locale` uses in `mas-repository.js` are correctly left unchanged (DAM paths, fragment search, placeholder creation)
- All 11 CI checks green

### Requesting changes: Test coverage gaps

The PR's core behavioral change — switching the locale source in `loadPreviewPlaceholders` — has insufficient test coverage. See inline comments for specifics.

**Three tests needed:**

1. **Repository-level unit test** for `loadPreviewPlaceholders` verifying it uses `Store.localeOrRegion()` for cache key and fetch context (the foundational change in this PR has zero direct unit tests — all existing repo tests stub the method entirely)

2. **Cached store path test** for `#initializeFromCachedStore` when variation locale differs (the 4 new lines at `mas-fragment-editor.js:703-708` — this is what Codecov flagged)

3. **Same-locale no-op test** verifying that when `fragmentLocale === Store.localeOrRegion()`, no extra placeholder reload occurs (regression guard)

studio/src/mas-repository.js
```diff
@@ -707,7 +707,7 @@ export class MasRepository extends LitElement {
             preview: {
                 url: 'https://odinpreview.corp.adobe.com/adobe/sites/cf/fragments',
             },
-            locale: this.filters.value.locale,
+            locale: Store.localeOrRegion(),
```
Axelcureno: ### Suggestion: Capture locale once to avoid mid-execution drift

`Store.localeOrRegion()` is called 3 times during a single `loadPreviewPlaceholders` execution (lines 649, 674, 677) plus once in `fetchDictionary` (line 710). If an async operation modifies `Store.search.region` between these calls, the cache key check at line 674 could mismatch.

Not a blocker (in practice these are sequential), but a defensive improvement:
```javascript
async loadPreviewPlaceholders() {
    if (!this.search.value.path) return;
    const locale = Store.localeOrRegion();
    const cacheKey = `${locale}_${this.search.value.path}`;
    // ... use `locale` throughout instead of re-calling Store.localeOrRegion()
}
```

yesil (commented):
## Sequential placeholder loading introduces a latency cost on every fragment open

The sequential load introduced in this PR (get fragment → set region → load placeholders) solves the wrong-locale bug but adds ~200ms to every fragment open, not just cross-locale variations.

**Root cause of the constraint:** you need `fragment.path` before you know which locale to use for the variation, so the placeholder fetch cannot start until `getById` resolves. That dependency is real but only applies to the variation locale — the surface locale is already known before `getById` is called.

### Proposed approach: locale-keyed dictionary map

Instead of a single `Store.placeholders.preview` slot, maintain a map of dictionaries keyed by locale. The load side writes into the correct slot when a fetch completes. The render side reads whichever slot matches the current active locale, reactively — no coordination, no abort, no explicit `resolvePreviewFragment` call needed.

**Loading:** fire for the surface locale immediately on fragment open. After `getById` resolves and the variation locale is known, fire for that locale too. Both calls are fire-and-forget. If a locale was already cached from a previous visit, it resolves instantly.

**Rendering:** the preview consumer reads `previewByLocale[currentLocale]`. When the correct locale slot populates, the reactive system triggers the re-render automatically.

**No abort needed:** if the user navigates away before a fetch completes, the result lands in its slot harmlessly. The cache makes the next visit to that locale instant. You trade a small amount of unnecessary network completion for zero coordination code.

### Timeline

```
getById()               ░░░░░░░░░░░░░░░░░░░░▓
placeholders[en_US]     ░░░░░░░░░░░░▓              → preview renders immediately (surface locale)
placeholders[fr_FR]                     ░░░░░░░░░░▓  → preview re-renders with correct locale
```

Compare to the current PR:

```
getById()               ░░░░░░░░░░░░░░░░░░░░▓
                                             │
placeholders[fr_FR]                          ░░░░░░░░░░▓  → first render, correct but delayed
```

And the regression vs main for non-variation fragments (the common case):

```
// main branch
getById()               ░░░░░░░░░░░░░░░░░░░░▓
placeholders[en_US]     ░░░░░░░░░░░░▓              → already done by the time getById lands

// this PR
getById()               ░░░░░░░░░░░░░░░░░░░░▓
placeholders[en_US]                          ░░░░░░░░░░▓  → ~200ms added on every open
```

mirafedas (approved):

yesil (approved):

Axelcureno (approved):

afmicka (approved):

PR #764, opened 2026-04-15
MWPW-192698 - Lower batchSize and add rps limit to reduce number of concurrent requests sent to Odin

honstar:
- use default `batch_size` of 10 for all odin environments: qa, stage and prod
- add new rps limit, default 10, stage and prod use 20
- add wait time after processing a batch, calculated based on `rpsLimit` and `batchSize`
- add note that `batch_size` should never exceed `rps_limit`, otherwise initial burst of the first batch would directly exceed rps limit
- add unit-tests

Resolves https://jira.corp.adobe.com/browse/MWPW-192698

## 🧪 Nala E2E Tests

### To run Nala tests again:

- Remove the `run nala` label

## Test URLs:

- After: https://mwpw-192698--mas--adobecom.aem.live/


npeltier (approved):

Andrei4226 (approved):

yesil (commented):
**Throttle logic duplicated** — `versionTargetFragments` has the same 4-line throttle block (`minBatchMs`, `batchStart`, wait math) copied from `processBatchWithConcurrency`. `sendSyncRequests` delegates correctly; `versionTargetFragments` should too.

**`env_template` inconsistent with defaults** — `BATCH_SIZE=15 / RPS_LIMIT=15` diverges from code defaults (10/10) and workflow defaults (10/10). Should match or be explicitly called out as a local override.

**`batch_size` drop in merge workflow is implicit** — stage/prod now omit `batch_size`, silently relying on the workflow default of 10. An explicit `batch_size: 10` would make the intent self-documenting.

honstar:
> **Throttle logic duplicated** — `versionTargetFragments` has the same 4-line throttle block (`minBatchMs`, `batchStart`, wait math) copied from `processBatchWithConcurrency`. `sendSyncRequests` delegates correctly; `versionTargetFragments` should too.
> 
> **`env_template` inconsistent with defaults** — `BATCH_SIZE=15 / RPS_LIMIT=15` diverges from code defaults (10/10) and workflow defaults (10/10). Should match or be explicitly called out as a local override.
> 
> **`batch_size` drop in merge workflow is implicit** — stage/prod now omit `batch_size`, silently relying on the workflow default of 10. An explicit `batch_size: 10` would make the intent self-documenting.

Thanks @yesil , addressed!

yesil (approved):

afmicka (approved):

PR #774, opened 2026-04-17
MWPW-188972: Validate translation project task name format

Andrei4226:
Resolves https://jira.corp.adobe.com/browse/MWPW-188972

The translation project title is used as the task name when sending the project to the localization service (Odin loc v2 API). The new required format is:
- only letters, numbers, hyphens, underscores and dots are allowed
- at least one letter or number must be present
- consecutive dots are not allowed
- maximum length is 255 characters

Authors see an informative error message if the title doesn't meet these requirements.

## 🧪 Nala E2E Tests

### To run Nala tests again:

- Remove the `run nala` label

## Test URLs:

- After: https://mwpw-188972--mas--adobecom.aem.live/


studio/src/translation/translation-utils.js
```diff
@@ -3,6 +3,35 @@ import { FRAGMENT_STATUS } from '../constants.js';
 import Store from '../store.js';
 import { getFragmentPartsToUse, MODEL_WEB_COMPONENT_MAPPING } from '../editor-panel.js';
 
+export const ODIN_LOC_TASK_NAME_MAX_LENGTH = 255;
+
+/**
+ * Returns an error message when `value` is not a valid.
```
honstar: Typo: should read `is not a valid task name` or `is invalid`.

io/studio/src/translation/project-start-service.js
```diff
@@ -31,6 +52,14 @@ async function prepareProjectStart(params, options = {}) {
 
     const authToken = getBearerToken(params);
     const { projectCF, etag } = await getTranslationProject(params.projectId, authToken, params);
+    const translationTitle = (getValue(projectCF, 'title')?.value ?? '').trim();
+    if (!translationTitle) {
+        throw createProjectStartError(400, 'Translation project title is required.');
+    }
```
honstar: This is redundant since `getOdinLocTaskNameValidationError` also checks for an empty string, this check can be removed.

honstar (changes requested):

studio/src/translation/translation-utils.js
Andrei4226: Done.
I added it: "is not a valid task name".

io/studio/src/translation/project-start-service.js
Andrei4226: You're right.
I removed the function accordingly.

honstar (approved):

io/studio/src/translation/project-start-service.js
```diff
@@ -170,7 +196,7 @@ async function getTranslationData(authToken, projectCF, surface, translationFlow
     logger.info(`Translation flow: ${translationFlow}`);
 
     return {
-        title: getValue(projectCF, 'title')?.value || 'Untitled Project',
+        title: getValue(projectCF, 'title')?.value?.trim() || 'Untitled-Project',
```
honstar: The fallback to `Untitled-Project` is effectively dead code in the normal flow since the preceding `prepareProjectStart` ensures a proper title is set or throws otherwise.

honstar (changes requested):
Found one more issue.

io/studio/src/translation/project-start-service.js
Andrei4226: I removed the fallback to `Untitled-Project` accordingly.

honstar (approved):

mirafedas (approved):

io/studio/src/translation/project-start-service.js
```diff
@@ -18,6 +18,27 @@ const PATH_TOKENS = /\/content\/dam\/mas\/(?<surface>[\w-_]+)\/(?<parsedLocale>[
 const logger = Core.Logger('translation', { level: 'info' });
 const DEFAULT_BATCH_SIZE = 10;
 const DEFAULT_RPS_LIMIT = 10;
+const ODIN_LOC_TASK_NAME_MAX_LENGTH = 255;
+
+function getOdinLocTaskNameValidationError(value) {
```
afmicka, line 23: @Andrei4226 i see you have the exact same method in the translation-utils.js. This looks like a duplicate. Do we need to duplicate it or can the utils method be used here?

Andrei4226: I kept the duplicate because `translation-utils.js` imports Lit, store and editor code at the module level, so it can't be imported from an I/O action without pulling the entire client bundle into the Node runtime (or we could later extract a small shared package with no UI dependencies).

afmicka: ok...

afmicka (approved):

PR #786, opened 2026-04-23
MWPW-193512: Add express, acom-cc and acom-dc surfaces for CTA title issue

Andrei4226:
Resolves https://jira.corp.adobe.com/browse/MWPW-193512

Before: https://main--mas--adobecom.aem.live/studio.html?aem.env=qa#fragmentId=d3093ce5-19ab-4f2c-9e70-7289d20a8b4b&page=fragment-editor&path=express

After: https://mwpw-193512--mas--adobecom.aem.live/studio.html?aem.env=qa#fragmentId=27ba5786-faa0-4823-8b02-08622220c626&page=fragment-editor&path=express

## 🧪 Nala E2E Tests

### To run Nala tests again:

- Remove the `run nala` label

## Test URLs:

- After: https://mwpw-193512--mas--adobecom.aem.live/


studio/src/rte/ost.js
```diff
@@ -146,7 +146,9 @@ export async function onPlaceholderSelect(offerSelectorId, type, offer, options,
 
     const ctaText = CHECKOUT_CTA_TEXTS[options.ctaText]; // no placeholder key support.
     if (ctaText) {
-        attributes['text'] = ['acom', 'sandbox', 'nala'].includes(Store.search.get().path) ? `{{${options.ctaText}}}` : ctaText;
+        attributes['text'] = ['acom', 'acom-cc', 'acom-dc', 'sandbox', 'nala', 'express'].includes(Store.search.get().path)
```
Axelcureno: Quick sanity check — should `express` land in the sibling allowlist at line 282 too? That one still reads `['acom', 'acom-cc', 'acom-dc', 'sandbox', 'nala']` (no express). If a surface gets placeholder-keyed CTA text, it usually wants `modalsAndEntitlements` on as well. Worth confirming with PM before merge so express doesn't end up half-onboarded.

Axelcureno: Non-blocking follow-up thought: this allowlist now lives in three places (here, line 282, and `TRANSLATIONS_ALLOWED_SURFACES` in `constants.js:263`). Next time someone onboards a surface they'll have to hunt them all down. Could be worth extracting to a named constant like `PLACEHOLDER_CTA_SURFACES` — outside this PR's scope though.

studio/test/rte/ost.test.js
```diff
@@ -134,6 +136,26 @@ describe('onPlaceholderSelect', () => {
         expect(event.detail).to.deep.equal(expectedAttributes);
     });
 
+    it('should dispatch checkout link with placeholder text on acom-cc', () => {
```
Axelcureno: Nice regression test! Super minor — since the three new surfaces all hit the same branch, you could fold them into one parameterized block for free coverage:

```js
['acom-cc', 'acom-dc', 'express'].forEach((path) => {
    it(`should dispatch checkout link with placeholder text on ${path}`, () => {
        Store.search.set({ path });
        // ...rest identical
    });
});
```

Not blocking — just guards `acom-dc` and `express` from silently regressing.

Axelcureno (approved):
LGTM — small, scoped config change with a test. One thing to double-check before merge: the sibling list at line 282 doesn't include `express`, so express users get placeholder CTAs but no modals/entitlements. If that's intentional, ignore me. Otherwise a one-line addition there too. The other two comments are optional nits for future follow-up.

studio/src/rte/ost.js
Andrei4226: Good point!
I will check with team and if it's necessary, I will a do new commit based on your suggestion.

Andrei4226: Fixed. 
I created a new constant (PLACEHOLDER_CTA_SURFACES) and used it in our scope to make it easier to modify if we add new surfaces in the future.

studio/test/rte/ost.test.js
Andrei4226: Yes, I included your suggestion because it's better to cover all scenarios in the regression test based on my changes.

mirafedas (approved):

afmicka (approved):

PR #789, opened 2026-04-23
[MWPW-191561] - Enable gradient border color options in Studio editor

antonio-rmrz:
Closes https://jira.corp.adobe.com/browse/MWPW-191561

## Summary
- Add `gradient-purple-blue` and `gradient-firefly-spectrum` to the Fries variant's `borderColor.specialValues` in `FRIES_AEM_FRAGMENT_MAPPING`, mapped to the global CSS variables in `global.css.js` (hoisting `--gradient-firefly-spectrum` to `:root` alongside the existing `--gradient-purple-blue` so it can be reused).
- Apply the gradient using the dual-background `background-clip: padding-box, border-box` technique on the host. The gradient is confined to the border ring without affecting the interior background, badge anchoring, or layout.
- In `merch-card-editor.js`, gate the `colorArray` extension on `showAllSpectrumColors`. Previously the variant's border `specialValues` were appended to `colorArray` unconditionally and also spread into `options` via the `!showAllSpectrum` branch, producing duplicate entries in the border-color picker for variants that define `specialValues` without `showAllSpectrumColors: true`. This is a behavior change in shared editor code, but the dedup matches what the picker visibly does today.
- Matches gradient border behavior already available on Express and other templates in M@S Studio.

## Test URLs:

- Before: https://main--mas--adobecom.aem.page/studio.html#fragmentId=dfbd0ea4-425f-4f81-8322-dfbab87498d0&page=fragment-editor&path=commerce
- After: https://mwpw-191561--mas--adobecom.aem.page/studio.html#fragmentId=dfbd0ea4-425f-4f81-8322-dfbab87498d0&page=fragment-editor&path=commerce

[Pinata PR](https://github.com/adobecom/mas-pinata/pull/280)


Roycethan (changes requested):
@antonio-rmrz Badge is misplaced after applying the gradient color. plz check the vedio:
https://mwpw-191561--mas--adobecom.aem.page/studio.html#fragmentId=20f93e46-5796-4a5f-9315-30e20d86af45&page=fragment-editor&path=sandbox&tags=mas%3Avariant%2Ffries
https://github.com/user-attachments/assets/6431b36c-d480-4544-a064-5fd3d583cac2

Blainegunn (approved):

nala/studio/commerce/fries/tests/fries_gradient_border.test.js
```diff
@@ -0,0 +1,69 @@
...
+    for (const stop of stops) {
+        const r = parseInt(stop.slice(1, 3), 16);
+        const g = parseInt(stop.slice(3, 5), 16);
+        const b = parseInt(stop.slice(5, 7), 16);
+        expect(background).toContain(`rgb(${r}, ${g}, ${b})`);
+    }
+};
+
+test.describe('M@S Studio Commerce Fries gradient border test suite', () => {
+    for (const feature of features) {
```
Axelcureno: The test covers edit + discard but not save. Other fries features have a dedicated `_save.test.js` — please add one here that selects a gradient option, saves, reloads the fragment, and asserts both `border-color` and `gradient-border` attributes persist.

Axelcureno: Nit: all other fries tests use indexed access (`features[0]`, `features[1]`) with one `test()` call per case rather than a `for...of` loop. Not a blocker, but worth aligning for consistency.

studio/src/editors/merch-card-editor.js
```diff
@@ -2391,13 +2391,15 @@ class MerchCardEditor extends LitElement {
...
         const isBadgeBorderColor = dataField === 'badgeBorderColor' || dataField === 'trialBadgeBorderColor';
 
+        const showAllSpectrum = this.currentVariantMapping?.showAllSpectrumColors;
+
         let colorArray = Array.isArray(colors) ? colors : Object.keys(colors || {});
 
         let variantSpecialValues = {};
         if (this.fragment && isBorder && this.currentVariantMapping) {
             const variant = this.currentVariantMapping;
             variantSpecialValues = variant?.borderColor?.specialValues || {};
```
Axelcureno, line 2401: nit: Add a note to the PR summary about what this fixes? Moving `showAllSpectrum` up and gating `colorArray` extension on it prevents special values from appearing twice in the picker (once via `colorArray`, once via the `!showAllSpectrum` spread below) for variants without `showAllSpectrumColors: true`. Worth calling out since it's a behaviour change in shared editor code.

Axelcureno (changes requested):
Gradient CSS approach is clean and consistent with /. Hydration wiring picks it up automatically via the existing  reverse-lookup — no changes needed there. Two things before merge: a save test, and a clarification on the editor fix.

nala/studio/commerce/fries/tests/fries_gradient_border.test.js
antonio-rmrz: Added — new spec/test pair: `nala/studio/commerce/fries/specs/fries_gradient_border_save.spec.js` and `tests/fries_gradient_border_save.test.js`. Each variant clones the card, selects the gradient option, calls `studio.saveCard()`, and asserts both `border-color` and `gradient-border` attributes on the saved card, plus a computed-style check on the gradient ring.

I followed the in-place verification pattern from `fries_save.test.js` rather than navigate-and-reload — no other save test in nala does a fragment reload after `saveCard()`. Happy to add an explicit reload step if you'd rather have it.

antonio-rmrz: Switched to indexed access (`features[0]` / `features[1]`) with one explicit `test()` per case. `grep`-checked the broader nala suite — 39 files use indexed, 0 use `for...of`, so the loop was an outlier rather than just inconsistent within fries. Aligned.

studio/src/editors/merch-card-editor.js
antonio-rmrz: PR description updated to call out the dedup behavior change. While there I also corrected a stale line that referenced a `::before` pseudo-element technique — the implementation actually uses dual-background `background-clip: padding-box, border-box` on the host.

Axelcureno (approved):
LGTM. Failing Nala check is a pre-existing flake in the ACOM individuals suite, unrelated to this change. The dual-path `showAllSpectrum`/`!showAllSpectrum` logic in `#renderColorPicker` works correctly — worth simplifying in a follow-up to prevent silent regressions.

Roycethan (commented):
@antonio-rmrz Used this card in da.live and get this error:
https://main--da-cc--adobecom.aem.page/drafts/reuben/fries

Image

Roycethan (approved):

Roycethan:
> * https://mwpw-191561

followup is created

PR #793, opened 2026-04-28
MWPW-193624: Studio search & filter perf — in-memory narrowing, variant fast-path, multi-word phrase filter

Axelcureno:
Resolves https://jira.corp.adobe.com/browse/MWPW-193624

Improves Studio fragment search across four scenarios. All changes client-side; AEM-side limitations (50-item page cap, narrow `fullText` index, EDGES-only query mode) are preserved and confirmed via direct probe.

## Summary

- **In-memory narrowing** when adding a tag-picker selection (variant, tag, query, createdBy) on top of an already-loaded surface — fires only when cursor is exhausted and no edits since last load. Hard guards against partial-cursor and stale-data failure modes.
- **Variant fast-path**: a single variant tag-picker selection routes through `fullText.EDGES` (one round-trip) instead of streaming the entire surface and discarding 99% client-side.
- **Variant + query fast-path**: variant goes to AEM as `fullText`, the user's query is filtered client-side via `skipQuery` over an expanded haystack covering all string field values (cardTitle, description, shortDescription, etc.) — AEM only indexes title/description.
- **Multi-word query fast-path**: when no variant is selected, the longest single token goes to AEM and the full phrase is filtered client-side. Sidesteps AEM's AND-across-tokens behavior on multi-word queries.
- `#fillPage` returns after one cursor page (was: blocked on accumulating 10 post-filter items) → progressive rendering on variant-scarce surfaces.
- New `Events.fragmentSaved` + `disconnectedCallback` unsubscribe to track edits without leaks.

## Local benchmarks

| Scenario | Before | After |
|---|---|---|
| Tag-picker narrowing on loaded surface | 5–9s, 2–4 req | ~50ms, 0 req |
| `path=sandbox + variant=mini-compare-chart-mweb` | ~52s, 17 req | ~0.7s, 1 req |
| `... + query=Photoshop` | 0 cards, 38s | 4 cards, ~1s |
| `query="creative cloud"` | 0 cards | 39 cards, ~10s |

## Test URLs

- After: https://mwpw-193624--mas--adobecom.aem.live/studio.html

## Screenshots

Not applicable — no UI changes.

## Checklist

- [x] Tests pass locally (121/121, +22 new)


mirafedas:
Screenshot 2026-04-28 at 12 21 03
It shows '1 Filter' correctly, and for 2 or more should show '2 filter*s*', '3 filter*s*'

mirafedas:
Screenshot 2026-04-28 at 12 23 22
Not sure if this is in the scope of this Jira, please ignore if not - can we vertically align the checkbox labels to the center?

mirafedas (approved):

studio/src/mas-repository.js
```diff
@@ -349,6 +403,66 @@ export class MasRepository extends LitElement {
             .map((tag) => tag.replace(TAG_VARIANT_PREFIX, ''));
         tags = tags.filter((tag) => !tag.startsWith(TAG_STUDIO_CONTENT_TYPE) && !tag.startsWith(TAG_VARIANT_PREFIX));
 
+        const tracing = typeof localStorage !== 'undefined' && localStorage.getItem('mas-perf-trace');
```
honstar, line 426: Looks like localStorage is accessed for every `searchFragments` call. Maybe this should be moved outside of this function and rather evaluated once during module load time?

studio/src/mas-repository.js
```diff
@@ -270,6 +283,74 @@ export class MasRepository extends LitElement {
...
+        for (const field of item.fields || []) {
+            for (const value of field.values || []) {
+                if (typeof value === 'string') parts.push(value);
+            }
+        }
+        return parts.join('\n').toLowerCase();
+    }
+
+    /** Pass `query` already lowercased to avoid re-lowercasing once per item. */
+    skipQuery(query, item) {
```
honstar: Does this function need to be public?

studio/src/mas-repository.js
Axelcureno: Considered hoisting, but kept it per-call on purpose: this flag is meant for live toggling — set `mas-perf-trace` in DevTools, run a search, see timings, no reload. Hoisting at module load loses that. Cost is also negligible: `localStorage.getItem` is in-memory after first read, and `searchFragments` is debounced 50ms (not per-keystroke). Happy to revisit if a profile shows it.

studio/src/mas-repository.js
Axelcureno: Good catch — no external callers, and the sibling helpers (`#fillPage`, `#applyInMemoryFilter`, `#queryHaystack`, etc.) are all private. Privatized both `skipQuery` and `skipVariant` in 9b229fcc6 and folded the standalone unit test into the integration tests that already covered the cardTitle case.

Axelcureno:
@mirafedas Agreed, "2 Filters" / "3 Filters" reads better. Out of scope for this PR (perf-only, doesn't touch `mas-toolbar.js`) and the fix needs to hit two spots — the visible label *and* the `label="Filter"` ARIA attribute on the same button — so it's a touch larger than a one-liner. Filed [MWPW-193995](https://jira.corp.adobe.com/browse/MWPW-193995) as a standalone follow-up. Thanks for the catch!

Axelcureno:
@mirafedas On the checkbox alignment — out of scope here as you suspected. Filed [MWPW-193996](https://jira.corp.adobe.com/browse/MWPW-193996) so it doesn't get lost. Thanks!

honstar (approved):

Roycethan (approved):

Roycethan (commented):
@Axelcureno 
On empty caches times the search is returning no results hence few tests are failing:
Can you plz check?
ex: https://mwpw-193624--mas--adobecom.aem.live/studio.html#locale=fr_FR&page=content&path=nala&query=c72789db-f4c0-4b72-a6ba-3b73b05ae91a

Image
First time users accessing might get impacted ...

yesil:
@Axelcureno I'm not able to observe in memory narrowing with 

Scenario A: (cursor is not fully loaded, aka I did not scroll down)
1. select template 
2. wait for the first result
3. add created by filter

expected: in memory narrowing
actual: search request to Odin

Scenario B: (cursor is fully loaded, aka I did scroll down)
1. select template 
2. scroll all the way down to exhaust the cursor
4. add created by filter

expected: in memory narrowing
actual: empty result
image


I don't know in which exact scenario this improvement will be seen by the user but I find it more confusing now.

yesil (commented):
please see my comments.

Axelcureno:
@yesil Thanks for catching this — Scenario B is a real bug. The in-memory predicate read `item.createdBy`, but `Fragment` instances expose the field as `item.created.by` (only `Promotion` has a `createdBy` getter, and it returns the display name). `[...].includes(undefined)` → always false → empty list. Fix + regression test pushed in 514bdbd: read `item.created?.by`, lowercase both sides to mirror the server-side filter.

Scenario A is the intentional partial-cursor guard — narrowing an incomplete haystack would mis-filter, so we round-trip. Verified locally on NALA: 49 results, 0 search requests after the fix.

yesil:
@Axelcureno in which case users exhausts the cursor so that this feature can be used? I have the impression that this improvement would rarely if never be used.

Axelcureno:
> @Axelcureno in which case users exhausts the cursor so that this feature can be used? I have the impression that this improvement would rarely if never be used.

@yesil mostly 2 scenarios:

**When the user is toggling filters back and forth**: on the same loaded surface: 2nd, 3rd, 4th toggle = in-memory.
**When the user has scrolled deep down**: adding a filter at that point used to wipe everything and re-fetch from page. Now it doesn't.

Roycethan (approved):

PR #771, opened 2026-04-16
MWPW-193327 - [infra] make nala test GH runners more stable and robust

honstar:
Intermittent nala test failures on self-hosted runners were caused by two apt failure modes: unattended-upgrades holding dpkg/apt locks, and transient Ubuntu mirror sync errors during apt-get update.

Add a composite action (.github/actions/prep-apt) that stops background apt services, waits up to 120s for all locks to clear, then runs npx playwright install-deps with up to 3 retries (15s apart).

Apply the composite action to all three self-hosted workflows (run-nala.yml studio job, run-nala-daily.yml, run-nala-milolibs.yaml), replacing the previous unconditional bare install-deps call. Also adds continue-on-error: true to the Playwright cache steps in daily and milolibs, consistent with the studio job.

The GitHub-hosted docs job in run-nala.yml is unchanged — it does not share the self-hosted runner apt environment.

Resolves https://jira.corp.adobe.com/browse/MWPW-193327
QA Checklist: https://wiki.corp.adobe.com/display/adobedotcom/M@S+Engineering+QA+Use+Cases

Please do the steps below before submitting your PR for a code review or QA

- [ ] C1. Cover code with Unit Tests
- [ ] C2. Add a Nala test (double check with #fishbags if nala test is needed)
- [ ] C3. Verify all Checks are green (unit tests, nala tests)
- [ ] C4. PR description contains working Test Page link where the feature can be tested
- [ ] C5: you are ready to do a demo from Test Page in PR (bonus: write a working demo script that you'll use on Thursday, you can eventually put in your PR)
- [ ] C.6 read your Jira one more time to validate that you've addressed all AC's and nothing is missing

## 🧪 Nala E2E Tests

Nala tests run automatically when you **open this PR**.

### To run Nala tests again:

1. Add the `run nala` label to this PR (in the right sidebar)
2. Tests will run automatically on the current commit
3. Any future commits will also trigger tests **as long as the label remains**

### To stop automatic Nala tests:

- Remove the `run nala` label

> **Note**: Tests only run on commits if the `run nala` label is present. Add the label whenever you need tests to run on new changes.

## Test URLs:

- Before: https://main--mas--adobecom.aem.live/
- After: https://mwpw-NUMBER--mas--adobecom.aem.live/

PR #808, opened 2026-04-29
Revert "MWPW-193327 - [infra] make nala test GH runners more stable and robust"

honstar:
Reverts adobecom/mas#771


honstar:
Closing, need MWPW-xxxx branch name to trigger nala.

PR #812, opened 2026-04-30
MWPW-192308: Add ability to create a grouped variation of a collection

Andrei4226:
Resolves https://jira.corp.adobe.com/browse/MWPW-192308

Adds support for grouped variations on collections in MAS Studio.

## 🧪 Nala E2E Tests

### To run Nala tests again:

- Remove the `run nala` label

## Test URLs:

- After: https://mwpw-192308--mas--adobecom.aem.live/

Milo Page
https://main--cc--adobecom.aem.page/eg_en/drafts/mariia/coll?maslibs=MWPW-192308
https://main--cc--adobecom.aem.live/eg_en/drafts/mariia/coll?mas-io-url=https://14257-merchatscale-qa.adobeioruntime.net/api/v1/web/MerchAtScale


mirafedas (approved):

io/www/src/fragment/transformers/customize.js
```diff
@@ -108,6 +108,20 @@ function personalizationMatchScore(pznTags, { regionLocale, country, pzn }) {
     return matchedTokens * 100 + (regionMatch ? 20 : 0) + (countryMatch ? 10 : 0);
 }
 
+function pznTagIdsForMatching(variation) {
+    const fromField = variation.fields?.pznTags;
+    if (fromField != null && fromField !== '') {
```
yesil: No unit tests for `pznTagIdsForMatching` — the `fromField !== ''` guard, empty-array path, and string-vs-object tag fallback are all worth covering.

io/www/src/fragment/utils/odinSchemaTransform.js
```diff
@@ -2,6 +2,7 @@
 
 const CF_REFERENCE_FIELDS = ['cards', 'collections', 'entries', 'variations'];
 const REFERENCE_FIELDS = [...CF_REFERENCE_FIELDS, 'tags'];
+const COLLECTION_MODEL_PATH = '/conf/mas/settings/dam/cfm/models/collection';
 
 function collectPathToIdMap(references, map = {}) {
```
yesil, line 6: `COLLECTION_MODEL_PATH` is already exported from `studio/src/constants.js`. Keeping two copies in sync means a path change needs two edits.

io/www/src/fragment/utils/odinSchemaTransform.js
```diff
@@ -85,17 +86,20 @@ function transformReferences(body, pathToIdMap) {
         }
 
         const fields = transformFields(ref.fields, pathToIdMap);
+        const isCollection = ref.model?.path === COLLECTION_MODEL_PATH;
+        const value = {
+            name: ref.name,
```
yesil: The `isCollection` tag-spreading branch in `transformReferences` has no unit tests.

studio/src/mas-repository.js
```diff
@@ -1931,7 +1934,12 @@ export class MasRepository extends LitElement {
             fields: groupedFields,
         });
 
-        if (parentFragment.tags?.length) {
+        if (isCollection) {
+            const mergedTags = [...(parentFragment.tags || []), ...(pznTags || [])];
+            if (mergedTags.length) {
+                await this.aem.sites.cf.fragments.copyFragmentTags(newFragment, mergedTags);
```
yesil: `parentFragment.tags` entries are objects (`{ id, path, ... }`) while `pznTags` are plain strings — `copyFragmentTags` receives a heterogeneous array. Worth verifying it handles this mix, or normalizing to one type before merging.

studio/src/mas-fragment-variations.js
```diff
@@ -66,13 +67,24 @@ class MasFragmentVariations extends LitElement {
...
      * @param {Object} variationFragment
      * @returns {string}
      */
     getGroupedVariationTagsValue(variationFragment) {
         const pznTagsField = variationFragment.fields?.find((field) => field.name === 'pznTags');
-        const tags = pznTagsField?.values || [];
-        return tags.join(',');
+        const fromField = pznTagsField?.values || [];
+        if (fromField.length) return fromField.join(',');
+        const seen = new Set();
```
yesil: The same `mas:locale/`/`mas:pzn/` metadata fallback pattern is duplicated in `mas-fragment-editor.js:displayGroupedVariationInfo`. A shared helper would prevent drift.

yesil (changes requested):
See comments.

io/www/src/fragment/utils/odinSchemaTransform.js
Andrei4226: I duplicated the constant because `io/www` cannot import `studio/src/constants.js`, when running tests, that relative import is resolved incorrectly by the dev server, causing test failures.

3ch023 (changes requested):
creation works well:
https://mwpw-192308--mas--adobecom.aem.live/studio.html#fragmentId=08d15928-9c84-4fac-84fe-a238a4e6462b&page=fragment-editor&path=sandbox

but i believe we need to set pznTags field similar to how we do it for cards.
see this card example:
https://mwpw-192308--mas--adobecom.aem.live/studio.html#fragmentId=9df02a22-3340-41f8-8694-877550b86cc8&page=fragment-editor&path=sandbox

Image


after this i think variation will start rendering correctly:
https://main--cc--adobecom.aem.page/eg_en/drafts/mariia/coll
https://main--cc--adobecom.aem.live/eg_en/drafts/mariia/coll

io/www/src/fragment/transformers/customize.js
```diff
@@ -108,6 +108,20 @@ function personalizationMatchScore(pznTags, { regionLocale, country, pzn }) {
     return matchedTokens * 100 + (regionMatch ? 20 : 0) + (countryMatch ? 10 : 0);
 }
 
+function pznTagIdsForMatching(variation) {
```
3ch023: not sure about it, is it needed?

3ch023 (changes requested):
pznTagIdsForMatching likely doesn't do any impact

io/www/src/fragment/transformers/customize.js
Andrei4226: The `pznTagIdsForMatching` function is meant to extract tags from a variation
For cards, `fields.pznTags` is populated, so the new `variation.tags` fallback is unused
However, for collections, `fields.pznTags` doesn’t exist, so the function falls back to `variation.tags`
Should we need to have `pznTags` for grouped variations in collections (just like we do for cards), or should we want to using `variation.tags`?

yesil (commented):
should the breadcrumbs display "collection" in the path?
Navigation to parent does not work.
Is table Table view in the scope? How do you see a grouped variation of a collection in table view?

yesil (approved):

yesil (approved):

3ch023 (approved):

PR #829, opened 2026-05-05
feat(studio): bulk publish MVP UI — editor, lock, skeletons, advanced tools

Axelcureno:
Jira: https://jira.corp.adobe.com/browse/MWPW-188626

## Summary
- Hardcode `allowedClientId` as `'mas-studio'` on the backend; extract shared `isAllowed()` to `io/studio/utils.js` — removes client-controlled security bypass vector (Honwai review)
- Wire `mas-add-items-dialog` to `Store.bulkPublishProjects` via `setItemsSelectionStore`; pass `renderFragmentStatusCell` and call `loadAllCollections()` on open
- Fix `loadAllCollections()` to write to `getItemsSelectionStore()` instead of hardcoded `Store.translationProjects`
- Add missing store fields to `bulkPublishProjects` (`cardsByPaths`, `collectionsByPaths`, `placeholdersByPaths`, `offerDataCache`)
- Fix column alignment and status cell rendering in `mas-collapsible-table-row`
- Fix skeleton/empty-state flash on all tabs: `dataReady` flag (driven by `onReady` callback in `loadAllFragments` and `loaded` meta on `allCollections`) prevents "No items found." from appearing before first load completes

## Test URLs
- Before: https://main--mas--adobecom.aem.page/studio.html
- After: https://MWPW-188626--mas--adobecom.aem.page/studio.html


studio/src/bulk-publish/mas-bulk-publish-locales.js
```diff
@@ -0,0 +1,137 @@
+import { LitElement, html, nothing } from 'lit';
+import { styles } from './mas-bulk-publish-locales.css.js';
+
+const REGION_GROUPS = [
```
3ch023: we should be using locales.js file here.
otherwise 2 mapping will get our of sync, list of locales differ per surface

3ch023 (dismissed):
1. en_US is missing in the locale picker. 
2. lots of other regional locales are missing too. If this is expected please add a note in Jira description, so that QA/PdM know that there is a diff with figma
3. example URL is misleading for GWP, there is no way to publish www page via this tool. better to either remove the example or give a real one

Image

io/studio/src/bulk-publish/index.js
```diff
@@ -27,7 +27,7 @@ async function run(params) {
         }
 
         const requiredHeaders = ['Authorization'];
-        const requiredParams = ['paths'];
+        const requiredParams = ['paths', 'allowedClientId'];
```
honstar: Now that `allowedClientId` is a mandatory parameter when calling the `bulk-publish` action, I don't see that it actually gets passed in the corresponding `fetch` call.

studio/src/bulk-publish/bulk-publish-client.js
```diff
@@ -0,0 +1,47 @@
...
+
+    let response;
+    try {
+        response = await fetch(`${ioBaseUrl}${ENDPOINT}`, {
+            method: 'POST',
+            headers: {
+                Authorization: `Bearer ${token}`,
+                'Content-Type': 'application/json',
+            },
+            body: JSON.stringify({ paths, locales, concurrencyLimit }),
```
honstar: `allowedClientId` is not sent as parameter, hence the call will be rejected with a 400 response. Also, `concurrencyLimit` is not read by the backend either and silently ignored.

honstar (changes requested):

studio/src/bulk-publish/mas-bulk-publish-locales.js
Axelcureno: Fixed — extracted `REGION_GROUPS` to `io/www/src/fragment/locales.js` and replaced both local declarations (in `mas-bulk-publish-locales.js` and `mas-translation-languages.js`) with an import from there. The locales themselves still come from `getSurfaceLocales`/project data; `REGION_GROUPS` is only the display bucketing for the UI.

io/studio/src/bulk-publish/index.js
Axelcureno: Fixed — `allowedClientId` is now read from `window.adobeIMS?.getAccessToken()?.clientId` (falling back to `'mas-studio'`) and threaded through `startPublishing` → `publishBulk` → the POST body. Also removed `concurrencyLimit` from the client since the backend doesn't read it.

studio/src/bulk-publish/bulk-publish-client.js
Axelcureno: `allowedClientId` is now included in the POST body and `concurrencyLimit` has been dropped.

Axelcureno:
**Re: en_US / missing locales** — the picker shows locales already stored on the project; they're added via `mas-translation-languages` which uses `getSurfaceLocales(surface)`. If locales are missing there, it's a separate issue — can you confirm which surface/locales are expected?

**Re: example URL** — fixed, now shows a real MAS Studio URL (`studio.html#query=<fragment-id>`).

@3ch023

io/studio/src/bulk-publish/index.js
honstar: I need to revise my initial change request, sorry about that. Letting the client provide both the token and clientID for which it was issued actually weakens the intended security model: by controlling both any client can send a request to the bulk publish API. Instead, the backend should check that the issued token is for `mas-studio` exclusively.
See https://github.com/adobecom/mas/blob/main/io/studio/src/translation/project-start.js#L120 and your initial implementation https://github.com/adobecom/mas/blob/main/io/studio/src/bulk-publish/index.js#L54.
This is actually a good opportunity to move the `isAllowed` function to a common utils file :)

studio/src/bulk-publish/mas-bulk-publish-editor.js
```diff
@@ -0,0 +1,613 @@
...
+    }
+
+    async publish() {
+        await this.#withPendingAction(QUICK_ACTION.PUBLISH, async () => {
+            if (this.isNewProject) await this.saveBulkProject();
+            const { startPublishing } = await import('./bulk-publish-store.js');
+            const { publishBulk } = await import('./bulk-publish-client.js');
+            const paths = this.items.filter((i) => i.status === 'valid').map((i) => i.path);
+            const token = window.adobeIMS?.getAccessToken()?.token;
+            const allowedClientId = window.adobeIMS?.getAccessToken()?.clientId ?? 'mas-studio';
```
honstar: See above, the client should not provide both the token and the allowed client ID to the API. Instead, the backend will need to ensure that the issued token is for the clientID `mas-studio`.

honstar (changes requested):

io/studio/src/bulk-publish/index.js
Axelcureno: Done —  is now in `io/studio/utils.js` (shared by both bulk-publish and translation/project-start). Both backends hardcode `'mas-studio'` and no longer accept a client-supplied `allowedClientId`. The parameter is also removed from the frontend request body end-to-end.

studio/src/bulk-publish/mas-bulk-publish-editor.js
Axelcureno: Fixed — removed from the editor, store, and client. The backend now hardcodes the clientId.

studio/src/bulk-publish/mas-bulk-publish-locales.js
```diff
@@ -0,0 +1,89 @@
+import { LitElement, html, nothing } from 'lit';
+import { styles } from './mas-bulk-publish-locales.css.js';
+import { REGION_GROUPS } from '../../../io/www/src/fragment/locales.js';
```
seanchoi0: Importing directly from io/www may not be a good long-term approach. Should we consider creating a neutral place for sharing instead?

studio/src/bulk-publish/mas-bulk-publish-editor.js
```diff
@@ -0,0 +1,611 @@
...
+                Store.bulkPublishProjects.inEdit.set(store);
+                showToast('Project duplicated successfully.', 'positive');
+            } catch (err) {
+                console.error('Failed to duplicate bulk publish project:', err);
+                showToast('Failed to duplicate the project.', 'negative');
+            }
+        });
+    }
+
+    async validate() {
```
seanchoi0, line 489: validate() has no abort/version guard; overlapping runs or navigation away could theoretically apply stale results. Please consider a small follow-up similar to the editor’s AbortController pattern.

seanchoi0 (changes requested):
Overall, it looks good. Please fix the unit test errors. I also left a few comments for you to consider.

studio/src/bulk-publish/mas-bulk-publish-locales.js
Axelcureno: Good call — added `studio/src/locales.js` as a neutral re-export of `io/www/src/fragment/locales.js`. Updated `mas-bulk-publish-locales`, `mas-translation-languages`, and `mas-locale-picker` to import from there instead.

studio/src/bulk-publish/mas-bulk-publish-editor.js
Axelcureno: Fixed — added a `#validateId` counter. Each run captures its ID at entry; async callbacks bail before writing results if a newer run has started.

studio/src/bulk-publish/mas-bulk-publish-editor.js
```diff
@@ -0,0 +1,611 @@
...
+                setFieldValue: (k, v) => {
+                    fields[k] = v;
+                },
+            });
+        }
+    }
+
+    disconnectedCallback() {
+        super.disconnectedCallback();
+        this.#abortController?.abort();
```
seanchoi0, line 119: I think we need to bump validateId here in case the user navigates away while validation is in flight.
`        this.#validateId++;
`

seanchoi0 (changes requested):
Just one more small thing to change.

studio/src/bulk-publish/mas-bulk-publish-editor.js
Axelcureno: Good catch — added `this.#validateId++` to `disconnectedCallback` so any in-flight validation run is invalidated when the component unmounts.

seanchoi0 (approved):

io/studio/src/bulk-publish/index.js
```diff
@@ -51,7 +50,7 @@ async function run(params) {
         }
 
         const authToken = getBearerToken(params);
-        const allowed = await isAllowed(authToken, params.allowedClientId);
+        const allowed = await isAllowed(authToken, 'mas-studio');
```
honstar: Please restore the initial code since `allowedClientId` is configured in and read from a central location for all mas-studio actions (see https://github.com/adobecom/mas/blob/main/io/studio/app.config.yaml#L11).

io/studio/src/translation/project-start.js
```diff
@@ -20,7 +20,7 @@ async function main(params) {
         }
 
         const authToken = getBearerToken(params);
-        const allowed = await isAllowed(authToken, params.allowedClientId, ims);
+        const allowed = await isAllowed(authToken, 'mas-studio', ims);
```
honstar: Same as above, please leave `params.allowedClientId`, it is read from https://github.com/adobecom/mas/blob/main/io/studio/app.config.yaml#L11.

honstar (changes requested):

honstar:
@Axelcureno : please resolve conflicts.

io/studio/src/bulk-publish/index.js
Axelcureno: Done — restored `params.allowedClientId` in both `bulk-publish/index.js` and `project-start.js`. The `isAllowed` extraction to `utils.js` is still in place.

io/studio/src/translation/project-start.js
Axelcureno: Done — same fix applied here.

honstar (approved):

Roycethan (changes requested):
@Axelcureno Feedbacks:
1) 'Advance tools' is overlapping over 'support'

Image

2)Duplicate project throws error:

Image

3)'Publish' 'Revert' 'Delete' are not implemented ?
figma:
Image

4)The third column in here is titled as "Actions" in figma, currently missing:

Image
Figma:
Image

5)Can we have this per figma?
image

Figma: 
image


6)We should not allow adding duplicates in either  by enter url or by search 

Image

7) "Add by search" button needs to have border around it and Correct font size and color format:

Image
Figma:
Image

8)In figma i see validation of 5 results and 2 errors and remove option, we don not have this now....

Image

9)In Locales section , Users directly prompted with Edit , even though theres is not locales added:

Image
Figma:

Image

10)I dont see "My saved views"  in Select locales window.

11)Its allowing me to save with Invalid urls 

Image

12) Save button is not grayed out after saving the project

13)The query generated by the "Add by search" wizard below is added and validated but opening that query yields no results:
https://mas.adobe.com/studio.html#content-type=merch-card&page=content&path=sandbox&query=e7563f87-7cf6-4126-b395-e4dc78b9099e


14) Edit the project and add a new item and save , after save trying publish again , but its not allowing me...
image

15)'Download report' is missing
Figma:
image

16) I'm able to publish already published fragment , there should be some way to block this and show users message saying fragment is already published... Draft and Modified ones only should be able to publish.

Roycethan (changes requested):
@Axelcureno Thanks for fixing previous ones ... plz take a look into these as well:

1) Plz add the warning icon on this below modal, it exist in figma:

Image

2)We are allowing publishing the project without saving it .... say if project is modified and user tried to publish - its getting published , but its not getting saved.

3) Deleting from wizard from the below window,

Image

Its get stuck here and not moved to home screen

Image

4)Search (with keyword) on Collections and Placeholder tab is not filtering out the results, works only for Fragments 

5)Plz update the 'Add by search' button to below figma:

Image

6) We are still allowing already published fragments to publish again , this should be blocked, since they are already published and no changes are made

afmicka:
@Axelcureno  additional observation to the @Roycethan's above

1. en_US is missing in all surfaces in the language picker. In ACOM and Sandbox for sure. If you create a project and go to select languages to publish, there is no en_US.
 
Screenshot 2026-05-11 at 14 22 14

2. If you start creating the project in one surface, the picker for surface is enabled and you can change the surface within the project create editor. Should it be? It will ask you to discard the changes but when you click to change the surface, but looks like nothing happens except the surface selection. The screen stays in the editor with all the values entered 

3. Search for fragments, collections and placeholders do not work. It was mentioned above but also for fragments. It starts working but then reloads all the fragments over the search results (see video) 

https://github.com/user-attachments/assets/92b854ab-a5cc-4168-adf4-88db2c0fdd80

Roycethan (changes requested):
@Axelcureno Thanks , few last things which are open:
1)Yes Mili is right in a way, Search in Fragment doesn't work at the first instance , its keep on loading the results and stops filtering on Search term... Second time when i do again it works... we should fix this....
See video: 

https://github.com/user-attachments/assets/9297b15e-69e8-4fb9-b826-df04cc9406c4

2) On Collections and Placeholders tab ,  the filtering dropdowns needed ? I didnt see them in Translations project ... But if needed then they should work , currently they are not ... 

Image

3) Plz fix the Nala failure and Formattting checks.

Roycethan (approved):

Roycethan:
> @Axelcureno Thanks , few last things which are open: 1)Yes Mili is right in a way, Search in Fragment doesn't work at the first instance , its keep on loading the results and stops filtering on Search term... Second time when i do again it works... we should fix this.... See video:
> 
>  Screenshare.-.2026-05-11.11_50_30.AM.mp4 
> 2. On Collections and Placeholders tab ,  the filtering dropdowns needed ? I didnt see them in Translations project ... But if needed then they should work , currently they are not ...
> 
> Image
> 3. Plz fix the Nala failure and Formattting checks.

Fixed

PR #845, opened 2026-05-07
MWPW-180778: Implement promotions transformer

npeltier:
- promo transformer now fetches and caches the promotions folder (TTL + localStorage in preview), selects the active project by surface/date/geo/tag, 
- grab variations for active projects for default & region language if any,
- it sets promoCode on matching offer fragments at process time,
- Add findPromoVariation to customize.js so promo variations take priority over regional/pzn variations
- Add promotions folder stub to setupFragmentMocks so pipeline tests don't time out
- Add mas.instant param in preview only to test in future / past
- Add unit tests for all new code paths (100% coverage)
- Move e2e tests of pipeline in a dedicated test file,
- Add full e2e test including promotions

Resolves https://jira.corp.adobe.com/browse/MWPW-180778

## Test URLs:

- [active promo project with instant setup that reorder sidenav of a collection, rewrite a card, and set promo code on 2 cards](https://main--milo--adobecom.aem.page/drafts/npeltier/mep-and-mas/mas-promo-collection?maslibs=MWPW-180778&instant=2026-04-02&debug.io#category=labelokokok&filter=all)

example is a bit lame because the promocode does not work on price but you can still see wcs is called with that promo code.
note that number of test urls is should be huge, but we'd better wait for authoring tools to be here so it's easier to setup

### mas/io
- non regression on: [nala docs plans page (with regional variation when no promo is active)](https://mwpw-180778--mas--adobecom.aem.live/web-components/docs/plans-collection.html?country=GR&language=en&mas-io-url=https://14257-merchatscale-npeltier.adobeioruntime.net/api/v1/web/MerchAtScale)
- non regression on us plans page https://main--cc--adobecom.aem.live/creativecloud/plans?mas-io-url=https://14257-merchatscale-npeltier.adobeioruntime.net/api/v1/web/MerchAtScale
- non regression on fr plans page https://main--cc--adobecom.aem.live/fr/creativecloud/plans?mas-io-url=https://14257-merchatscale-npeltier.adobeioruntime.net/api/v1/web/MerchAtScale
- non regression on catalog page https://main--cc--adobecom.aem.page/products/catalog?mas-io-url=https://14257-merchatscale-npeltier.adobeioruntime.net/api/v1/web/MerchAtScale


io/www/src/fragment/utils/common.js
```diff
@@ -199,6 +199,25 @@ async function getRequestInfos(context) {
...
+ * or null if neither matches.
+ * Tags are CQ tag paths (e.g. /content/cq:tags/mas/geo/en_US), matched by final path segment.
+ * Falls back to extracting country from regionLocale when country is not provided.
+ * @param {string[]} tags
+ * @param {{ regionLocale?: string, country?: string }} param1
+ * @returns {{ region: boolean, country: boolean } | null}
+ */
+function matchesGeo(tags, { regionLocale, country }) {
+    const effectiveCountry = country ?? regionLocale?.split('_')[1];
+    const region = Boolean(regionLocale && tags.some((tag) => tag.toLowerCase().endsWith(`/${regionLocale.toLowerCase()}`)));
```
Axelcureno: I see the JSDOC says "matched by final path segment", although previously it was checking for `pzn/country/<country>` and now it just checks for `/<country>`. That means now a tag like `cq:tags/foo/bar/US` (unrelated to pzn) would now match for US. Wondering if we should make the check more resilient?

io/www/src/fragment/transformers/promotions.js
```diff
@@ -1,20 +1,379 @@
...
+    const instant = toInstant(context.preview ? context['mas.instant'] : undefined);
+    const { locale, country, regionLocale } = context;
+    const effectiveRegionLocale = regionLocale ?? locale;
+
+    let active = null;
+    let matchCount = 0;
+    for (const project of projects) {
+        if (matchesProject(project, { surface, locale, country, regionLocale: effectiveRegionLocale, instant }, context)) {
+            matchCount++;
+            if (!active) active = project;
```
Axelcureno, line 281: `active` is whichever project the folder listing returned first from Odin response. I'm guessing it's expected?

web-components/test/aem-fragment.test.html.js
```diff
@@ -534,6 +534,89 @@ runTests(async () => {
...
+
+                const AemFragment = customElements.get('aem-fragment');
+                const urlStub = sinon
+                    .stub(AemFragment.prototype, 'getFragmentClientUrl')
+                    .returns(blobUrl);
+
+                const originalUrl = window.location.href;
+                const url = new URL(window.location.href);
+                if (masInstantValue != null) {
+                    url.searchParams.set('mas.instant', masInstantValue);
```
yesil: can we use `instant` parameter, which is already supported in Milo? Also, I propose to add instant attribute on `mas-commerce-service` for use cases where users cannot use `instant` parameter.

yesil (approved):

io/www/src/fragment/transformers/promotions.js
```diff
@@ -1,20 +1,379 @@
...
+    }
+    if (project.startDate && instant < new Date(project.startDate).getTime()) {
+        logDebug(
+            () =>
+                `Project "${project.name}" skipped: instant ${new Date(instant).toISOString()} is before startDate ${project.startDate}`,
+            context,
+        );
+        return false;
+    }
+    if (project.endDate && instant > new Date(project.endDate).getTime()) {
```
Axelcureno: Anchoring here because the public-facing risk lives at this call site: `context.preview` ends up truthy on any public request with `?preview=true` since params spread straight into the pipeline context. Two things break — `promotions.js:388` reads `context['mas.instant']` so anyone can pin date and preview future or expired promos, and this `localStorage.getItem(...)` hits a runtime that has no `localStorage` and throws `ReferenceError` → 500.

Gate on `context.preview?.url` instead of truthy — only the studio fragment-client builds the object form, server callers can't fake it.

io/www/src/fragment/transformers/promotions.js
```diff
@@ -1,20 +1,379 @@
...
+        startDate: fields?.startDate ?? null,
+        endDate: fields?.endDate ?? null,
+        tags: fields?.tags ?? [],
+        offerLines: fields?.offers ?? [],
+    }));
+
+    return cacheProjects(context.preview, projects);
+}
+
+function toInstant(value) {
```
Axelcureno, line 104: `new Date("lol").getTime()` is `NaN`, and `NaN < startDate` is `false`, so every project passes the date window. Combined with the `?preview=true` thing above, a single query string activates the whole promotions folder.

```js
const t = new Date(value).getTime();
return Number.isFinite(t) ? t : Date.now();
```

Worth a fixture for `'lol'`, `'2026-13-99'`, `null`.

io/www/src/fragment/transformers/promotions.js
```diff
@@ -1,20 +1,379 @@
...
 
-async function init(context) {
-    return await getActivePromotion(context);
+/**
+ * Builds a flat OSI → promoCode lookup map from project-level promoCode and offer overrides.
+ * Project-level promoCode is the default wildcard ('*').
+ * Overrides can target specific OSIs or act as wildcards (empty osis list).
+ * Specific OSI overrides take priority over the wildcard.
+ */
+function buildPromoMap(offerOverrides, country, projectPromoCode) {
```
Axelcureno: Wildcard-vs-wildcard collision isn't defined anywhere. A typo like `:OOPS:` in offerOverrides silently overwrites `projectPromoCode` for every fragment — with the 5min cache + CDN swr=86400 that's a 24h blast radius.

Two things: `log()` at INFO when `map['*']` gets set so SRE can spot it in Splunk, and a test pinning project-default vs override-wildcard precedence.

io/www/test/fragment/promotions.test.js
```diff
@@ -0,0 +1,570 @@
...
+            const result = await promotionsTransformer.init(createContext({ regionLocale: 'en_US' }));
+            expect(result.status).to.equal(200);
+            expect(result.activeProject.id).to.equal('proj-1');
+            expect(result.activeProject.fragmentPaths).to.have.length(1);
+            expect(result.activeProject.promoCode).to.equal('SAVE20');
+        });
+
+        it('supports mas.instant for time-travel testing in preview mode', async () => {
+            const storage = {};
+            globalThis.localStorage = {
```
Axelcureno: The localStorage shim gets installed inline, then `clearPromoCache(true)` + `delete globalThis.localStorage` run *after* the assertions (also at lines 319 and 355). Any failing `expect()` leaks both into the next test, and `afterEach` only resets the in-memory cache.

`try/finally`, or pull the preview tests into their own `describe` with proper setup/teardown. Otherwise a real bug here turns into confusing failures three tests later.

io/www/src/fragment/transformers/promotions.js
```diff
@@ -1,20 +1,379 @@
...
+    } else {
+        projectsCache = undefined;
+        promoVariationsCache = {};
+    }
+}
+
+function getCachedProjects(preview) {
+    const cacheEntry = preview ? JSON.parse(localStorage.getItem('promotions')) : projectsCache;
+    if (cacheEntry) {
+        cacheEntry.isExpired = Date.now() - cacheEntry.timestamp > CONFIG_CACHE_TTL;
```
Axelcureno: If `timestamp` is in the future (preview author with a skewed clock, or NTP jump) the diff goes negative, `isExpired` becomes false, entry lives forever.

```js
cacheEntry.isExpired = Math.abs(Date.now() - cacheEntry.timestamp) > CONFIG_CACHE_TTL;
```

io/www/src/fragment/transformers/customize.js
```diff
@@ -138,8 +134,31 @@ function findPersonalizationVariation(variations, customizeContext) {
...
+}
+
 function mergeVariations(root, customizeContext) {
     const { isRegionLocale } = customizeContext;
+    // Promo variation takes priority, independent of fields.variations
+    const promoVariation = findPromoVariation(root, customizeContext);
+    if (promoVariation) {
+        logDebug(() => `Merging promo variation ${promoVariation.id} for fragment ${root.id}`, customizeContext);
+        const merged = deepMerge(root, promoVariation);
+        merged.id = root.id;
```
Axelcureno, line 158: `merged.id` gets restored but `merged.path` doesn't — the variation's promo-folder path (`/content/dam/mas/<surface>/<locale>/promotions/<project>/...`) overwrites the original. Anything downstream that re-parses `customizedRoot.path` (e.g. `customizeTree` at line 259) now sees the promo path.

`merged.path = root.path;` next to the id line.

io/www/test/fragment/pipeline-config.test.js
```diff
@@ -3,15 +3,20 @@ import sinon from 'sinon';
...
+import { clearPromoCache } from '../../src/fragment/transformers/promotions.js';
+import { createResponse } from './mocks/MockFetch.js';
 import { MockState } from './mocks/MockState.js';
 
 let fetchStub;
 
 describe('pipeline configuration caching', () => {
     beforeEach(() => {
-        fetchStub = sinon.stub(globalThis, 'fetch');
+        fetchStub = sinon.stub(globalThis, 'fetch').callsFake(() => {
```
Axelcureno: Catch-all 404 silently eats unexpected fetches. `pipeline-e2e.test.js` at least `console.warn`s the URL — this file doesn't, so a transformer making an unintended call passes invisibly.

Match the warn line from `pipeline-e2e.test.js:27`, and ideally fail the test if the spy fires.

Axelcureno (commented):
Multi-agent follow-up review — 7 inline comments below (3 blocking, 4 high). B3 (matchesGeo regression) and H4 (multi-match ordering) already covered in earlier comments, dropped to avoid duplicating. Mediums and lows tracked separately, none merge-blocking.

Axelcureno (approved):

io/www/src/fragment/transformers/promotions.js
Axelcureno: This one looks real and potentially merge-blocking.

io/www/src/fragment/transformers/promotions.js
Axelcureno: If the `?preview=true` one gets fixed, this becomes more of a defensive fix.

io/www/src/fragment/transformers/promotions.js
npeltier: future promotions projects should not be published in advance. Preview calls will not help you as they are hitting VPN gated odinpreview. I will still add a check for preview parameter in case it's somehow passes in io to fail early and avoid unnecessary computing

afmicka:
@npeltier not sure if related to the failed IO deployment and checks, but with your mas-io-changes, the grouped variations of the fragments are not shown inside the collections (regional variations DO show)
Check: https://mwpw-180778--mas--adobecom.aem.live/web-components/docs/plans-collection.html?country=GR&language=en&mas-io-url=https://14257-merchatscale-npeltier.adobeioruntime.net/api/v1/web/MerchAtScale

Creative Cloud in the first collection and Adobe Express in the second collection should show grouped card variations. Compare with https://main--mas--adobecom.aem.live/web-components/docs/plans-collection.html?country=GR&language=en

Screenshot 2026-05-18 at 12 31 24
Screenshot 2026-05-18 at 12 31 19


I do see regional variations displaying correctly though.

npeltier:
> Creative Cloud in the first collection and Adobe Express in the second collection should show grouped card variations. 

for now @afmicka promo variation takes over grouped variation & regional variation. We will want to discuss things i guess with business, but not in that first iteration

afmicka:
> > Creative Cloud in the first collection and Adobe Express in the second collection should show grouped card variations.
> 
> for now @afmicka promo variation takes over grouped variation & regional variation. We will want to discuss things i guess with business, but not in that first iteration

as discussed offline, in this example there is no promo applied (no instant parameter used nor we are in the set promo period (april 2026)). It breaks the grouped variation outside of the promo setup.

afmicka (approved):

PR #846, opened 2026-05-08
MWPW-194457: mas-compare-chart editor

yesil:
Resolves 
 - https://jira.corp.adobe.com/browse/MWPW-194457
 - https://jira.corp.adobe.com/browse/MWPW-190754
 
## 🧪 Nala E2E Tests

### To run Nala tests again:

- Remove the `run nala` label

## Test URLs:

- After: https://mwpw-194457--mas--adobecom.aem.live/studio.html#fragmentId=c481dd2a-a079-47d7-9383-3dbeda5b0575&page=fragment-editor&path=sandbox

## Express test URL:
- https://main--da-express-milo--adobecom.aem.page/drafts/ilyas/pricing?milolibs=mwpw-194457--milo--adobecom&maslibs=mwpw-194457--mas--adobecom
### IE variation (first card)
- https://main--da-express-milo--adobecom.aem.page/drafts/ilyas/pricing?milolibs=mwpw-194457--milo--adobecom&maslibs=mwpw-194457--mas--adobecom&akamaiLocale=IE


studio/src/aem/aem-tag-picker-field.js
```diff
@@ -368,6 +377,15 @@ class AemTagPickerField extends LitElement {
         });
     }
 
+    addContentTypeTags() {
```
yesil, line 375: why this is needed?

io/www/src/fragment/transformers/customize.js
```diff
@@ -29,6 +30,11 @@ async function resolveFragmentInit(context, requestInfos) {
 }
 
 function deepMerge(...objects) {
+    // variation template is empty
+    if (objects?.[0]?.fields?.variant === COMPARE_CHART_VARIANT) {
+        const [parent, child] = objects;
+        child.fields.features.value = [...(parent?.fields?.features.value || []), ...(child?.fields?.features.value || [])];
```
Andrei4226: `child.fields.features` is not null-guarded, if `child.fields.features` is `undefined` this throws `TypeError`

io/www/src/fragment/transformers/customize.js
```diff
@@ -44,6 +50,10 @@ function deepMerge(...objects) {
             }
         }
     }
+    if (objects?.[0]?.fields?.template === 'mas-compare-chart') {
+        const features = objects.flatMap((obj) => (Array.isArray(obj?.fields?.features) ? obj.fields.features : []));
```
Andrei4226: In this block, we only handle `features` as a plain array
Compare-chart payloads in `compare-chart-express.json` use AEM shape: features: `{ value: [...] }`
Could we also handle `features.value` here?

studio/src/editors/mas-compare-chart-editor.js
```diff
@@ -0,0 +1,2439 @@
...
+
+    #emitValue(table) {
+        if (!this.fragmentStore || !table) return;
+        const newValue = table.outerHTML;
+        this.#ensureCompareChartTag();
+        this.fragmentStore.updateField(COMPARE_CHART_FIELD, [newValue]);
+        this.requestUpdate();
+
+        this.dispatchEvent(
+            new CustomEvent('change', {
```
Andrei4226: Nit: we already have `EVENT_CHANGE = 'change'` in `constants.js`

Andrei4226 (changes requested):
awesome work, just few comments to confirm👍

io/www/src/fragment/transformers/customize.js
yesil: this is only for runtime/preview rendering and since we moved to Freyja preview, features array is the only supported shape.

studio/src/editors/mas-compare-chart-editor.js
yesil: done

Axelcureno:
Still marked as WIP in the description. Is that intentional?

io/www/src/fragment/transformers/customize.js
Axelcureno: Heads up — `parent?.fields?.features.value` guards `parent` and `.fields` but then hits `.features.value` without the `?.`. A compare-chart variation that overrides other fields but **omits** `features` will throw a `TypeError` here and 500 the fragment. (Verified live: the IE variation that *does* carry features renders fine, so this only bites the features-less variation case.) Worth guarding both sides: `parent?.fields?.features?.value`. Different spot than the :54 thread.

io/www/src/fragment/transformers/customize.js
Axelcureno: Minor: the variant branch up at L36 reads `features.value` (object) and this template branch reads `features` (plain array) — two shapes for the same field in one function. The renderer's coercer handles both so nothing breaks today, but it's an easy footgun. Maybe a one-line comment noting which kind hits which branch.

studio/test/editors/mas-compare-chart-editor.test.html
```diff
@@ -0,0 +1,338 @@
...
+                    const path = String(resource);
+                    if (path.startsWith('/test/mocks/offers.json') || path.startsWith('/test/mocks/sites/')) {
+                        return originalFetch(`/__wds-outside-root__/1/web-components${path}`, options);
+                    }
+                    return originalFetch(resource, options);
+                });
+
+            mockIms();
+            mockLana();
+            await mockFetch(withWebComponentsMocks(withWcs), withWebComponentsMocks(withAem));
```
Axelcureno: These two tests (`renders three cards…` and `uses mas-items-selector…`) time out at 5s locally — `aem-tag-picker-field` calls `loadTags()` on render and this fixture's `mockFetch` doesn't stub the tags endpoint, so it rejects and the editor never settles. Stubbing the tags fetch here should green them. (Repro'd in full suite: 2104 passed / 2 failed.)

web-components/src/hydrate.js
```diff
@@ -397,13 +397,81 @@ function processMnemonicElements(htmlContent) {
...
+            return;
+        }
+        const p = doc.body.querySelector('p[name]');
+        if (p) {
+            container.append(p);
+            return;
+        }
+        // Trust boundary: features payload originates from AEM-authored
+        // fragments via IMS-gated write path. Do not call with untrusted input.
+        container.insertAdjacentHTML('beforeend', value);
```
Axelcureno, line 461: The `p[name]` fast-path + trust-boundary comment are nice. One thought: the fallback still `insertAdjacentHTML`'s raw author HTML — same trust level as prices/description so not new, but if a payload isn't `p[name]`-shaped maybe reject it rather than inject raw? Optional.

Axelcureno (commented):
Did a team pass over this with a few agents + a local build/run. Render looks good — pulled it up on the Express IE page and the compare-chart + regional variation both render clean (EUR/VAT pricing, "Free Var" title, feature glyphs all correct), no console errors. Two things worth a look before C2/C3: the `customize.js` features merge can NPE on a variation that omits `features` (latent, didn't hit it live since the test variation carries features), and the 2 editor tests are timing out locally on an unstubbed tags fetch. Rest are small nits inline. Nice feature.

studio/src/editors/mas-compare-chart-editor.js
```diff
@@ -0,0 +1,2440 @@
...
+            rowEl.innerHTML = innerHtml;
+        });
+    }
+
+    #findRow(table, groupName, rowName) {
+        const groupEl = table.querySelector(`:scope > div[name="${CSS.escape(groupName)}"]`);
+        return groupEl?.querySelector(`:scope > p[name="${CSS.escape(rowName)}"]`) || null;
+    }
+
+    #addGroup() {
```
Axelcureno, line 902: Big file, and the mutation/responsive/override methods (`#addGroup`, `#removeRow`, `#handleCellChange`, `#enterMobile`/`#exitMobile`, `#isFeatureOverridden`/`#writeColumnFeatures`) don't have unit coverage yet — that's the most regression-prone logic in the PR. Worth a few tests before C2/C3.

web-components/src/svg/plus-sign.svg.js
```diff
@@ -0,0 +1,18 @@
+import { html } from 'lit';
```
Axelcureno: This and the two sibling SVGs (`black-bg-minus`, `down-arrow-filter`) don't have any importers in src — looks like the toggles use CSS icons instead. Dead on arrival, can probably drop all three.

web-components/src/compare-chart-table-parser.js
```diff
@@ -0,0 +1,100 @@
+export const normalizeCompareChartKey = (value) => {
```
Axelcureno, line 1: `normalizeCompareChartKey` here and `normalizeKey` in the editor produce different slugs for the same label (`a-b` vs `a--b`). Disjoint markup paths today so it's latent, but if table-format and div-format groups ever mix, row-key matching silently drops cells. Reusing one normalizer would close it.

studio/src/aem/aem-tag-picker-field.js
npeltier: who are you asking this to :) ?

web-components/AGENTS.md
```diff
@@ -0,0 +1,86 @@
...
+
+- Write concise, imperative commit messages (e.g., "Add user auth flow").
+- One logical change per commit.
+
+## General
+
+- Do NOT add code "just in case." Only write code that is needed now.
+- Do NOT add backwards-compatible shims unless explicitly asked.
+- When modifying existing code, match the surrounding style.
+- Prefer deleting dead code over commenting it out.
```
npeltier: please remove i'm not against having one or any of the rules in it, just this deserves dedicated PR

io/www/src/fragment/transformers/customize.js
```diff
@@ -29,6 +30,11 @@ async function resolveFragmentInit(context, requestInfos) {
 }
 
 function deepMerge(...objects) {
+    // variation template is empty
```
npeltier: what does this mean?

studio/AGENTS.md
```diff
@@ -61,6 +61,22 @@ Rules for AI agents working in this codebase.
...
+- For HTML test pages under `/test`, open `http://localhost:2024/test/<file>?wtr-test-file=<encoded-test-path>`, where `<file>` is the path relative to `/test` and `<encoded-test-path>` is the URL-encoded form of `/test/<file>?wtr-manual-session=true`.
+- Example: for `/test/editors/mas-compare-chart-editor.test.html`, open `http://localhost:2024/test/editors/mas-compare-chart-editor.test.html?wtr-test-file=%2Ftest%2Feditors%2Fmas-compare-chart-editor.test.html%3Fwtr-manual-session%3Dtrue`.
+- For JS-only WTR test files without an HTML page, open `http://localhost:2024/?wtr-test-file=<encoded-test-path>`, where `<encoded-test-path>` is the URL-encoded form of `/test/<file>?wtr-manual-session=true`.
+- If opening the test URL fails because nothing is reachable on port `2024`, stop and report exactly: `The unit test runner is not reachable at http://localhost:2024. Please start it, then tell me to continue.`
+
```
npeltier: same here, we should probably move this off current PR

web-components/src/variants/compare-chart.js
```diff
@@ -0,0 +1,213 @@
...
+ * Wrappers (not inner <slot>) are measured so bordered chips match row mates.
+ */
+const HEIGHT_SYNC_TARGETS = [
+    { key: 'header', selector: '.seg-header' },
+    { key: 'price', selector: '.seg-price' },
+    { key: 'detail', selector: '.seg-detail' },
+    { key: 'cta', selector: '.seg-cta' },
+];
+
+export class CompareChart extends VariantLayout {
```
npeltier: shouldn't we call this a CompareChartColumn?

npeltier:
why do you need specific IO process for features? it's not empty when it's empty?

io/www/src/fragment/transformers/customize.js
yesil: done

web-components/src/compare-chart-table-parser.js
yesil: done

web-components/src/svg/plus-sign.svg.js
yesil: done

studio/test/editors/mas-compare-chart-editor.test.html
yesil: done

studio/src/aem/aem-tag-picker-field.js
yesil: I was self reviewing, but I got interrupted.
I was trying something ok :D

io/www/src/fragment/transformers/customize.js
yesil: @npeltier the variant field(template) of variation fragments are empty.
is this by design?
I put the comment to remember it.

web-components/src/variants/compare-chart.js
yesil: Small details IMO, but if it is a column, I would store it in mas-compare-chart fragment directly.
We will keep calling it a card when we give demo to GWP, e.g: create a compare-chart card, add it, create a variation, etc.
if I say create a compare chart column, it can sound configusing no?

studio/AGENTS.md
yesil: done

io/www/src/fragment/transformers/customize.js
yesil: thx, removed this block, it was a slop.

yesil:
@npeltier 
>why do you need specific IO process for features? it's not empty when it's empty?

features field in variation is not an override but delta content compared to parent.
so in customize.js I merge them instead of applying the regular override logic.

studio/src/editors/mas-compare-chart-editor.js
yesil: let me know if these are blockers.

web-components/src/hydrate.js
yesil: let me know if these are blockers.

io/www/src/fragment/transformers/customize.js
npeltier: it's by design that variations have nothing in their fields but what they vary on.

web-components/src/variants/compare-chart.js
npeltier: compare chart sounds equally (if not more?) confusing to me. And with that template, concretely you are designing a column (header + features)

io/www/src/fragment/transformers/customize.js
```diff
@@ -29,6 +30,15 @@ async function resolveFragmentInit(context, requestInfos) {
...
+    if (objects?.[0]?.fields?.variant === COMPARE_CHART_VARIANT) {
```
npeltier: i wonder if we shouldn't rather have a `mergeMode` registry of template/property for arrays, where you can choose wether you want to merge or override.

io/www/src/fragment/transformers/customize.js
```diff
@@ -3,6 +3,7 @@ import { getRequestInfos, matchesGeo } from '../utils/common.js';
 import { logDebug } from '../utils/log.js';
 
 const PZN_FOLDER = '/pzn/';
+const COMPARE_CHART_VARIANT = 'compare-chart';
```
npeltier: please use the word template

yesil: done

io/www/src/fragment/transformers/customize.js
yesil: @npeltier I tried to find a better solution but nothing brilliant.  Is it ok we defer it to post PR?
Since you work on customize.js a lot, you might have a better idea.

web-components/src/variants/compare-chart.js
yesil: I agree, renamed it to CompareChartColumn.

Axelcureno (approved):

io/www/src/fragment/transformers/customize.js
```diff
@@ -29,6 +30,15 @@ async function resolveFragmentInit(context, requestInfos) {
 }
 
 function deepMerge(...objects) {
+    // variation template is empty
+    if (objects?.[0]?.fields?.variant === COMPARE_CHART_TEMPLATE) {
+        const [parent, child] = objects;
+        const parentFeatures = parent?.fields?.features?.value || [];
+        const childFeatures = child?.fields?.features?.value || [];
+        if (child?.fields?.features && (parentFeatures.length || childFeatures.length)) {
```
honstar: `child` is the second `objects` arg — for promo/regional/pzn merges it is the variation taken from `customizeContext.references`. Mutating `child.fields.features.value` in place poisons that reference for any later sibling that merges the same variation (siblings share references).
Clone the field before mutating, or build the merged features into the returned result map after the existing loop.

studio/src/mas-fragment-editor.js
```diff
@@ -1351,7 +1405,24 @@ export default class MasFragmentEditor extends LitElement {
...
+            if (compareChartEditor) {
+                compareChartEditor.commitFeatureDrafts();
+                dirtyCardFragmentStores = compareChartEditor.dirtyCardFragmentStores.filter(
+                    (store) => store.get?.().hasChanges,
+                );
+            }
+            if (dirtyCardFragmentStores.length) showToast('Saving fragment...');
+            for (const cardFragmentStore of dirtyCardFragmentStores) {
+                const savedCard = await this.repository.saveFragment(cardFragmentStore, false);
+                if (!savedCard) return;
```
honstar, line 1482: If saving the fragment fails for whatever reason, this `return` would leave the user with a stuck toast message `Saving fragment...`. Instead, surface a separate toast with a meaningful error message instead.

studio/src/mas-fragment-editor.js
```diff
@@ -1351,7 +1405,24 @@ export default class MasFragmentEditor extends LitElement {
...
+            }
+            Store.editor.referencedFragmentStoresHaveChanges.set(false);
+            const savedFragment = await this.repository.saveFragment(this.fragmentStore, !dirtyCardFragmentStores.length);
+            if (dirtyCardFragmentStores.length && savedFragment) {
+                showToast('Fragment successfully saved.', 'positive');
```
honstar, line 1487: Similar to the above, if `savedFragment` is falsy, no toast with any error message would be shown to the user.

studio/src/mas-repository.js
```diff
@@ -1580,6 +1655,7 @@ export class MasRepository extends LitElement {
             if (!savedFragment) throw new Error('Invalid fragment.');
 
             fragmentStore.refreshFrom(savedFragment);
+            await initFragmentCache();
```
honstar, line 1266: If the cache is already initialized this is a cheap no-op, but if it triggers (re)hydration mid-save it could race with the subsequent `fragmentCache.add(new Fragment(savedFragment))` call. Confirm this is idempotent under concurrent saves (compare-chart save kicks off multiple `saveFragment` calls in a loop).

studio/src/common/components/mas-items-selector.js
```diff
@@ -57,7 +67,69 @@ class MasItemsSelector extends LitElement {
...
+        store.displayCards.set(upsert(store.displayCards.value));
+        store.allCards.set(upsert(store.allCards.value));
+        return fragment;
+    }
+
+    #appendSelectedCard(fragment) {
+        if (!fragment?.path) return;
+        const store = getItemsSelectionStore();
+        const selectedCards = store.selectedCards.value || [];
+        if (selectedCards.includes(fragment.path) || selectedCards.length >= this.maxSelectedCards) return;
```
honstar: no-ops without toast when `selectedCards.length >= this.maxSelectedCards`. Consider showing a toast so the user understands why the paste did nothing.

web-components/src/mas-compare-chart.js
```diff
@@ -0,0 +1,1288 @@
...
+                    : String(raw).trim()
+                : `${DEFAULT_STICKY_OFFSET}px`;
+        this.style.setProperty('--compare-chart-sticky-offset', offset);
+    }
+
+    #setupSticky() {
+        this.#teardownStickyHeader();
+        this.#setStickyTopOffset();
+        this.#applyStickyOffset();
+        window.addEventListener('scroll', this.#boundSyncStickyHeader, true);
```
honstar: Capture-phase scroll listeners fire for every scrollable element on the page. The handler is debounced via `requestAnimationFrame`, so cost is bounded, but if a page contains multiple compare-charts (or a chart sits inside a heavy app like Studio's preview iframe) you'll get `N×scrolls` of work per frame. Consider scoping to the nearest scroll container, or using `IntersectionObserver` for the "is the header stuck" decision.

studio/src/common/components/mas-search-and-filters.js
```diff
@@ -4,20 +4,37 @@ import { VARIANTS } from '../../editors/variant-picker.js';
...
-import { toAttribute } from '../../aem/tag-path-utils.js';
+import '../../aem/aem-tag-picker-field.js';
 
 const MAS_TAG_NAMESPACE = '/content/cq:tags/mas';
-const CUSTOM_TAG_ROOT = `${MAS_TAG_NAMESPACE}/custom/`;
+const VARIANT_TAG_PREFIX = 'mas:variant/';
+const EMPTY_TAGS_BY_TYPE = {
+    offer_type: [],
+    plan_type: [],
+    market_segment: [],
```
honstar: Are these duplicate tag namespaces by any chance? On odin prod author, I only see `/content/cq:tags/mas/market_segments`.

honstar (changes requested):

io/www/src/fragment/transformers/customize.js
npeltier: i would also make that change more generic: have an hardcoded object
```
const MERGE_CONFIG = {
   "COMPARE_CHART_TEMPLATE": { 'arraysToMerge' = ['features'] }
}
``` 
and then something like
```
const arraysToMerge = MERGE_CONFIG[fields?.variant]?.arrayMerging;
arrayToMerge?.forEach(...
```

studio/src/editors/mas-compare-chart-editor.js
```diff
@@ -0,0 +1,2446 @@
...
+    #renderCardStatus(fragment) {
+        const status = fragment?.status || '';
+        const variant = this.#statusLightVariant(status);
+        return html`
+            <span class="compchart-card-status ${status.toLowerCase()}">
+                <sp-status-light size="s" variant=${variant}></sp-status-light>
+                <span>${status ? status[0] + status.slice(1).toLowerCase() : 'Unknown'}</span>
+            </span>
+        `;
+    }
```
npeltier, line 558: #renderXxx methods violate CLAUDE.md — mas-compare-chart-editor.js has 8+ private #renderGroup/#renderRow/#renderCell/#renderCardsSection/etc. Project convention is getters for HTML sections.

;-)

npeltier (commented):
in addition to Honwai's PR, i'm left with comments: 
Missing unit tests for #effectiveFeatureValues override behavior, #updateCellFeature / #stageColumnFeatures / #writeColumnFeatures round-trip, #moveRow /
  #moveGroup drag-reorder, and discardCardFragmentChanges / commitFeatureDrafts pairing. The single big integration test won't cover the 85% branch threshold
  cleanly.

studio/src/editors/mas-compare-chart-editor.js
yesil: @npeltier render is ok if the method has parameters.

render is not ok if it can be a pure getter.

studio/src/mas-repository.js
honstar: Claude flagged this as false alarm ;)

studio/src/common/components/mas-search-and-filters.js
yesil: renamed tag name space to /content/cq:tags/mas/market_segments

honstar:
The changes look good to me. Only outstanding thing is the issues regarding a potentially stuck toast message in case a card save is not successful.

npeltier (approved):
thanks!

Andrei4226 (approved):

honstar (approved):
I'm not holding up this major PR with the minor issue regarding potentially stuck toast messages. This is rather a minor issue compared to the overall feature.

afmicka:
@yesil 
1. i am not able to create compare chart on your branch. What am i missing?

Screenshot 2026-06-08 at 14 01 43

2. When i create a translation project and i pick the items for it and i click Add selected files, i can the discard pop-up. The discard should not popup here (see video for the flow)
Screenshot 2026-06-08 at 14 33 09


https://github.com/user-attachments/assets/dd83f8b6-d43a-48ee-a931-d7550f782467

afmicka:
@yesil i am seeing some more issues:

1. When i was creating Compare Chart Column, when i chose the template and offer, i got offer not resolved in the preview. Then i refreshed the page and it worked
Screenshot 2026-06-08 at 16 23 30
Screenshot 2026-06-08 at 16 23 24


2. These changes are breaking editor around Promotion code field 
Screenshot 2026-06-08 at 16 25 03
Screenshot 2026-06-08 at 16 23 58

3. Editor fields expand its width to the maximum screen's space, pushing the preview to the very right. Is that intentional?
Screenshot 2026-06-08 at 16 35 11


4. Regular merch card collections are being opened in the full screen editor and in the compare chart template. They should open in the floating editor with different set of fields. If you go back in browser, you will see the floating editor but cannot close it. The close button does not work. This might be because compare chart is also seen as a merch-card-collection. These need to be distinguished. 
https://mwpw-194457--mas--adobecom.aem.live/studio.html#page=content&path=sandbox&query=cea1632b-af47-4ad5-a563-17e66a7fd1dd

Screenshot 2026-06-08 at 16 40 28

should be: 
Screenshot 2026-06-08 at 16 40 53

close button does not work: 
Screenshot 2026-06-08 at 16 41 49

5. When i edit the compare chart and save, then i switch to the preview, end then back to Edit, if i keep editing, the save button is not getting enabled (see video). The same happens if i only save the fragment but keep editing. 

https://github.com/user-attachments/assets/2e796cfc-4dbb-4378-9389-b06f3d71a5d0

6. Editor fields are overlapping the discard modal and i am not able to click on any button (Discard nor Cancel), they are inactive
https://mwpw-194457--mas--adobecom.aem.live/studio.html#fragmentId=9713e893-6180-492f-a053-27302d18517c&page=fragment-editor&path=sandbox
Screenshot 2026-06-08 at 17 25 36

7. Show/hide selected items toggle in translation project has changed the styles and behavior of the side nav. On main it toggles the width of the table (completely hides the sidenav) and also has the greyed out background the full height of the side nav. With the changes on the branch, the space for sidenav and table are fixed (not toggling) and the sidenav background is only behind the added fragments.
Screenshot 2026-06-08 at 17 36 18
Screenshot 2026-06-08 at 17 35 58

on prod: 

Screenshot 2026-06-08 at 17 36 41

Screenshot 2026-06-08 at 17 44 38

afmicka:
@yesil 
1. regional variations (Locale tab) should not be exposed in Translation project
Screenshot 2026-06-09 at 15 40 42

2. buttons should be on the right side in translations picker

Screenshot 2026-06-09 at 15 52 43

afmicka:
@yesil 
1. toggle in promotions item picker does not toggle sidenav. Nothing happens when you click selected itemr 
Screenshot 2026-06-09 at 16 15 52

Screenshot 2026-06-09 at 16 15 47

2. Don't know why but i have noticed that when you click on the tab to duplicate, it does not open the same link where you are but the main content page in sandbox. If you are in the fragment or translations, it will duplicate tab with opening in the main fragment content page in sandbox
Screenshot 2026-06-09 at 16 14 57

afmicka:
@yesil  there is one bug left. 
When you open the compare-chart and click back in the browser, you land on the content page with collection editor opened that can not be closed. 
https://mwpw-194457--mas--adobecom.aem.live/studio.html#page=content&path=sandbox&query=02b1b0c8-9a8c-4ccb-92bb-677d418320c5

afmicka:
@yesil 
1. The size field in the editor pops up when Discard is opened
Screenshot 2026-06-10 at 11 14 03

2. The item picker in promotions and translations are not loading fragments on Edit. Select some items, and then click on Edit to add more
Screenshot 2026-06-10 at 11 31 02
Screenshot 2026-06-10 at 11 24 23

PR #854, opened 2026-05-12
MWPW-194867: Add cleanup-variations IO action and extract locales to JSON

mirafedas:
Adds a new \`cleanup-variations\` IO action that removes invalid variation locale references from AEM fragments. Also extracts the locale arrays hardcoded in \`locales.js\` into a shared \`locales.json\` file so the Node.js action and browser-side code use the same source of truth.
To keep the response manageable, it is recommended to run the script for one locale at a time.

Follow-up Jira to remove the locale.js duplication: https://jira.corp.adobe.com/browse/MWPW-199498 

Follow-up Jira for refactoring: https://jira.corp.adobe.com/browse/MWPW-199821

Resolves https://jira.corp.adobe.com/browse/MWPW-194867

- [x] C5: you are ready to do a demo from Test Page in PR

Test URLs:

- After: https://mwpw-194867--mas--adobecom.aem.live/

To test the script in dry run for de_DE locale only:
1. Get your IMS token:
- Go to https://main--mas--adobecom.aem.live/studio.html#expires_in=86399998&locale=de_DE&page=content&path=acom&token_type=bearer
- Open Dev tools > Network tab, filter by 'author'
- In the request Header tab find the Authorization: Bearer, the string after "Bearer" is your IMS token. Usually it starts with 'eyJh...'

2. Insert your IMS token in this command and run it in your terminal:

`curl -sk -X POST \
  "https://14257-masstudio-mirafedas.adobeioruntime.net/api/v1/web/MerchAtScaleStudio/cleanup-variations" \
  -H "Authorization: Bearer /replace this with your IMS token/" \
  -H "x-gw-ims-org-id: 9E1005A551ED61CA0A490D45@AdobeOrg" \
  -H "Content-Type: application/json" \
  -d '{"surface":"sandbox","locale":"de_DE","dryRun":true}' | python3 -m json.tool`
  
 What to verify in the response:
- statusCode: 200 — action ran cleanly
- body.dryRun: true — confirms no changes were made
- body.details — lists fragments where invalid variations were found
- body.errors — should be empty []

#### Preview

- non regression & fix on us plans page https://main--cc--adobecom.aem.page/creativecloud/plans?maslibs=MWPW-194867
- non regression on fr plans page https://main--cc--adobecom.aem.page/fr/creativecloud/plans?maslibs=MWPW-194867
- non regression on catalog page https://main--cc--adobecom.aem.page/products/catalog?maslibs=MWPW-194867

#### mas/io

- non regression on us plans page https://main--cc--adobecom.aem.live/creativecloud/plans?mas-io-url=https://14257-merchatscale-mirafedas.adobeioruntime.net/api/v1/web/MerchAtScale
- non regression on fr plans page https://main--cc--adobecom.aem.live/fr/creativecloud/plans?mas-io-url=https://14257-merchatscale-mirafedas.adobeioruntime.net/api/v1/web/MerchAtScale
- non regression on catalog page https://main--cc--adobecom.aem.live/products/catalog?mas-io-url=https://14257-merchatscale-mirafedas.adobeioruntime.net/api/v1/web/MerchAtScale


io/studio/src/cleanup-variations/index.js
```diff
@@ -0,0 +1,255 @@
...
+    const paths = [];
+    let cursor = null;
+
+    do {
+        const qs = cursor ? `path=${folderPath}&cursor=${encodeURIComponent(cursor)}` : `path=${folderPath}`;
+        const response = await withBackoffRetry(`listing ${surface}/${locale}`, () =>
+            fetchOdin(odinEndpoint, `/adobe/sites/cf/fragments?${qs}`, authToken, {
+                ignoreErrors: [400, 404, 429],
+            }),
+        );
```
npeltier: please rather use already existing postToOdinWithRetry from common

npeltier (changes requested):
please reuse retry mechanism from common. it could be you need to change common code a bit, as it's just for post requests for now. i guess we would need to integrate the retry to all fetch requests

Andrei4226 (approved):
LGTM

io/studio/src/cleanup-variations/index.js
```diff
@@ -0,0 +1,225 @@
...
+ * @returns {Promise<string[]>} array of fragment paths
+ */
+async function listFragmentPaths(odinEndpoint, surface, locale, authToken) {
+    const folderPath = `/content/dam/mas/${surface}/${locale}`;
+    const paths = [];
+    let cursor = null;
+
+    do {
+        const qs = cursor ? `path=${folderPath}&cursor=${encodeURIComponent(cursor)}` : `path=${folderPath}`;
+        const response = await fetchOdin(odinEndpoint, `/adobe/sites/cf/fragments?${qs}`, authToken, {
```
Axelcureno: **Blocker:** `?path=` is the single-fragment-by-exact-path endpoint — `fetchFragmentByPath`/`fetchIndexFragment` use it and read `items[0]`. Pointed at a folder it returns the folder node (or nothing) with no `cursor`, so this loop processes ~0 fragments per locale. Use the folder-enumeration pattern from `scripts/content/common.js` `listFolderFragments`:

```js
const query = JSON.stringify({ filter: { path: folderPath }, sort: [{ on: 'created', order: 'ASC' }] });
// GET /adobe/sites/cf/fragments/search?query=…&cursor=…
```

io/studio/src/cleanup-variations/index.js
```diff
@@ -0,0 +1,225 @@
...
+ * Lists all fragments under a locale folder using cursor-based pagination.
+ * Uses the same ?path= param as the rest of the codebase, pointed at the folder.
+ * @param {string} odinEndpoint
+ * @param {string} surface e.g. 'acom'
+ * @param {string} locale e.g. 'de_DE'
+ * @param {string} authToken
+ * @returns {Promise<string[]>} array of fragment paths
+ */
+async function listFragmentPaths(odinEndpoint, surface, locale, authToken) {
+    const folderPath = `/content/dam/mas/${surface}/${locale}`;
```
Axelcureno, line 119: `locale` (from the POST body) is concatenated raw into `folderPath` with no allow-list — unlike `surface`, which 400s on unknown values at line 174. Today exact-match `?path=` limits the blast radius, but once the endpoint becomes `/search` with a path filter (see the listing comment), a value like `../../cc` would sweep and mutate outside `/content/dam/mas/<surface>/`. Validate `targetLocale` against the surface's locale list (or `^[a-z]{2,3}_[A-Z]{2}$`) before use.

io/studio/src/cleanup-variations/index.js
```diff
@@ -0,0 +1,225 @@
...
+        const qs = cursor ? `path=${folderPath}&cursor=${encodeURIComponent(cursor)}` : `path=${folderPath}`;
+        const response = await fetchOdin(odinEndpoint, `/adobe/sites/cf/fragments?${qs}`, authToken, {
+            ignoreErrors: [400, 404, 429],
+        });
+        if (!response.ok) break;
+        const data = await response.json();
+        for (const item of data.items ?? []) {
+            if (item.path) paths.push(item.path);
+        }
+        cursor = data.cursor ?? null;
```
Axelcureno: `while (cursor)` terminates only when Odin drops the cursor. No defense if it ever echoes the same cursor (infinite loop, bounded only by the 1h timeout) and no page cap. Consider `if (cursor === previousCursor) break;` plus a max-pages guard.

io/studio/src/cleanup-variations/index.js
```diff
@@ -0,0 +1,225 @@
...
+    logger.info(
+        `${dryRun ? '[dry-run] ' : ''}Fragment ${fragmentPath}: removing ${removedPaths.length} variation(s): ${removedPaths.join(', ')}`,
+    );
+
+    if (!dryRun) {
+        // Variations field is locked by live relationships — must use PUT with full fragment
+        const updatedFields = fragment.fields.map((field) =>
+            field.name === 'variations' ? { ...field, values: validPaths } : field,
+        );
+        const result = await putToOdin(odinEndpoint, fragment.id, authToken, {
```
Axelcureno: Re: @npeltier's retry comment — there's no bespoke retry to remove here, the gap is the inverse: `postToOdinWithRetry` in `common.js` only wraps POST, so this PUT gets zero retry on transient 5xx. For a 1h bulk job that means a blip fails a fragment outright (it's caught + accumulated into `summary.errors`, so state stays consistent — but the write is lost). Worth a PUT-retry wrapper.

io/studio/app.config.yaml
```diff
@@ -114,3 +114,14 @@ application:
                             timeout: 60000
                         annotations:
                             require-adobe-auth: false
+                    cleanup-variations:
+                        function: src/cleanup-variations/index.js
+                        web: 'yes'
+                        runtime: nodejs:22
+                        inputs:
+                            LOG_LEVEL: debug
```
Axelcureno: `LOG_LEVEL: debug` shipped to the action config. No token leakage (Authorization is never logged), but it emits full fragment paths at debug. Drop to `info`.

Axelcureno (dismissed):
Requesting changes — the core logic and the `locales.json` extraction are solid (diffed all 5 surfaces, they're faithful), but the action won't actually do anything as written, and one input needs validating before it can be enabled safely.

**1. Wrong Odin endpoint — action is inert.** `listFragmentPaths` queries `/adobe/sites/cf/fragments?path=…`, which is the single-fragment-by-path endpoint (the same one `fetchFragmentByPath` and `fetchIndexFragment` use to grab `items[0]`). Folder walks everywhere else in the repo — `listFolderFragments` (`scripts/content/common.js`), `bulk-publish-discover.mjs` — hit `/adobe/sites/cf/fragments/search` with `query={filter:{path}}` and a cursor. Pointed at a folder, `?path=` returns ~nothing with no cursor, so you'll get `processed: 0` per locale. Needs the search endpoint.

**2. `locale` isn't validated like `surface` is — fix in the same pass.** `surface` 400s on unknown values; `locale` flows straight into the folder path. Harmless *today* because exact-match `?path=` can't traverse — but the moment you move to `/search` with a path filter, an unvalidated `locale` like `../../cc` becomes a recursive sweep+mutate outside `/content/dam/mas`. So please land the endpoint fix and a `locale` allow-list together — they compound.

**Smaller stuff (inline):** PUT path has no retry (the existing `postToOdinWithRetry` only wraps POST — for a 1h bulk job a transient 5xx on a PUT just fails that fragment); `LOG_LEVEL: debug` is still set; and the cursor loop could use an echo-guard / page cap.

Tests are genuinely thorough otherwise — both dryRun paths, error accumulation, 404/429/500 all covered. 👍

io/studio/src/common.js
```diff
@@ -372,6 +376,39 @@ async function putToOdin(odinEndpoint, fragmentId, authToken, { title, descripti
...
+                logger.info(`Waiting ${delay}ms before retry...`);
+                await new Promise((resolve) => setTimeout(resolve, delay));
+                continue;
+            }
+        }
+    }
+    throw new Error(`Failed to PUT fragment ${fragmentId} after ${maxRetries} attempts: ${lastError}`);
+}
+
 /**
```
npeltier, line 390: i think we should not add this, remove postToOdinWithRetry, and "just" manage non 429 retry in fetch if method is POST or PUT

npeltier (approved):
thanks! looks good to me

npeltier (changes requested):
@mirafedas this is now the QA hat that requests a change :) 
i think i found a regression, if you go on 
https://www.adobe.com/fr/express/pricing?mas-io-url=https:/14257-merchatscale-mirafedas.adobeioruntime.net/api/v1/web/MerchAtScale
you will find several requests to your workspace that 404ed
vs 
https://www.adobe.com/fr/express/pricing where it just works.

npeltier (approved):

npeltier:
@mirafedas getting the following trying to execute it with my token (that should have sufficient rights):
```⏺ Bash(curl -sk -X POST \
        "https://14257-masstudio-mirafedas.adobeioruntime.net/api/v1/web/MerchAtScaleStudio/cleanup-variations" \…)
  ⎿  {
       "error": "request is invalid: [ERR1202] User '81191F54631C19740A495E2E@7fc220a1631c0d17495e72.e' does not belong to organization '9E1005A551ED61CA0A490D45@AdobeOrg'"
     }
 
⏺ The token is valid but the user (81191F54631C19740A495E2E@...) doesn't belong to the org 9E1005A551ED61CA0A490D45@AdobeOrg. You'll need either a token minted for that org or the correct org ID for your
  account.```

PR #863, opened 2026-05-13
MWPW-178157: finalize aio cli update for deploying to stage&prod

3ch023:
Resolves https://jira.corp.adobe.com/browse/MWPW-178157

- Use aio cli 11 for deploy to stage and prod.
- Remove QA fallback for PR deploy. This will prevent QA validating 'wrong' PR using ?mas-io-url
- add linter for github workflows https://jira.corp.adobe.com/browse/MWPW-194974 

## Test URLs:

- After: https://MWPW-178157--mas--adobecom.aem.live/

### DIFF
--- a/.github/workflows/io-merge.yaml
+++ b/.github/workflows/io-merge.yaml
@@ -40,7 +40,7 @@ jobs:
         needs: deploy_stage
         uses: ./.github/workflows/io-tests.yaml
         secrets:
-            workspace: ${{ secrets.AIO_NS_Stage }}
+            workspace: ${{ secrets.AIO_NS_STAGE }}
 
     deploy_production:
         needs: test_stage
--- a/.github/workflows/io-merge.yaml
+++ b/.github/workflows/io-merge.yaml
@@ -33,8 +33,8 @@ jobs:
         needs: build_and_test
         uses: ./.github/workflows/io-deploy.yaml
         secrets:
-            ENV_CONTENT: ${{ secrets[format('AIO_WWW_ENV_{0}', github.actor)] || secrets.AIO_WWW_ENV_QA }}
-            AIO_CONTENT: ${{ secrets[format('AIO_WWW_AIO_{0}', github.actor)] || secrets.AIO_WWW_AIO_QA }}
+            ENV_CONTENT: ${{ secrets.AIO_WWW_ENV_STAGE }}
+            AIO_CONTENT: ${{ secrets.AIO_WWW_AIO_STAGE }}
 
     test_stage:
         needs: deploy_stage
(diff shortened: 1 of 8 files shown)


.github/workflows/io-pr.yaml
```diff
@@ -38,12 +38,12 @@ jobs:
         if: success() && github.event_name == 'pull_request'
         uses: ./.github/workflows/io-deploy.yaml
         secrets:
-            ENV_CONTENT: ${{ secrets[format('AIO_WWW_ENV_{0}', github.actor)] || secrets.AIO_WWW_ENV_QA }}
```
honstar, line 41: By removing the fallback to the QA workspace in case there are no corresponding personal GH secrets for the current `github.actor`, this action will fail. Do we really want that?

honstar: Checked this with @3ch023, this is intentional to avoid competing PRs targeting io/www to overwrite/conflict with each other in the QA fallback workspace.

.github/workflows/lint-workflows.yaml
```diff
@@ -0,0 +1,16 @@
+name: Lint GitHub Workflows
+on:
+    pull_request:
+        types: [opened, synchronize, reopened]
+        paths:
+            - '.github/**'
```
honstar: The scope should be narrowed to `github/workflows/**`.

.github/workflows/io-merge.yaml
```diff
@@ -33,32 +33,24 @@ jobs:
...
-            ENV_CONTENT: ${{ secrets[format('AIO_WWW_ENV_{0}', github.actor)] || secrets.AIO_WWW_ENV_QA }}
-            AIO_CONTENT: ${{ secrets[format('AIO_WWW_AIO_{0}', github.actor)] || secrets.AIO_WWW_AIO_QA }}
+            ENV_CONTENT: ${{ secrets.AIO_WWW_ENV_STAGE }}
+            AIO_CONTENT: ${{ secrets.AIO_WWW_AIO_STAGE }}
 
     test_stage:
         needs: deploy_stage
         uses: ./.github/workflows/io-tests.yaml
         secrets:
             workspace: ${{ secrets.AIO_NS_Stage }}
```
honstar: Unrelated to this PR but still worth fixing the casing: instead of `AIO_NS_Stage` use `AIO_NS_STAGE`. No functional impact if left unchanged since GH actions treat secret labels case-insensitive.

honstar (changes requested):

.github/workflows/lint-workflows.yaml
3ch023: good catch, thank you

.github/workflows/io-pr.yaml
3ch023: i think so, it just means that developer needs to setup personal workspace (or fix if setup broke)

honstar (approved):

Andrei4226 (approved):

PR #866, opened 2026-05-13
MWPW-195154 Show template-scoped overrides when templates match the top-level setting

yesil:
Resolves https://jira.corp.adobe.com/browse/MWPW-195154

## Summary

When a top-level setting has a non-empty template set, an override sharing the same `templateIds` (no locales, template-scoped) was silently dropped during index parsing in `#setRowsFromIndex`. The user could create such an override via the dialog, see a success toast, and then see nothing in the table.

The dedup branch in `studio/src/settings/settings-store.js` previously skipped any secondary no-locale fragment whose templates matched the top-level. That dedup is now narrowed to only apply when the candidate has **no** templates — i.e. a true duplicate top-level shape. Template-scoped overrides now fall through and are rendered nested under their top-level row.

The unused `normalizedStringList` / `areStringListsEqual` helpers were removed.

## 🧪 Nala E2E Tests

### To run Nala tests again:

- Remove the `run nala` label

## Test URLs:

- Before: https://main--mas--adobecom.aem.live/studio.html#page=settings&path=sandbox
- After: https://mwpw-195154--mas--adobecom.aem.live/studio.html#page=settings&path=sandbox

## Test plan

- [x] Unit test added: `studio/test/settings/settings-store.test.js` — "keeps a template-scoped override when its templates match the top-level templates"
- [x] Full settings test suite passes (`settings-store.test.js`, `mas-settings.test.html`, `mas-settings-table.test.html` — 71 tests)
- [ ] Manual: in MAS Studio settings page, create a top-level setting scoped to template X; add a template-scoped override on the same template X (no locale); reload — override row should now appear nested under the top-level.

### DIFF
--- a/studio/src/settings/mas-settings.js
+++ b/studio/src/settings/mas-settings.js
@@ -526,7 +526,7 @@ class MasSettings extends LitElement {
             label: row.label,
             name: row.name,
             description: row.description,
-            templateIds: [],
+            templateIds: [...row.templateIds],
             tags: [],
             valueType,
             value,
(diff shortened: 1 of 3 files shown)


seanchoi0 (approved):
lgtm

npeltier (approved):

Roycethan (approved):

PR #873, opened 2026-05-18
MWPW-193946: Add bulk publish revert with pre-publish snapshot

seanchoi0:
Adds a revert workflow to bulk publish projects: before publishing, a pre-publish snapshot is created (fragment versions + publication state). After publish, the project can be reverted to restore all fragments to their pre-publish state.

- **Snapshot**: creates a versioned snapshot of all fragments before publish, recording `wasPublished` state per fragment
- **Revert**: restores each fragment version via CF API; for fragments that were unpublished before publish, sets status back to Draft via Sling POST on `_jcr_content` (removing `cq:lastReplicationAction`)
- **UX**: loading overlay during revert, success toast on completion, error shown in banner on failure
- **Version page**: shows `label` and `comment` as separate lines for snapshot versions

Resolves https://jira.corp.adobe.com/browse/MWPW-193946

Test URLs:

- After: https://mwpw-193946--mas--adobecom.aem.live/studio.html?io.studio.env=seanchoi#page=bulkPublish&path=sandbox


studio/src/bulk-publish/bulk-publish-snapshot.js
```diff
@@ -0,0 +1,109 @@
...
+    const results = [];
+
+    for (const [fragmentId, entry] of Object.entries(snapshot.fragments)) {
+        try {
+            const fragment = await aem.sites.cf.fragments.getById(fragmentId);
+            const modifiedAt = fragment.modified?.at;
+            const modified = modifiedAt ? new Date(modifiedAt).getTime() > snapshotTime : false;
+            results.push({ path: entry.path, modified });
+        } catch {
+            results.push({ path: entry.path, modified: false });
```
mirafedas: If the getById method throws an error (for example, the fragment does not exist), we mark this fragment as not modified - shouldn't we surface an error status for such a fragment instead, so that authors know which fragments need to be fixed/do not exist etc? And let's add a unit test for this case

studio/src/bulk-publish/bulk-publish-store.js
```diff
@@ -9,21 +10,38 @@ function setField(project, name, value) {
...
+                ? project.value.getFieldValue('title')
+                : (project.getFieldValue?.('title') ?? project.title ?? '');
+        snapshot = await createSnapshot({ items, id: project.id, title: projectTitle }, aem, userEmail);
+    } catch (err) {
+        setField(project, 'lastError', err.message);
+        setField(project, 'status', BULK_PUBLISH_STATUS.DRAFT);
+        await repository.saveFragment(project, false);
+        return;
+    }
+    setField(project, 'snapshot', JSON.stringify(snapshot));
```
mirafedas: What will happen if we create a snapshot of a very big project, and the stringified JSON exceeds the size limit of the AEM text field? Maybe we should store the snapshot as an AEM asset, and store a snapshot ID, or limit the number of fragments that may be added to the project?

studio/src/bulk-publish/bulk-publish-store.js
```diff
@@ -9,21 +10,38 @@ function setField(project, name, value) {
...
 
+    const profile = await window.adobeIMS?.getProfile?.().catch(() => null);
+    const userEmail = profile?.email ?? '';
+
+    const aem = repository.aem;
+    let snapshot;
+    try {
+        const projectTitle =
+            typeof project.value?.getFieldValue === 'function'
+                ? project.value.getFieldValue('title')
```
mirafedas: Let's use the `getProjectField` helper here if possible

studio/src/bulk-publish/mas-bulk-publish-items.css.js
```diff
@@ -142,4 +142,21 @@ export const styles = css`
     .status-pending {
         color: var(--spectrum-gray-600, #505050);
     }
+    .modification-cell {
+        display: flex;
+        align-items: center;
+        gap: 6px;
+        font-size: 14px;
```
mirafedas: Let's use var(--spectrum-font-size-100) (and maybe for other properties it is possible to use Spectrum variables as well?)

studio/src/bulk-publish/bulk-publish-store.js
```diff
@@ -41,3 +59,41 @@ export async function startPublishing({ project, paths, locales, token, ioBaseUr
...
+        setField(project, 'status', BULK_PUBLISH_STATUS.REVERTED);
+    } catch (err) {
+        console.error('Failed to revert bulk publish project:', err);
+        setField(project, 'lastError', err.message);
+        setField(project, 'status', BULK_PUBLISH_STATUS.PUBLISHED);
+    }
+    await repository.saveFragment(project, false);
+
+    const current = Store.bulkPublishProjects.list.data.get() ?? [];
+    Store.bulkPublishProjects.list.data.set([...current]);
```
mirafedas, line 44: Should we maybe show a toast message to the user here informing the reverting went well / failed?

mirafedas (changes requested):

studio/src/bulk-publish/bulk-publish-store.js
seanchoi0: @mirafedas The `Snapshot` field isn’t a text field in AEM; it is a JSON field. I’m more concerned about the `Items` field exceeding the limit. In fact, the snapshot size tends to be about half the size of Items. In my tests, it still worked even when the limit was exceeded.

@Axelcureno Do you have any insight about this?

image

studio/src/bulk-publish/bulk-publish-snapshot.js
seanchoi0: Good point. I still think this is a very edge-case scenario since Studio doesn’t provide a delete option. So if a fragment gets deleted, it’s most likely because someone modified things directly through Odin or AEM.

Unfortunately, the current approach won’t restore deleted fragments (the Package Manager approach would). I’ll add some kind of marker, such as Not found.

studio/src/bulk-publish/bulk-publish-store.js
seanchoi0: The toast message is showing from here: https://github.com/adobecom/mas/pull/873/changes/BASE..bf7d85cbcbb99860eca7f3a4e29adf4addcab2e0#diff-67c1ee513e10121518c1f82fc2c4c7dde0d1f50cde765c89af5972337c072359R609
image

studio/src/bulk-publish/bulk-publish-snapshot.js
seanchoi0: Done
image

studio/src/aem/aem.js
```diff
@@ -522,6 +522,32 @@ class AEM {
...
+    /**
+     * Set a fragment's publication status back to DRAFT by removing the replication
+     * action property via the Sling POST Servlet on the fragment's jcr:content node.
+     * @param {string} fragmentPath - The DAM path of the fragment (e.g. /content/dam/mas/...)
+     * @returns {Promise<void>}
+     */
+    async setFragmentToDraft(fragmentPath) {
+        const csrfToken = await this.getCsrfToken();
+        const formData = new FormData();
+        formData.append('cq:lastReplicationAction@Delete', '');
```
honstar: This is actually wrong and would only be cosmetics for the frontend: the actual fragment would still be live and served on publish/CDN. Instead, it should be unpublished for real.

honstar (changes requested):

studio/src/bulk-publish/bulk-publish-store.js
honstar: I know it's tempting to just dump a serialized JSON string into a CF field but that is not the right approach imo. If you want to store a list of items that are in fact objects, you should rather consider creating an additional CF model that holds the data you need for each item. We are not storing plain JSON content anywhere else and we should not start doing that going forward.

studio/src/bulk-publish/bulk-publish-snapshot.js
```diff
@@ -0,0 +1,109 @@
...
+            const rawFragment = results[0];
+            if (!rawFragment) {
+                throw new Error(`Fragment not found at path: ${item.path}`);
+            }
+            fragmentId = rawFragment.id;
+        }
+
+        const fragment = await aem.sites.cf.fragments.getById(fragmentId);
+        const wasPublished = fragment.status === STATUS_PUBLISHED || fragment.status === STATUS_MODIFIED;
+        const versionId = await aem.sites.cf.fragments.createVersion(fragmentId, {
```
honstar: This is in fact a sequential versioning operation that duration increases with each fragment added to a snapshot. This loop scales poorly with an increasing item count and there is no UX feedback per item or overall progress.
To overcome the scalability concern, I'd recommend looking into `Promise.all` with a concurrency cap to not overload the odin backend and run into 429 issues.

studio/src/aem/aem.js
```diff
@@ -522,6 +522,32 @@ class AEM {
...
     }
 
+    /**
+     * Set a fragment's publication status back to DRAFT by removing the replication
+     * action property via the Sling POST Servlet on the fragment's jcr:content node.
+     * @param {string} fragmentPath - The DAM path of the fragment (e.g. /content/dam/mas/...)
+     * @returns {Promise<void>}
+     */
+    async setFragmentToDraft(fragmentPath) {
+        const csrfToken = await this.getCsrfToken();
```
honstar: Minor: please add a guard against an empty `fragmentPath`.

studio/src/aem/aem.js
```diff
@@ -522,6 +522,32 @@ class AEM {
...
+        const formData = new FormData();
+        formData.append('cq:lastReplicationAction@Delete', '');
+        formData.append('_charset_', 'UTF-8');
+        const response = await fetch(`${this.baseUrl}${fragmentPath}/_jcr_content`, {
+            method: 'POST',
+            headers: {
+                ...this.headers,
+                'csrf-token': csrfToken,
```
honstar: nit: `forceDeleteFragment` uses `CSRF-Token`, the casing should be consistent for grep-ability.

studio/test/bulk-publish/bulk-publish-utils.test.js
```diff
@@ -0,0 +1,59 @@
...
+        const project = { value: { status: 'Published' } };
+        expect(getProjectField(project, 'status')).to.equal('Published');
+    });
+
+    it('returns fallback when field is undefined and no property exists', () => {
+        const project = { getFieldValue: () => undefined };
+        expect(getProjectField(project, 'missing', 'default')).to.equal('default');
+    });
+
+    it('returns undefined (not fallback) when field value is null', () => {
```
honstar: The name of this test is misleading: the test actually asserts that `default` is returned when field value is `null` which is the opposite of what the test name suggests.

honstar (dismissed):

studio/src/bulk-publish/bulk-publish-store.js
seanchoi0: @honstar  Just to clarify, are you suggesting that we avoid storing JSON even in fields that are specifically defined as JSON types?

studio/src/bulk-publish/bulk-publish-snapshot.js
seanchoi0: Good catch. I've updated for revertSnapshot and checkModifications as well.

studio/src/aem/aem.js
seanchoi0: @honstar ,
Thanks for the feedback. I actually tried using the replication API to unpublish first, but I ran into an issue where once a fragment is set to Unpublished, it can only transition back to Published, all other statuses seem to be blocked. That’s what led me to this workaround.

I have a couple of genuine questions as I work toward the right fix:

- How should we verify CDN behavior here?
- What is a realistic scenario where a fragment being “cosmetically draft” but still live on CDN would cause a real problem?

I’m happy to rework this once I understand the right path forward.

studio/src/bulk-publish/bulk-publish-store.js
Axelcureno: I agree with @honstar but adding new models needs to be discussed with the team cc @npeltier @yesil @3ch023

studio/src/aem/aem.js
npeltier: "activation" on odin triggers:
- distribution to odin publish,
- distribution to odin freyja that will consume odin publish content

every 15 min, content is being refreshed from freyja layer.

So yes, `setFragmentToDraft` if status is published, should trigger an unpublish action first to update all those layers above

studio/src/bulk-publish/bulk-publish-snapshot.js
```diff
@@ -0,0 +1,132 @@
...
+    let next = 0;
+    async function worker() {
+        while (next < items.length) {
+            const i = next++;
+            results[i] = await fn(items[i]);
+        }
+    }
+    await Promise.all(Array.from({ length: Math.min(limit, items.length) }, worker));
+    return results;
+}
```
npeltier: please re-use processBatchWithConcurrency from common

npeltier (changes requested):

seanchoi0:
> So yes, setFragmentToDraft if status is published, should trigger an unpublish action first to update all those layers above

@npeltier , As I mentioned during the meeting, once it is marked as unpublished, it seems unable to transition to any status other than published.

https://github.com/user-attachments/assets/a6acb11a-5a72-41b9-806d-29ae96b42a95

studio/src/bulk-publish/bulk-publish-snapshot.js
seanchoi0: `processBatchWithConcurrency` is in io/studio/src/common.js (CJS/Node.js) so it can't be imported from the browser side directly, but the concern about duplication is valid. Would moving `runConcurrent` to bulk-publish-utils.js work as an alternative?

npeltier: i think it can if people agree to do what we are doing already in io/www

npeltier (changes requested):
as discussed offline please move all the snapshot & revert code backend

npeltier:
> @npeltier , As I mentioned during the meeting, once it is marked as unpublished, it seems unable to transition to any status other than published.

as discussed please keep the 3 statuses:
- unpublished
- published
- modified / draft

studio/src/bulk-publish/bulk-publish-client.js
```diff
@@ -44,3 +47,45 @@ export async function publishBulk({ ioBaseUrl, paths, locales = [], token }) {
 
     return body;
```
npeltier: you should use "callAction" there too. 
while we are at it, why not making this even more generic?

studio/src/bulk-publish/bulk-publish-store.js
```diff
@@ -9,21 +10,43 @@ function setField(project, name, value) {
     }
 }
 
+function setSnapshots(project, entries) {
+    if (typeof project.updateField === 'function') {
+        project.updateField('snapshots', entries);
+    } else {
+        project.setFieldValue('snapshots', entries);
+    }
+}
```
npeltier: why do we do content update "again" from client, why isn't it snapshotaction updating it?

studio/src/bulk-publish/bulk-publish-store.js
```diff
@@ -9,21 +10,43 @@ function setField(project, name, value) {
...
+    let snapshotEntries;
+    try {
+        const { createSnapshotAction } = await import('./bulk-publish-client.js');
+        const projectTitle = getProjectField(project, 'title', '');
+        snapshotEntries = await createSnapshotAction({ ioBaseUrl, paths, projectId: project.id, projectTitle, token });
+    } catch (err) {
+        setField(project, 'lastError', err.message);
+        setField(project, 'status', BULK_PUBLISH_STATUS.DRAFT);
+        await repository.saveFragment(project, false);
+        return;
```
npeltier: same here, this should be snapshot action managing project status

npeltier (changes requested):
asked claude to shape out my issue with that mixture of client & server writting: 
 
Move all project-fragment writes into the io actions. Client passes only {projectId, token, ioBaseUrl, locales} and receives the final project state back.

  /bulk-publish action (extend existing one)

  1. Read project fragment by id.
  2. Patch status=PUBLISHING, lastError='' → save.
  3. Build snapshot (createSnapshot already lives in snapshot.js).
  4. Call publish workflow.
  5. Patch snapshots, status=PUBLISHED|DRAFT, publishedAt, publishedBy (from IMS profile lookup), lastResult, lastError → save. One write, both branches.
  6. Return {status, lastError, ...}.

  /bulk-revert action (extend existing one)

  1. Read project fragment, parse snapshots.
  2. Patch status=REVERTING, lastError='' → save.
  3. Run revert.
  4. Patch snapshots (with skipped removed), status=REVERTED|PUBLISHED, lastError → save.
  5. Return final state.

  Client side
  
  async function startPublishing({ projectId, token, ioBaseUrl, locales, repository }) {
      const result = await publishAction({ ioBaseUrl, projectId, locales, token });
      await repository.refreshProject(projectId);
      return result;
  }

  bulk-publish-store.js shrinks from ~140 lines to ~10. No setField, no setSnapshots, no catch-block state management.

npeltier (approved):
LGTM

mirafedas (approved):

studio/src/bulk-publish/mas-bulk-publish.js
```diff
@@ -93,6 +93,30 @@ class MasBulkPublish extends LitElement {
...
+        const title = getProjectField(projectStore.get(), 'title', 'Untitled project');
+        this.revertPending = { projectStore, title };
+    }
+
+    handleRevertCancel() {
+        this.revertPending = null;
+    }
+
+    async handleRevertConfirmed() {
+        const { projectStore } = this.revertPending;
```
yesil, line 106: Functionally, this path calls the backend revert successfully, but then `startReverting` receives `projectStore.get()` instead of the store wrapper. `patchProjectStore()` calls `.get()` / `.refreshFrom()`, so the overview revert flow will throw after the IO action returns; users can see a failed or stale UI even though content may already have been reverted. Suggest passing `projectStore` to `startReverting` here, or making `startReverting`/`patchProjectStore` explicitly support raw fragment objects and update the backing list store.

io/studio/src/bulk-publish/index.js
```diff
@@ -133,4 +147,86 @@ function buildSummary(details) {
...
+    const chunks = groupAndChunk(resolved, MAX_CHUNK_SIZE);
+    const details = [];
+    for (const chunk of chunks) {
+        const results = await publishOneChunk(chunk, odinEndpoint, authToken);
+        details.push(...results);
+    }
+
+    const summary = buildSummary(details);
+    const failures = details.filter((d) => d.status === STATUS.FAILED);
+    const finalStatus = failures.length === 0 ? PROJECT_STATUS.PUBLISHED : PROJECT_STATUS.DRAFT;
```
yesil, line 248: Functionally, if this final project-fragment write fails, the action still returns success after fragments were published. The project can remain stuck in `Publishing` and miss `snapshots`, `publishedAt`, `publishedBy`, `lastResult`, and `lastError`, while the client shows local success until reload. Suggest making the final metadata write part of the required success path for this action: return an error/failed project state when it cannot be persisted, or retry and only return success after the project fragment reflects the final state.

io/studio/src/bulk-publish/revert-action.js
```diff
@@ -0,0 +1,114 @@
...
+            } catch {
+                return true;
+            }
+        });
+    }
+
+    const finalStatus = failures.length === 0 ? PROJECT_STATUS.REVERTED : PROJECT_STATUS.PUBLISHED;
+    const finalError = failures.length === 0 ? '' : `REVERT:\n${failures.map((f) => `${f.path}: ${f.error}`).join('\n')}`;
+
+    await updateProjectFragment(odinEndpoint, projectId, authToken, {
```
yesil: Functionally, if this final project update fails, the fragments may already be restored/unpublished but the project fragment can stay in `Reverting`, keep stale snapshot entries, and lose the final `Reverted`/failure state. On reload, users cannot tell whether revert completed. Suggest not swallowing this write failure: retry it or return a failure response so the UI does not report a successful revert unless the project metadata was persisted.

afmicka:
@seanchoi0 there are some comments that probably need addressing first. 
Also:
1. i have created a project that was created on your branch, it contains 3 fragments, 1 default and 2 its variations.
https://mwpw-193946--mas--adobecom.aem.live/studio.html?io.studio.env=seanchoi#bulkPublishProjectId=c908e309-faab-4fa5-adee-b396d4f2f8ef&page=bulkPublishEditor&path=sandbox
 
Screenshot 2026-05-29 at 13 48 06

When i afterwards open the same project on main, it is empty. 
https://main--mas--adobecom.aem.live/studio.html?io.studio.env=seanchoi#bulkPublishProjectId=c908e309-faab-4fa5-adee-b396d4f2f8ef&page=bulkPublishEditor&path=sandbox
Screenshot 2026-05-29 at 13 49 23

2. when i published the project nothing happened! There is no toast, no green header that the project has been published and when, it stays in the draft status and can be edited further. No error either 

3. You introduced that clicking on the project in the list opens the project editor. However, with the current changes it happens on the first click. All our other features are such that it should open the editor on double-click.

seanchoi0:
@afmicka ,
-  I’ll address his comments.
1. This is not a bug; it’s due to a change in how the project saves items after the content model update. The main branch uses the old fields, while my branch uses the new ones.
2. I had forgotten to deploy my AIO workspace after the last cleanup. I just did, and it seems to be working fine now.
3. I made it a one-click action because Tomislav Reil suggested that a single click feels more natural than a double click to open the project. If that does not fit the UX in overall, then we should update the other features as well.

io/studio/src/bulk-publish/index.js
seanchoi0: Good catches. The current .catch() behavior is intentional — propagating a failure response after content was already published/reverted felt more misleading than a stale status field. But you're right that silently leaving the project in Publishing/Reverting is not acceptable either.

The reason we can't simply retry inside the action or have the user retry safely is snapshot integrity. The snapshot is meant to capture the pre-publish state for revert. If the final write fails and the user retries, createSnapshot runs again — this time capturing the already-published state, which makes the snapshot useless for revert.

**How I fix this problem:**

  1. Write the snapshot to the project fragment before publishing (publishComplete: false), not as part of the final metadata write
  2. After publish succeeds, final write sets publishComplete: true along with status, publishedAt, etc.
  3. If the final write fails → fail-hard with a clear message: "Content was published but project state could not be saved. Please retry — if the issue persists, contact support."
  4. On retry, the existing snapshot (publishComplete: false) is detected and reused — so snapshot integrity is preserved and the retry is safe

If the initial snapshot write (step 1) fails, we abort before publishing — no harm done.

This approach handles the core concern without introducing retry logic inside the action, and keeps the failure message honest about what actually happened.

io/studio/src/bulk-publish/revert-action.js
seanchoi0: Same fix will be applied as above.

yesil (approved):

npeltier:
> I made it a one-click action because Tomislav Reil suggested that a single click feels more natural than a double click to open the project. If that does not fit the UX in overall, then we should update the other features as well.

@seanchoi0 let's keep the behaviour aligned (double click), and make the change to single click a separate ticket we groom & discuss with @reil-adobe the normal way. not a blocker for me :-)

afmicka:
@seanchoi0 
1. On your branch with io.studio.env i am not seeing "Settings" anymore. It is missing from the Advanced options (i think it was there before, not sure). I see it only without on your branch only without IO changes. I do have access to it on main branch. Could you check why it is not there and if that is expected? 
Screenshot 2026-06-01 at 11 44 44

2. When i clicked to publish the project from previous messages it got stuck in publishing with the corrupted state
https://mwpw-193946--mas--adobecom.aem.live/studio.html?io.studio.env=seanchoi#bulkPublishProjectId=c908e309-faab-4fa5-adee-b396d4f2f8ef&page=bulkPublishEditor&path=sandbox

Screenshot 2026-06-01 at 11 40 11

3. When i clicked to revert the published (new) project (see [project](https://mwpw-193946--mas--adobecom.aem.live/studio.html?io.studio.env=seanchoi#bulkPublishProjectId=27e559ee-d745-4c55-8dc1-a202360f5a66&page=bulkPublishEditor&path=sandbox)), the status of the fragment remained published instead of returning to Modified like it was before publishing. It also still references on the live page the newly published content instead of the last published content (content that was published before the revert). There are no visible indicators that the revert indeed happened event though the toast message was green.
Screenshot 2026-06-01 at 12 19 33

live page: https://milo.adobe.com/drafts/mili/test-revert-publish
history (version before the latest publish): https://mwpw-193946--mas--adobecom.aem.live/studio.html?io.studio.env=seanchoi#fragmentId=2d343d49-a990-40df-8350-6af83c94c1d6&page=version&path=sandbox
Screenshot 2026-06-01 at 12 21 02


4. Regarding single/double click, as mentioned above, every suggestion should go through team's flow and be recorded in the requirements and/or jiras. Not a blocker but we should really try to keep uniformity through out the app and not changing the behavior only in one module. Given there are still issues in this branch to fix, could you please keep the action aligned with the rest of the app and make it a double-click? We will create a separate jira to make the single-click action throughout the app. cc: @3ch023

afmicka:
@seanchoi0 i have discussed with @npeltier offline regarding this functionality and have come to conclusion that it does what it was intended to just a name is a bit misleading.

We will merge this PR as is but will keep the jira opened so that you can add a separate PR to bring the double-click for opening the project to align with the rest of the app. The change to the single-click will be a separate jira that will be discussed with Tomislav and the rest of the team.

PR #874, opened 2026-05-18
MWPW-194472: Add locales & fragments for promotions

Andrei4226:
Resolves https://jira.corp.adobe.com/browse/MWPW-194472

Extends MAS Studio Promotions so authors can configure geos (locales) and included fragments/collections from the promotion editor. Introduces a shared items-selection pattern (same building blocks as translation item pickers) so card/collection lists, search/filters and selected-items UI stay consistent.

Add search with fragment ID in selecting items window for:
- translations
- bulk publish
- promotions

Test:
https://mwpw-194472--mas--adobecom.aem.live/studio.html#page=promotions&path=sandbox

## 🧪 Nala E2E Tests

### To run Nala tests again:

- Remove the `run nala` label

## Test URLs:

- After: https://mwpw-194472--mas--adobecom.aem.live/


studio/src/mas-promotions-editor.js
```diff
@@ -82,13 +137,108 @@ class MasPromotionsEditor extends LitElement {
...
+            }
+        }
+        if (!allPaths.length) {
+            Store.promotions.selectedCards.set([]);
+            Store.promotions.selectedCollections.set([]);
+            return;
+        }
+        const cards = [];
+        const cols = [];
+        for (const path of allPaths) {
```
mirafedas: Maybe it's better to use Promise.all or Promie.allSettled here to speed it up?

studio/src/router.js
```diff
@@ -75,6 +75,25 @@ export class Router extends EventTarget {
         return false;
     }
 
+    promotionsEditorHasUnsavedChanges() {
+        const inEdit = Store.promotions.inEdit?.get()?.get();
+        if (!inEdit) return false;
+        if (inEdit.hasChanges) return true;
+
+        const fragPaths = inEdit.getFieldValues('fragments');
```
mirafedas: This part looks very similar to what #itemsSelectionDirty does, maybe we could extract this to some util function

studio/src/mas-promotions-editor.js
```diff
@@ -82,13 +137,108 @@ class MasPromotionsEditor extends LitElement {
...
+        Store.promotions.displayCollections.set([]);
+        Store.promotions.selectedCollections.set([]);
+        Store.promotions.allPlaceholders.set([]);
+        Store.promotions.placeholdersByPaths.set(new Map());
+        Store.promotions.displayPlaceholders.set([]);
+        Store.promotions.selectedPlaceholders.set([]);
+        Store.promotions.showSelected.set(false);
+    }
+
+    #syncItemStoresFromFragment() {
```
mirafedas: Maybe we could rename it to `resetPlaceholders` or smth like that, otherwise the name is a bit misleading. Or maybe we could even get rid of this function if it's only one line, and use `Store.promotions.selectedPlaceholders.set([]);` directly, no need to wrap it into a function

studio/src/mas-promotions-editor.js
```diff
@@ -272,11 +483,116 @@ class MasPromotionsEditor extends LitElement {
...
+        this.#itemsConfirmed = false;
+        this.#cardsSnapshot = Store.promotions.selectedCards.value;
+        this.#collectionsSnapshot = Store.promotions.selectedCollections.value;
+
+        const selector = this.renderRoot.querySelector('mas-promotions-items-selector');
+        if (selector) {
+            selector.searchQuery = '';
+            selector.selectedTab = TABLE_TYPE.CARDS;
+            const searchAndFilters = selector.shadowRoot?.querySelector('mas-search-and-filters');
+            if (searchAndFilters) {
```
mirafedas: Can we expose a method `resetFilters` in the mas-search-and-filters component, and then call it here instead?

mirafedas (changes requested):
Could you please add unit tests for:
1) mas-promotions-editor.js — the #hydratePromotionItemSelectionFromFragment, #itemsSelectionDirty, #serializeSurfacesForAem, and validation changes;
2) mas-promotions-items-selector.js
3) mas-promotions-items-table.js

Andrei4226:
> Member

Added tests under `studio/test/promotions/`.
Editor-related behavior is covered in `promotion-editor-utils.test.js` (extracted helpers) plus `mas-promotions-editor.test.js` for `selectedItemsCount`.

studio/src/mas-promotions-editor.js
Andrei4226: Done.
Now, it's use `Promise.allSettled` in `classifyPromotionPathsForSelection` so fetches run in parallel.

studio/src/router.js
Andrei4226: Done.
Extracted shared logic into `isPromotionItemSelectionDirty` in `promotion-editor-utils.js`. It is now used by both: `router.js` (`promotionsEditorHasUnsavedChanges`) and `mas-promotions-editor`, eliminating the need to duplicate the comparison logic.

studio/src/mas-promotions-editor.js
Andrei4226: Removed that helper and call `Store.promotions.selectedPlaceholders.set([])`.

studio/src/mas-promotions-editor.js
Andrei4226: Done, `mas-search-and-filters` now exposes `resetFilters(),` and the promotions editor calls it on each `mas-search-and-filters` instance instead of resetting filter state.

mirafedas (approved):

npeltier:
hey @Andrei4226 promotions are cross surfaces, so having an enabled surface switcher does not make sense here 
image

npeltier:
i think with your PR we could start showing promotions button to those who can, for now i think it's ADMINS 
image

and have it hidden for the others

afmicka:
> hey @Andrei4226 promotions are cross surfaces, so having an enabled surface switcher does not make sense here image

@Andrei4226 @npeltier selecting the fragments are following that picker. At the moment it opens only the files from the selected surface in the top nav. How are we going to select fragments across surfaces in the projects? Should it load them all and then we have a surface filter in the picker or what? 

Screenshot 2026-05-20 at 09 56 40

npeltier:
that's a great question @afmicka, can we add surface picker to the selection popup, showing only selected surfaces in the project? make it optional so it does not re-appear in translation projects

afmicka:
> that's a great question @afmicka, can we add surface picker to the selection popup, showing only selected surfaces in the project? make it optional so it does not re-appear in translation projects

@npeltier @Andrei4226 i would try with that solution because at the moment selecting the fragments is inconvenient and somewhat broken if you need to change the surface.

Andrei4226:
> that's a great question @afmicka, can we add surface picker to the selection popup, showing only selected surfaces in the project? make it optional so it does not re-appear in translation projects

Yes, that works also for me (without any impact on the translation component)

studio/src/mas-promotions-editor.js
```diff
@@ -4,23 +4,34 @@ import { MasRepository } from './mas-repository.js';
...
+import { renderFragmentStatusCell } from './common/utils/render-utils.js';
 
 const typeMap = {
     title: { type: 'text' },
     promoCode: { type: 'text' },
     startDate: { type: 'date-time' },
     endDate: { type: 'date-time' },
     tags: { type: 'tag', multiple: true },
-    surfaces: { type: 'long-text' },
+    surfaces: { type: 'text', multiple: false },
```
honstar: This will cause a regression: with this change, selected surfaces will appear as a comma-separated list once saved to odin. Before, individual surfaces were stored as separate entry in odin, now it would be an array which - when serialized - will appear as comma-separated list which breaks the UI.

See my test project: https://mwpw-194472--mas--adobecom.aem.live/studio.html#page=promotions-editor&path=sandbox&promotionId=af797b03-a05b-4a13-9f77-b2f6b3eeaf29

studio/test/promotions/mas-promotions-items-table.test.js
```diff
@@ -0,0 +1,356 @@
...
+                        setTimeout(() => resolve(null), 500);
+                    }),
+            },
+        }));
+        document.body.appendChild(el);
+        await el.updateComplete;
+        await new Promise((r) => setTimeout(r, 5));
+        expect(el.viewOnlyLoading).to.be.true;
+        el.remove();
+        expect(el.viewOnlyLoading).to.be.true;
```
honstar: This test doesn't really cover an abort and is wrong imo: `viewOnlyLoading` should be `false` at this point but there is no `await` to give async cleanup a chance to even call `disconnectedCallback`. It would be `true` regardless.
Instead, you should test an actual code path and verify that `processAbortController` has actually been called.

Also, the following 2 lines are meaningless: you set a variable to `true` yourself and assert it is `true` :)

studio/src/promotions/mas-promotions-items-table.js
```diff
@@ -0,0 +1,518 @@
...
+    get repository() {
+        return document.querySelector('mas-repository');
+    }
+
+    get typeUppercased() {
+        return this.type.charAt(0).toUpperCase() + this.type.slice(1);
+    }
+
+    get selectedPaths() {
+        return getItemsSelectionStore()[`selected${this.typeUppercased}`].value ?? [];
```
yesil: **Nitpick — dead fallback.** The store is initialised to `[]`, so `?? []` never fires. Either drop it or add a comment explaining the defensive intent.

studio/src/mas-promotions-editor.js
```diff
@@ -272,11 +447,110 @@ class MasPromotionsEditor extends LitElement {
...
+        this.promotionItemsAddButtonLabel = 'Add selected fragments';
+        this.#itemsConfirmed = false;
+        this.#cardsSnapshot = Store.promotions.selectedCards.value;
+        this.#collectionsSnapshot = Store.promotions.selectedCollections.value;
+
+        const selector = this.renderRoot.querySelector('mas-promotions-items-selector');
+        if (selector) {
+            selector.searchQuery = '';
+            selector.selectedTab = TABLE_TYPE.CARDS;
+            selector.shadowRoot?.querySelectorAll('mas-search-and-filters').forEach((el) => el.resetFilters());
```
yesil: **Suggestion — crossing the shadow boundary.** The editor reaches into the selector's `shadowRoot` to find `<mas-search-and-filters>` and call `resetFilters()`. The editor now knows the selector's internal DOM, so any restructure inside the selector silently breaks reset behaviour.

Expose a method on the selector and call that:

```js
// in mas-promotions-items-selector.js
resetFilters() {
  this.renderRoot.querySelectorAll('mas-search-and-filters').forEach((el) => el.resetFilters());
}

// here
selector.resetFilters();
```

studio/src/mas-promotions-editor.js
```diff
@@ -272,11 +447,110 @@ class MasPromotionsEditor extends LitElement {
...
+        this.#itemsConfirmed = true;
+        const closeEvent = new Event('close', { bubbles: true, composed: true });
+        target.dispatchEvent(closeEvent);
+    };
+
+    #cancelItemSelection = ({ target }) => {
+        Store.promotions.selectedCards.set(this.#cardsSnapshot);
+        Store.promotions.selectedCollections.set(this.#collectionsSnapshot);
+        this.showSelectedEmptyState = this.selectedItemsCount === 0;
+        this.#itemsConfirmed = true;
```
yesil: **Nitpick — misleading name.** `#cancelItemSelection` sets `#itemsConfirmed = true`. The flag is really "dialog is closed, don't restore snapshot in the close handler" — rename to `#itemsDialogClosed` (or split into two flags) so future readers don't think cancel = confirm.

studio/test/promotions/mas-promotions-editor.test.js
```diff
@@ -0,0 +1,456 @@
...
+            expect(repo.createFragment.called).to.be.false;
+            expect(el.isCreated).to.be.false;
+        });
+
+        it('creates promotion when all required fields are valid', async () => {
+            const { el, repo } = await mountEditorWithRepo();
+            await fillValidFields(el);
+            const buttons = el.renderRoot.querySelectorAll('.promotions-form-buttons sp-button');
+            buttons[1].dispatchEvent(new MouseEvent('click', { bubbles: true, composed: true }));
+            await new Promise((r) => setTimeout(r, 50));
```
yesil: **Suggestion — flake risk.** Several tests in this file `await new Promise(r => setTimeout(r, 20|50))` to wait for async work. On a loaded CI runner these timeouts will occasionally lose the race.

After button click, if you expect a re-render, you can await for it as follows:

```js
await el.updateComplete;
```

studio/test/mas-repository.test.js
```diff
@@ -663,6 +663,106 @@ describe('MasRepository dictionary helpers', () => {
...
+                status: 'DRAFT',
+                created: { by: 'u', fullName: 'U', at: '2024-01-01T00:00:00.000Z' },
+                modified: { by: 'u', fullName: 'U', at: '2024-01-02T00:00:00.000Z' },
+                fields: [
+                    { name: 'title', type: 'text', values: ['T'] },
+                    { name: 'promoCode', type: 'text', values: ['X'] },
+                    { name: 'startDate', type: 'date-time', values: ['2024-01-01T00:00:00.000Z'] },
+                    { name: 'endDate', type: 'date-time', values: ['2024-12-31T23:59:59.999Z'] },
+                    { name: 'tags', type: 'tag', values: [] },
+                    { name: 'surfaces', type: 'long-text', values: [] },
```
yesil: **Suggestion — fixture uses the old schema.** The PR migrates `surfaces` from `long-text` to `text`; this fixture still seeds `type: 'long-text'`. The test locks in dead behaviour — update to match the new model.

npeltier:
@Andrei4226 another not common thing in promotions project: you should not be able to select variations for now in projects, just plain fragment 
image

studio/src/promotions/promotion-editor-utils.js
```diff
@@ -0,0 +1,70 @@
...
+    const cards = [];
+    const cols = [];
+    results.forEach((result, i) => {
+        const path = allPaths[i];
+        if (result.status === 'fulfilled') {
+            const modelPath = result.value?.model?.path;
+            if (modelPath === collectionModelPath) cols.push(path);
+            else cards.push(path);
+        } else {
+            cards.push(path);
```
honstar: Any failed attempt to fetch a fragment would still land in the `cards` array that is surfaced to the user which will see a card they cannot preview nor edit. Not sure if this is deliberate or what the purpose is, but the funny thing is that there is an actual unit-test that explicitly checks for `bad` paths: https://github.com/adobecom/mas/pull/874/changes#diff-1f46a2aed4b8eca4ea2f731677d8af4e49e5c22d5813f1485972658b5e5e71faR74

npeltier:
> that's a great question @afmicka, can we add surface picker to the selection popup, showing only selected surfaces in the project? make it optional so it does not re-appear in translation projects

@Andrei4226 cc @afmicka 
there is also a not so rare case too where we may want to select non en_US fragments (e.g. Korea only promotion) in which case i guess we may want geo selection in the popup as well

honstar (dismissed):

studio/src/promotions/promotion-editor-utils.js
yesil: isn't `cols.push ` needed here as well?

studio/src/mas-promotions-editor.js
```diff
@@ -159,6 +300,26 @@ class MasPromotionsEditor extends LitElement {
         this.fragmentStore.updateField(fieldName, [utcDate]);
     }
 
+    #patchPromotionSurfacesFieldForAem() {
```
yesil, line 415: I wonder why `#patchPromotionSurfacesFieldForAem` is needed at the first place?

npeltier (changes requested):
thanks a lot for this! 

in addition to my previous comments, some requested changes:
- please add possibility to publish a project! 
- please adapt status as follow: if not published and future or now: "Draft", if published and in future "Scheduled", if now an published "Active", if past, "Expired".
- BONUS: If you have "Expired" and it's published, please unpublish,
- please add possibility to unpublish a project! 
- i know it's in figma, but i would add "promotion" to "Edit project" or "Create new project" so it's a bit clearer what we are doing

yesil (approved):
Really nice work!

nit: checkboxes not aligned.
Image

nit: could we support pasting a fragment link in search? 
Image

npeltier:
@Andrei4226 
as discussed offline:
- please make tags field only on promotions namespace, and mandatory,
- please add publication action

studio/src/mas-promotions-editor.js
Andrei4226: You've right.
Fixed, saves for surfaces don't serialize as a single comma-separated value.

studio/test/promotions/mas-promotions-items-table.test.js
Andrei4226: Fixed.
The disconnect test now spies `abort()`, asserts `viewOnlyLoading` goes false and waits for the slow fetch to finish.
Removed set and that assert with true :)

studio/src/promotions/promotion-editor-utils.js
Andrei4226: Thanks!
`classifyPromotionPathsForSelection` now uses `Promise.allSettled` and routes failed fetches + null/missing model into unreachable, so they don’t get pushed into cards. 
Only successful resolves with a `model.path` become cards/cols. 
unreachable is also wired into `itemHydrateUnreachablePaths` for the selection dirty check, matching the unit tests around bad paths.

Andrei4226:
> thanks a lot for this!
> 
> in addition to my previous comments, some requested changes:
> 
> * please add possibility to publish a project!
> * please adapt status as follow: if not published and future or now: "Draft", if published and in future "Scheduled", if now an published "Active", if past, "Expired".
> * BONUS: If you have "Expired" and it's published, please unpublish,
> * please add possibility to unpublish a project!
> * i know it's in figma, but i would add "promotion" to "Edit project" or "Create new project" so it's a bit clearer what we are doing

Update,  all set:

- Publish / Unpublish - list + editor
- Status: Draft / Scheduled / Active / Expired (promotionStatus + list UI)
- Bonus:  auto-unpublish when Expired and still published (editor load + promotions list load)
- Specific sentences: “Edit / Create promotion project” in nav + editor

studio/src/promotions/mas-promotions-items-table.js
Andrei4226: Thanks for this catch!
Now it's fixed.

studio/src/mas-promotions-editor.js
Andrei4226: Good catch, agreed on the encapsulation issue. 
I moved this behind `MasPromotionsItemsSelector.resetFilters()` and the editor now calls `selector.resetFilters()` instead of reaching into the shadow tree.

studio/src/mas-promotions-editor.js
Andrei4226: Fixed.

studio/test/promotions/mas-promotions-editor.test.js
Andrei4226: Fixed.

studio/test/mas-repository.test.js
Andrei4226: Updated it also for tests.
Thanks!

Andrei4226:
> @Andrei4226 as discussed offline:
> 
> * please make tags field only on promotions namespace, and mandatory,
> * please add publication action

- Added a promotion tag, which is required to create a promotion project.
- Added the ability to publish and unpublish promotion projects.

studio/src/promotions/promotion-editor-utils.js
Andrei4226: Not really
Now failed fetches and fragments without a model go to `unreachable`.
So I fixed it.

studio/src/aem/promotion.js
```diff
@@ -25,6 +27,13 @@ export class Promotion extends Fragment {
         return this.created?.fullName || 'Unknown';
     }
 
+    get isPromotionPublished() {
+        const tagsPublished = this.getFieldValues('tags').some((t) => normalizeTagId(t) === TAG_STATUS_PUBLISHED);
+        const statusUpper = typeof this.status === 'string' ? this.status.toUpperCase() : '';
+        const statusPublished = statusUpper === STATUS_PUBLISHED || statusUpper === FRAGMENT_STATUS.MODIFIED;
+        return tagsPublished || statusPublished;
+    }
```
npeltier, line 31: The status field on Fragment is set by AEM itself (PUBLISHED / DRAFT / MODIFIED per FRAGMENT_STATUS) and is the source of truth. The mas:status/published tag is application-level metadata — it
  can drift from real AEM publication state (stale after publish/unpublish, copy/paste, manual edits, etc.).

  So:

  get isPromotionPublished() {
      const tagsPublished = this.getFieldValues('tags').some((t) => normalizeTagId(t) === TAG_STATUS_PUBLISHED);
      const statusUpper = typeof this.status === 'string' ? this.status.toUpperCase() : '';
      const statusPublished = statusUpper === STATUS_PUBLISHED || statusUpper === FRAGMENT_STATUS.MODIFIED;
      return tagsPublished || statusPublished;
  }

  should be:
  
  get isPromotionPublished() {
      return this.status === FRAGMENT_STATUS.PUBLISHED || this.status === FRAGMENT_STATUS.MODIFIED;
  }

  Two side notes:
  - STATUS_PUBLISHED and FRAGMENT_STATUS.PUBLISHED are duplicates (constants.js:113 and :261). Only FRAGMENT_STATUS is needed; the new STATUS_PUBLISHED import in promotion.js should go away with the fix.
  - Two PR tests will need updating: "returns true with published tag" (test/aem/promotion.test.js:269–273) — that case should be removed (or flipped to assert false). The setPublishedTag(true) calls in
  promotionStatus tests should be replaced with mockFragmentData.status = 'PUBLISHED'.

studio/src/mas-repository.js
```diff
@@ -1068,6 +1118,16 @@ export class MasRepository extends LitElement {
...
+            const signal = this.#abortControllers.promotions.signal;
+            for (const store of promotions) {
+                if (signal.aborted) break;
+                const p = store.get();
+                if (p?.promotionStatus !== 'expired' || !p.isPromotionPublished) continue;
+                const ok = await this.unpublishFragment(p, false);
+                if (!ok || signal.aborted) continue;
+                const fresh = await this.aem.sites.cf.fragments.getById(p.id);
+                if (fresh) store.set(new Promotion(fresh));
+            }
```
npeltier: i know i asked you for this, but i feel uncomfortable having this loop on client in main thread. For now can we at least collect expired & published promotions, and trigger an async unpublish?

later on we may want to move this to studio/io

npeltier (changes requested):

studio/src/aem/promotion.js
Andrei4226: Fixed.
`isPromotionPublished` now uses only AEM status (`PUBLISHED / MODIFIED`), not `mas:status/published` and tests updated accordingly.

studio/src/mas-repository.js
Andrei4226: Fixed.
`loadPromotions` collects expired+published items, sets the list and runs unpublish in the background via `#unpublishExpiredPromotions`

studio/src/aem/promotion.js
npeltier: sorry @Andrei4226, so is it fine if a promotion is published & modified to consider it just published? i mean, shouldn't we have a MODIFIED status?

npeltier (approved):
looks good to me! minor remark on the modified status we could treat

afmicka:
@Andrei4226 @npeltier 

1. I was able to save a project without chosen surface. Also if the surface is not chosen, the item picker loads the fragments from the surface in the topnav greyed out field. Is that ok?
Screenshot 2026-05-28 at 12 43 30

2. I can not save the project if the geo is not selected but that field is not marked as required. Therefore, user gets a message "please fill in all the required fields" but not indication what that field might be
Screenshot 2026-05-28 at 12 42 21

3. I guess it should say Project successfully published, not Fragment
Screenshot 2026-05-28 at 12 52 19

Andrei4226:
Added the new changes for promotions:

- MODIFIED status: Published projects with unpublished edits show as MODIFIED in the list
- Save validation: Specific error toasts for missing Title, Promo Code, dates, tags, Geos, Surface or items (required labels)
- Update publish toast: “Project successfully published.”

afmicka:
@Andrei4226 

1. When you update the active (published) project, it gets the modified status but disappears from the list of active projects. the only place i can find that project is tab All. That feels weird to me. I would guess that promotion is still active in the old version, just not with the new changes until published again. Does modification unpublishes the promo (it should not)? Also, i am not sure it should be removed from the Active list since i would expect the promo still to be active

2. I have just noticed, shouldn't these buttons be on the right-hand side like in translations? 
Screenshot 2026-05-28 at 15 15 06

afmicka:
> @Andrei4226 cc @afmicka there is also a not so rare case too where we may want to select non en_US fragments (e.g. Korea only promotion) in which case i guess we may want geo selection in the popup as well

@Andrei4226 @npeltier this implementation seems to be missing in this PR. Do we want it here or in a separate jira/PR?

afmicka (approved):

PR #875, opened 2026-05-18
MWPW-195641 fix(mini-compare-chart): preserve shortDescription across legal price re-renders

yesil:
Resolves https://jira.corp.adobe.com/browse/MWPW-195641

## Summary

`adjustShortDescription` in `mini-compare-chart` moved the `[slot="body-xxs"]` text into `.price-plan-type` (inside the cloned legal `inline-price`) and then removed the source from the DOM. When the legal `inline-price` later re-rendered (locale change, quantity selector change, WCS data refresh, settings flip), it wiped the appended `<em>`. Since the source was already gone, the short description was lost — race condition: removed without ever being displayed (or displayed once then lost).

### Fix

- Cache the `bodyXxs` element reference on `this.shortDescriptionSource` before removing it from the DOM, so the text survives any re-render.
- Rewrite `adjustShortDescription` as idempotent: skips if `<em data-short-description>` is already in `.price-plan-type`; re-applies from the cached source otherwise.
- Listen for the `mas:resolved` event that `inline-price` dispatches after every render-settle cycle (`MasElement.notify()`), and re-run `adjustShortDescription` from the handler. Listener is removed in `disconnectedCallbackHook`.

## 🧪 Nala E2E Tests

### To run Nala tests again:

- Remove the `run nala` label

## Test URLs:

- After: https://mwpw-195641--mas--adobecom.aem.live/

## Test plan

- [ ] Load a Plans page with `mini-compare-chart` (new variant with `merch-whats-included`) and a `body-xxs` short description; confirm it renders as an italic suffix to the plan type.
- [ ] Trigger a legal-price re-render (quantity selector change, locale switch, or `data-display-plan-type` toggle in devtools) and verify the `<em data-short-description>` reappears each time.
- [ ] Devtools: `card.variantLayout.shortDescriptionSource` should hold the detached element reference; no `[slot="body-xxs"]` should exist in the card.
- [ ] Run `card.variantLayout.adjustShortDescription()` twice; confirm only one `<em data-short-description>` exists.


yesil:
Reopening from upstream branch.

PR #876, opened 2026-05-18
MWPW-195641: preserve shortDescription across legal price re-renders

yesil:
Resolves https://jira.corp.adobe.com/browse/MWPW-195641

## Summary

`adjustShortDescription` in `mini-compare-chart` moved the `[slot="body-xxs"]` text into `.price-plan-type` (inside the cloned legal `inline-price`) and then removed the source from the DOM. When the legal `inline-price` later re-rendered (locale change, quantity selector change, WCS data refresh, settings flip), it wiped the appended `<em>`. Since the source was already gone, the short description was lost — race condition: removed without ever being displayed (or displayed once then lost).

### Fix

- Cache the `bodyXxs` element reference on `this.shortDescriptionSource` before removing it from the DOM, so the text survives any re-render.
- Rewrite `adjustShortDescription` as idempotent: skips if `<em data-short-description>` is already in `.price-plan-type`; re-applies from the cached source otherwise.
- Listen for the `mas:resolved` event that `inline-price` dispatches after every render-settle cycle (`MasElement.notify()`), and re-run `adjustShortDescription` from the handler. Listener is removed in `disconnectedCallbackHook`.

## 🧪 Nala E2E Tests

### To run Nala tests again:

- Remove the `run nala` label

## Test URLs:

- Before: https://www.stage.adobe.com/fr/acrobat.html?country=be&akamaiLocale=be
- After: https://www.stage.adobe.com/fr/acrobat.html?country=be&akamaiLocale=be&maslibs=mwpw-195641--mas--adobecom


Axelcureno (commented):
Fix looks right — `mas:resolved` is the correct hook and the duplicate-guard (`planType.querySelector('em')`) is clean.

One catch: in `adjustLegal`, `this.legalAdjusted = true` is set before the `!headingPrice` / `!options` early returns. If either trips during a transient hydration window, the lockout sticks and won't retry. Flip the flag back to `false` on those returns:

```js
if (!headingPrice) { this.legalAdjusted = false; return; }
...
if (!headingPrice?.options) { this.legalAdjusted = false; return; }
```

Also the `<em>` lost its `slot="body-xxs"` attr — `.price-plan-type [slot="body-xxs"]` in the CSS no longer matches. Worth a quick visual check.

Worth adding a unit test for C1.

web-components/src/variants/mini-compare-chart.js
```diff
@@ -572,25 +581,38 @@ export class MiniCompareChart extends VariantLayout {
...
+            }
         } catch {
             // Proceed with other adjustments
         }
     }
 
     adjustShortDescription() {
-        const bodyXxs = this.card.querySelector('[slot="body-xxs"]');
-        const text = bodyXxs?.textContent?.trim();
+        if (!this.shortDescriptionSource) {
```
Axelcureno, line 599: Catch in `adjustLegal`: `this.legalAdjusted = true` is set before the `!headingPrice` / `!options` early returns higher up. If either trips during a transient hydration window, the lockout sticks and won't retry. Flip the flag back on those returns:

```js
if (!headingPrice) { this.legalAdjusted = false; return; }
...
if (!headingPrice?.options) { this.legalAdjusted = false; return; }
```

Side note: the `<em>` lost its `slot="body-xxs"` attr — `.price-plan-type [slot="body-xxs"]` in the CSS no longer matches. Worth a quick visual check.

Axelcureno (approved):
Only one lockout catch, not severe. Rest looks good 👍🏼

web-components/src/variants/mini-compare-chart.js
yesil: Thanks @Axelcureno — good catches.

**`legalAdjusted`:** Keeping the existing pattern for now — same early-return flow as `segment`, `product`, `plans`, etc. Happy to revisit with a repro.

**`slot="body-xxs"`:** The `<em>` lives under `.price-plan-type` inside legal `inline-price`, not as a slotted `merch-card` child, so a slot name would be misleading. `<em>` styling looks correct on the test page; we can add `.price-plan-type em` if we need more than that.

yesil:
Thanks @Axelcureno — good catches.

**`legalAdjusted`:** Keeping the existing pattern for now — same early-return flow as `segment`, `product`, `plans`, etc. Happy to revisit with a repro.

**`slot="body-xxs"`:** The `<em>` lives under `.price-plan-type` inside legal `inline-price`, not as a slotted `merch-card` child, so a slot name would be misleading. `<em>` styling looks correct on the test page; we can add `.price-plan-type em` if we need more than that.

TsayAdobe (approved):

Roycethan (approved):

PR #776, opened 2026-04-21
MWPW-183572: MAS Studio — AI assistant, product catalog, release flow, OST monorepo integration

Axelcureno:
MAS Studio AI assistant, product catalog, release-flow enhancements, and the mas-ost monorepo workspace integration.

Resolves https://jira.corp.adobe.com/browse/MWPW-183572
QA Checklist: https://wiki.corp.adobe.com/display/adobedotcom/M@S+Engineering+QA+Use+Cases

## What is in this PR

Two commits, both tagged with the Jira ticket:

1. **`MWPW-183572: AI chat assistant, product catalog, release flow + IO Runtime`** — the full feature delivery plus the pre-review audit fixes baked in.
2. **`MWPW-183572: integrate mas-ost as a monorepo workspace`** — OST source relocated from tacocat.js/mas-ost into this repo as the 4th npm workspace. Distinct architectural change, separate commit so reviewers can isolate the build/workspace question.

### Feature commit contents

**Studio frontend**
- Conversational AI chat (`mas-chat.js` + friends): Bedrock-backed, persistent sessions via `chat-session-manager`, operation preview + confirmation gating, destructive-tool allowlist enforced on the client.
- Product catalog (`mas-product-catalog.js` / `mas-product-detail.js`): multi-select filters, surface-scoped editor/settings access (`groups.js` introduces `isMasAdmin`, `canAccessSettings`, `getUserSurfaces`), DRAFT landscape toggle, applied-filter chips.
- MCS-integrated release flow: deterministic fragment content sourced from MCS merchandising data, dual-OSI plumbing for plans variants, multi-select OST hookup in `rte/ost.js`.
- `services/mcp-client.js`: defensive guards on MCP response shapes (publish/get/copy/update contract drifts), `AbortController` on the AI chat fetch, URL-scrubbed `logError` helper applied across 14 call sites.
- `utils/ai-card-mapper.js`: XSS-safe badge HTML via escape helpers; preserves explicit falsy field values (`0`, `''`, `false`); trim-aware title fallback chain.
- `utils/mas-chat-helpers.js`: pure helpers extracted for direct unit testing.
- `markdown-parser.js`: LRU-bounded cache to short-circuit repeat renders.

**IO Runtime — AI chat backend (`io/studio/src/ai-chat/`)**
- AWS Bedrock integration with prompt-injection defense via `wrapUntrusted` sentinel envelopes. Attached-card fragment IDs wrapped element-wise instead of raw `JSON.stringify` to close a breakout vector.
[...]

PR #882, opened 2026-05-19
MWPW-195761: integrate mas-ost as monorepo workspace

Axelcureno:
Lifts and shifts the mas-ost Spectrum 2 OST source tree into this monorepo as the 4th npm workspace, replacing the external tacocat-based OST bundle.

Resolves https://jira.corp.adobe.com/browse/MWPW-195761

Originally bundled in #776; extracted for focused review.

## What is in here
- `mas-ost/` workspace: Lit 3 + Spectrum Web Components, Vite build, WTR tests
- `package.json` + `package-lock.json`: register the new workspace
- `studio/ost/index.js`: rebuilt OST IIFE bundle (Vite output)
- `data-testid` attrs on ~25 OST surfaces (search input, filters, product cards, placeholder chips, options, checkout config, Use/Back buttons, promo controls) for stable NALA selectors
- `nala/studio/ost.page.js`: rewritten to use CSS testid selectors (was XPath against the old React OST)

## Fixes in this PR
- 7 stale OST unit tests aligned with current Lit source (country-picker custom dropdown, mas-ost-app store init, OstStore slot ordering)
- Real bug: OstStore `currentSlot` defaulted to `'trial'` — inverted base/trial UX in tryBuy flow. Now `'base'` first, matching `applyFlowSwitch`
- Repo Prettier config applied across the workspace (36 files reformatted)

## Verification
- `npm run build --workspace=mas-ost` → 981 KB IIFE, copied to `studio/ost/index.js`
- `npm test --workspace=mas-ost` → **284 pass / 0 fail**
- `npm test --workspace=mas-ost -- --coverage` → **87.25%** (up from 79.43%, +49 new unit tests across 4 new suites: `mas-ost-promo-tag`, `mas-ost-selection-list`, `mas-ost-offer-detail-focused`, `mas-ost-product-detail`)

- [x] C1. Cover code with Unit Tests — 20 test files, 284 cases, 87.25% coverage
- [ ] C2. Add a Nala test (existing NALA suite covers OST via `individuals_save`, `individuals_edit_and_discard`, `variations`)
- [ ] C.6 read your Jira one more time to validate that you have addressed all ACs and nothing is missing

Test URLs:

- After: https://mwpw-195761--mas--adobecom.aem.live/


npeltier (commented):
mas/mas-ost sounds redundant to me, i would rather pick mas/ost
unless we go with <mas:ost>?

npeltier (changes requested):
actually @Axelcureno can you make this in 2 PR: one being just a move (where we still need to agree on naming ;) ), and the other one a fix? this will help time of review & history

yesil:
Review notes from manual testing — leaving as draft for the author to address.

1. **Stage toggle loses context** — Opening OST by clicking a price, then hitting the stage toggle at the top, drops the user back to the "What would you like to do?" welcome screen instead of preserving the current offer/flow context.

2. **"Use" button emits wrong markup — and the markup display is likely unnecessary** 
   ```
   {{price osi="r_JXAnlFI7xD6FxWKl2ODvZriLYBoSL701Kd1hRyhe8" displayFormatted="true" displayRecurrence="true" forceTaxExclusive="true" displayOldPrice="true" workflowStep="email" wcsOsi="r_JXAnlFI7xD6FxWKl2ODvZriLYBoSL701Kd1hRyhe8"}}
   ```

3. **Price mismatch between header and Live Preview** — For `ccsn_direct_individual` in FR, the price shown in the header differs from the price in the Live Preview section. They should be sourced from the same offer/locale resolution.

4. **No per-price/checkout-link promo cancellation** — When a fragment has a promo applied, there's no way to cancel it on a specific price or checkout link. The convention is the sentinel `promotionCode="cancel-context"` on the individual price / checkout-link.

5. **UX: `sp-action-button` misused as tab toggles** — The top-level switcher is built from `sp-action-button`s acting like tabs, but they aren't real tabs. Use `sp-tabs` / `sp-tab` (proper Spectrum tab semantics, keyboard nav, ARIA) instead of action buttons mimicking tab behavior.

6. **Options section: too many clicks, too much space** — Users must click each button individually to see values for different templates, multiplying clicks. The Options section also takes too much vertical real estate.

ost/dev/debug.html
yesil, line 1: at https://mwpw-195761--mas--adobecom.aem.live/mas-ost/dev/debug.html

debug.html:50 Uncaught ReferenceError: applyToken is not defined
    at HTMLButtonElement.onclick (debug.html:50:44)

ost/src/controllers/checkout-controller.js
yesil, line 1: this is used only in MasOstCheckoutOptions, was this for state management purpose?

if so, can I shill you https://www.npmjs.com/package/picosm :)

ost/src/controllers/placeholder-options-controller.js
```diff
@@ -0,0 +1,38 @@
+export class PlaceholderOptionsController {
```
yesil: same here, it is used as part of MasOstPlaceholderPanel.

It does not bother me directly, but it will create a false pattern for the next AI which will try to imitate this approach.

From what I know, a controller should not be specific to another web component.

ost/src/controllers/search-controller.js
```diff
@@ -0,0 +1,104 @@
+import { isOfferId, isOfferSelectorId } from '../utils/offer-utils.js';
+import { getOfferSelector, getOfferById } from '../utils/aos-client.js';
+import { store } from '../store/ost-store.js';
+
+export class SearchController {
```
yesil: same here, specific to MasOstSearch.

If this pattern is recommended somehow by Lit, let's discuss, maybe this is the right approach to state management. But I'm not sold on the idea yet ::)

ost/src/data/countries.js
yesil, line 1: do we need both js and json files?

ost/src/data/language-mappings.js
yesil, line 1: ideally we should be able to re-use those of Studio and avoid creating duplicates.

ost/src/utils/plan-types.js
```diff
@@ -0,0 +1,52 @@
...
+ */
+const PlanType = { ABM, PUF, M2M, PERPETUAL, P3Y };
+
+/**
+ * Computes the plan type from commitment and term strings.
+ * @param {string|undefined} commitment
+ * @param {string|undefined} term
+ * @returns {string} plan type label or empty string
+ */
+function computePlanType(commitment, term) {
```
yesil: this is duplicating the logic of @dexter/tacocat-core which we still use. I wonder here if we are duplicating logic or not.

ost/src/utils/promo-status.js
```diff
@@ -0,0 +1,36 @@
...
+    return `${promoText(promo)}${suffix}`;
+}
+
+/**
+ * Computes the full promo status from overridden and configured promo codes.
+ * @param {string|undefined} overriden - overridden promotion code
+ * @param {string|undefined} configured - configured promotion code
+ * @returns {{ effectivePromoCode: string|undefined, overridenPromoCode: string|undefined, className: string, text: string, variant: string, isOverriden: boolean }}
+ */
+function computePromoStatus(overriden, configured) {
```
yesil: same we are still using computePromoStatus from @dexter/tacocat-core in price.js

I find the core methods very sensible and duplicating them would guarantee us nasty bugs.

ost/src/utils/safe-define.js
```diff
@@ -0,0 +1,29 @@
+// Guard against duplicate custom element registration.
```
yesil: I don't like this kind of guards, they tend to hide the real issue.
Let it fail and understand why it fails and we fix the root cause.

ost/src/components/ost-selection-list.js
```diff
@@ -0,0 +1,304 @@
...
+        return html`
+            <div class="confirm-bar">
+                <span class="confirm-text">Keep "${offerNames}" in your new selection?</span>
+                <sp-button size="s" variant="primary" @click=${() => store.confirmFlowSwitch(true)}>Keep</sp-button>
+                <sp-button size="s" variant="secondary" @click=${() => store.confirmFlowSwitch(false)}>Discard</sp-button>
+            </div>
+        `;
+    }
+
+    renderTryBuy() {
```
yesil: I see a lot of renderXyz methods, can we use `get xyz` method?

yesil (changes requested):
in Studio, we don't import spectrum web components whenever they are used, but assume that the import is in swc.js

I prefer `get abc` to `renderAbc` because ultimately, calling the method does not render it but return the markup, also this is the existing pattern in our codebase.

please avoid having inline js in html markup in event handlers, suggesting to move those to specific handler functions.

ost/src/utils/plan-types.js
Axelcureno: I thought the plan was to deprecate tacocat.js in which case this would be a **_replacement_** rather than duplication.

Axelcureno: As discussed on slack, this will be handled in a separate ticket and for now will continue using tacocat-core.

ost/src/components/ost-country-picker.js
```diff
@@ -0,0 +1,250 @@
...
+
+    getFilteredCountries() {
+        const query = this.countryInput;
+        if (!query) return this.countries;
+        return this.countries.filter((c) => c.startsWith(query));
+    }
+
+    handleEnvToggle(event) {
+        const env = event.target.checked ? 'STAGE' : 'PRODUCTION';
+        store.setEnv(env);
```
Axelcureno, line 195: Fix for the stage-toggle context loss. Was calling store.init({env}) which wipes selectedProduct/Offer/OSI/flowChosen/selectedOffers/deepLink. New store.setEnv only updates env and clears the cached offers (so they re-resolve at the new env); all flow context is preserved. Commit f260660be.

ost/src/components/ost-code-output.js
```diff
@@ -0,0 +1,200 @@
...
+        }
+
+        const typeConfig = store.placeholderTypes.find((t) => t.type === type);
+        const typeOverrides = typeConfig?.overrides || {};
+        const defaults = { ...store.defaultPlaceholderOptions, ...typeOverrides };
+
+        const optionParts = [];
+        Object.entries(options).forEach(([key, val]) => {
+            if (val === undefined || val === false || val === '') return;
+            if (defaults[key] === val) return;
```
Axelcureno, line 88: Fixed in d0b022fcd. Computes defaults from store.defaultPlaceholderOptions merged with the type's overrides, then skips emitting any option that matches its default — mirrors the getObjectDifference pass in studio/src/rte/ost.js. Also added promotionCode emission below when storedPromoOverride is set. Three new tests cover both.

ost/src/components/ost-product-detail.js
```diff
@@ -0,0 +1,501 @@
...
+
+    render() {
+        const product = store.selectedProduct;
+        if (!product) {
+            return html`<div class="empty-state">Select a product to view offers.</div>`;
+        }
+
+        const arrangementCode = product.arrangement_code || product.arrangementCode || store.aosParams.arrangementCode;
+
+        if (this.summary) {
```
Axelcureno, line 422: Root cause was two different resolution paths: header used raw AOS price_details (tax-inclusive in FR), Live Preview re-resolved through mas-commerce-service / WCS with the user's tax-display options applied. Removed the static header price entirely in c1eb5ede7 — Live Preview is now the single source of truth and shows the price right below the header.

ost/src/components/ost-promo-tag.js
```diff
@@ -0,0 +1,91 @@
...
+                ${hasContextualPromo
+                    ? html`
+                          <sp-action-button
+                              data-testid="ost-promo-cancel-context"
+                              quiet
+                              size="s"
+                              ?selected=${isCancelled}
+                              @click=${() => store.setPromoCode(isCancelled ? undefined : PROMO_CONTEXT_CANCEL_VALUE)}
+                          >
+                              ${isCancelled ? 'Restore context promo' : 'Cancel context promo'}
```
Axelcureno, line 82: Added in 91387077e. New quiet button appears only when a contextual promotionCode is set on the fragment; clicking sets storedPromoOverride to the sentinel PROMO_CONTEXT_CANCEL_VALUE. The previous commit (d0b022fcd) wires ost-code-output to emit promotionCode="cancel-context" in the markup. Button toggles between "Cancel context promo" and "Restore context promo".

ost/src/components/ost-placeholder-panel.js
```diff
@@ -0,0 +1,196 @@
...
+        }
+
+        const types = store.placeholderTypes;
+        const selected = this.selectedType;
+
+        return html`
+            <div class="section-label">
+                Placeholder Type <ost-help-icon text="${HELP_TOOLTIPS.placeholderType}"></ost-help-icon>
+            </div>
+            <sp-tabs size="s" selected=${selected} @change=${(e) => this.selectType(e.target.selected)}>
```
Axelcureno, line 148: Swapped in d2dd72ffa. The placeholder type row is now sp-tabs + sp-tab — real Spectrum tab semantics, keyboard nav, ARIA. Added sp-tabs / sp-tab to ost/src/swc.js. Existing NALA data-testid=ost-placeholder-chip-<type> selectors preserved on each sp-tab so e2e tests are unaffected.

ost/src/components/ost-placeholder-options.js
```diff
@@ -0,0 +1,77 @@
...
+    forceTaxExclusive: 'Exclude tax from price',
+    displayOldPrice: 'Show original price',
+};
+
+export class OstPlaceholderOptions extends LitElement {
+    static styles = css`
+        :host {
+            font-family: inherit;
+            display: grid;
+            grid-template-columns: repeat(2, 1fr);
```
Axelcureno, line 18: 2-column grid in a35b32b4e: 6 options now fit in 3 rows instead of 6. Row height 40px → 32px, label 14px → 13px to match Spectrum's compact density. Cuts the section's vertical footprint by roughly half. NALA selectors (data-testid on each sp-switch) unchanged. The "fewer clicks" half of your note is addressed by the sp-tabs swap.

ost/dev/debug.html
Axelcureno: Fixed in ca16ea8c8. The `window.applyToken =` assignment was gated behind a top-level `await import(...)` so the click could fire before the assignment ran. Replaced the inline `onclick="applyToken()"` with an addEventListener on the button inside the same module script — kills the race and removes the inline-JS-in-HTML smell in one edit.

ost/src/controllers/checkout-controller.js
Axelcureno: Agreed. Inlined CheckoutController into OstCheckoutOptions in 9e93ab229 — fields and methods now live on the component, with `get checkout() { return this }` as a back-compat getter so external `querySelector(...).checkout` reads still work. Controller file + tests deleted. Added a note in ost/CLAUDE.md so the pattern isn't recreated. Will look at picosm in a follow-up if we standardize Studio state.

ost/src/controllers/placeholder-options-controller.js
Axelcureno: Inlined into OstPlaceholderPanel in 9e93ab229 along with checkout and search. The component now owns selectedType / options and exposes `get placeholderCtrl() { return this }` so ost-placeholder-options and ost-code-output keep reading via `panel.placeholderCtrl.options` without changes. The "false pattern" concern is addressed by a CLAUDE.md note under ost/ explaining controllers must be reusable across components.

ost/src/controllers/search-controller.js
Axelcureno: Lit's docs actually reserve Reactive Controllers for cross-component behaviors (clipboard, focus-trap, gestures) — a controller specific to one component is using the wrong tool. Inlined SearchController into OstSearch in 9e93ab229; the `detectType` regex matcher and the AOS resolve calls now live on the component. Same back-compat `get search() { return this }` to keep any external readers working.

ost/src/data/countries.js
Axelcureno: No — the .json copy was dead, only the .js file is imported (by ost-country-picker). Deleted both .json duplicates (countries.json, language-mappings.json) in 59d442fb0.

ost/src/data/language-mappings.js
Axelcureno: Grepped studio/src — no canonical module to reuse today (only hits were inside compiled studio/libs/ bundles). The real source is tacocat-OST's country→lang map. Left a one-line comment in d645b58e2 pointing at this so the next maintainer knows to consolidate when a shared module lands in mas/, rather than re-deriving the table. Happy to move it earlier if you want a specific home now.

ost/src/utils/plan-types.js
Axelcureno: Per the slack follow-up (tacocat.js stays for now, deprecation goes in a separate ticket): switched to importing applyPlanType from @dexter/tacocat-core/src/wcsUtils.js in 91130f91d, and deleted the ost duplicate. Deep-import path is intentional — the barrel transitively pulls in @pandora/data-models-odm which WTR's CommonJS loader chokes on. When the deprecation ticket lands, the canonical-impl swap will be trivial.

ost/src/utils/promo-status.js
Axelcureno: Same fix in 91130f91d. ost-code-output and ost-promo-tag now import computePromoStatus + PROMO_CONTEXT_CANCEL_VALUE from @dexter/tacocat-core/src/promotion.js (deep path avoids the pandora barrel issue). Duplicate ost/src/utils/promo-status.js and its tests removed. Tests for downstream behavior live in ost-code-output.test.js + ost-promo-tag.test.js so coverage on the integration path stays intact.

ost/src/utils/safe-define.js
Axelcureno: Removed in 1a1d56d30. Killed both the runtime guard AND the vite plugins (safeDefinePlugin + safeDefineEsbuildPlugin) that did the same thing at build time — same critique. The real fix lives in ost/src/swc.js: all @spectrum-web-components/* imports are now declared once there and pulled out of component files, so we don't hit re-registration in the first place. Component imports cleanly fail if a tag isn't in the registry.

ost/src/components/ost-selection-list.js
Axelcureno: Applied across the codebase in 40f02a6af:
- ost-selection-list: renderConfirmBar → get confirmBar; renderTryBuy → get tryBuyContent; renderBundle → get bundleContent
- ost-app: renderRightPanel → get rightPanel; renderContent → get content
- ost-product-list: renderSkeletons → get skeletons

Skipped ost-live-preview.renderPreview — that one's imperative shadow-DOM work (removes/appends children directly), not a template return. Converting it to a getter would silently break the rendering side effect.

Axelcureno:
Split per @npeltier's review into two stacked PRs (parent ticket MWPW-195761):

- **#894** (MWPW-195919) — ships the new Lit OST behind `?ost=new`, legacy tacocat stays default. No production behavior change on merge.
- **#895** (MWPW-195920) — flips default to new OST and deletes the legacy bundle. Stacked on #894; merge only after #894 soaks.

Leaving this PR open as a historical record of the review threads — please move new comments to #894 / #895.

PR #888, opened 2026-05-20
MWPW-195137 fixing country & pzn in state cache key

npeltier:
both those parameters change the response, and are invalidating previous cache hash & last modified

Splunk export — 30 min window (2026-05-20 03:30–03:59 UTC), fragment 817d9a19…/en_US
  
  - 2938 pipeline completions, 0 × 304, 100% 200. Despite no content change, every single request was served full.
  - 4 distinct api_keys hit the same fragment: CreativeCloud_v6_9 (2762 ×, 94%), v6_8 (173 ×), v6_10 (2 ×), v6_6 (1 ×). This confirmed multiple callers colliding on a single state bucket — the old cache
  key req-${id}-${locale} was shared by all of them.
  - Even the dominant key (94%) never got a 304, which ruled out api_key collision as the sole cause and pointed to country / pzn varying per request — different countries from Akamai geo produce
  different body hashes and re-stamp lastModified each time.
  - Timing: 622 calls sub-100 ms (warm in-container state), 1508 in the 200–500 ms range — confirming the state was being found but hash mismatches were preventing 304.

Bottom line: the cache key was computed *after* cache was looked at, so fr_LU cache key was never found

Resolves https://jira.corp.adobe.com/browse/MWPW-195137
partly resolves https://jira.corp.adobe.com/browse/MWPW-195647 (that will become less critical)

## Test URLs:

non regression only, no "visible" feature

Preview

- non regression & fix on us plans page https://main--cc--adobecom.aem.page/creativecloud/plans?maslibs=MWPW-195137
- non regression on fr plans page https://main--cc--adobecom.aem.page/fr/creativecloud/plans?maslibs=MWPW-195137
- non regression on catalog page https://main--cc--adobecom.aem.page/products/catalog?maslibs=MWPW-195137

mas/io

- non regression on us plans page https://main--cc--adobecom.aem.live/creativecloud/plans?mas-io-url=https://14257-merchatscale-npeltier.adobeioruntime.net/api/v1/web/MerchAtScale
- non regression on fr plans page https://main--cc--adobecom.aem.live/fr/creativecloud/plans?mas-io-url=https://14257-merchatscale-npeltier.adobeioruntime.net/api/v1/web/MerchAtScale
- non regression on catalog page https://main--cc--adobecom.aem.live/products/catalog?mas-io-url=https://14257-merchatscale-npeltier.adobeioruntime.net/api/v1/web/MerchAtScale


seanchoi0 (approved):
lgtm

npeltier:
got misguided by {{context.locale}} looked in the state and we already have "regionalized" locales which should be enough for now (no pzn) will come back after digging a bit more

npeltier:
> got misguided by {{context.locale}} looked in the state and we already have "regionalized" locales which should be enough for now (no pzn) will come back after digging a bit more

while we have fr_LU cache keys, those entries are never hit, because at that stage locale was still fr_FR (with country LU). this made me think we should not have mutated locale as it is very misleading, which i fixed together with cache key.

Also if locale is not found/valid, send 400, which partially solves https://jira.corp.adobe.com/browse/MWPW-195647

yesil (approved):
👍

mirafedas (approved):

honstar:
I'm sharing an issue that Claude flagged which I don't understand 100% but still worth sharing imo:
```
settings.js and replace.js still consume context.locale — they no longer see the regional locale (likely regression)

  Removing locale: regionLocale from customize.js:345 and defaultLanguage.js (both init return and defaultLanguageProcess) is necessary for the cache-key fix (so the write key matches the read key). wcs.js:103
  was updated to compensate by reading regionLocale ?? locale. But two other downstream consumers were missed:

  - io/www/src/fragment/transformers/settings.js:286 — destructures locale from context and passes it to applySettings / applyCollectionSettings, which match against overrideSetting.locales at settings.js:173.
  The mocked settings (io/www/test/fragment/mocks/settings-sandbox.json:52) include overrides like "locales": ["fr_FR", "fr_BE", "fr_CH"] and "locales": ["en_AU"]. After this PR a request with locale=fr_FR +
  country=BE resolves regionLocale=fr_BE, but settings sees locale=fr_FR — any setting override targeting only a regional code (e.g. ["fr_BE"] or ["en_AU"]) will silently stop firing for the country-override
  flow.
  - io/www/src/fragment/transformers/replace.js:60-61 — getDictionaryId builds the Odin URL from context.locale. Before this PR, replace.init() merged defaultLanguage's result (which carried locale: regionLocale)
   into context at replace.js:163, so the dictionary URL hit the regional path (e.g. …/sandbox/fr_BE/dictionary/index). After this PR the URL hits the requested locale (e.g. …/sandbox/fr_FR/dictionary/index). The
   comment at replace.js:158 — "Dictionary cache key needs merged locale (region) from defaultLanguage init" — is now stale and misleading.

  The QA test pages in the PR description (/creativecloud/plans, /fr/creativecloud/plans, /products/catalog) all use default locales (en_US, fr_FR with no country override), so they will not surface either
  regression. The pipeline tests that involve country=CA (pipeline-e2e.test.js:124) pass only because the relevant fr_FR dictionary URL happens to be explicitly stubbed to 404 in that test — the assertion
  dictionary-id != 'sandbox_fr_FR_dictionary' now passes by side-effect rather than by intent.
  
  This wasn't fixed in MWPW-189894 by accident: PR #748 explicitly added locale: regionLocale to make replace.init use the regional dictionary. Dropping it should be a deliberate, documented behavior change — not
   collateral damage of the cache-key fix.
```

I'll leave it to you @npeltier whether this is actually a concern or not.

io/www/src/fragment/pipeline.js
```diff
@@ -152,6 +153,15 @@ async function main(params) {
 }
 
 async function mainProcess(context) {
+    if (!context.id || !context.locale) {
```
honstar, line 156: The exact same check is present in https://github.com/adobecom/mas/blob/main/io/www/src/fragment/transformers/fetchFragment.js#L14-L16, the latter becoming dead code it seems and can be removed.

honstar:
There is a unit-test failure:
```
1) pipeline configuration caching
       "after each" hook for "should use stale cache when configuration refresh times out":

      AssertionError: unexpected fetches: https://odin.adobe.com/adobe/contentFragments/417dd2aa-c08b-46fe-8d18-8a7f9be85666?references=all-hydrated: expected [ Array(1) ] to have a length of +0 but got 1
      + expected - actual

      -1
      +0
      
      at Context.<anonymous> (file:///home/runner/work/mas/mas/io/www/test/fragment/pipeline-config.test.js:32:96)
      at process.processImmediate (node:internal/timers:484:21)
```

io/www/src/fragment/transformers/wcs.js
```diff
@@ -73,7 +73,7 @@ async function wcs(context) {
         log(`No WCS configurations found for API key ${context.api_key}`, context);
         return context;
     }
-    const { body, locale } = context;
+    const { body, locale, regionLocale } = context;
```
3ch023: 3 different locales in context object, could you add a js doc for them? smth like
```
/**
   * @typedef {Object} PipelineContext
   * @property {string} locale - Original request locale (e.g. 'fr_FR'). Never mutated by transformers.
   * @property {string} regionLocale - Resolved regional variant based on locale + country + surface (e.g. 'fr_LU').
   * @property {string} defaultLocale - Base locale for the surface, used to fetch the default fragment (e.g. 'fr_FR').
   * @property {string} id - Fragment ID
   * @property {string} country - Explicit country override from request (e.g. 'LU')
   * @property {string} surface - Surface identifier (e.g. 'acom')
   * ...
   */

```

3ch023 (approved):

afmicka:
@npeltier the IO deploy is failing and also Nala is still failing for the same reason. Collections are not loading for country=AR
https://mwpw-195137--mas--adobecom.aem.live/web-components/docs/plans-collection.html?mas-io-url=https://14257-merchatscale-npeltier.adobeioruntime.net/api/v1/web/MerchAtScale&country=AR&language=en

see prod https://main--mas--adobecom.aem.live/web-components/docs/plans-collection.html?country=AR&language=en

afmicka:
@npeltier with your changes the annual price is NOT displaying on ccd mini in AU.
https://mwpw-195137--mas--adobecom.aem.live/web-components/docs/ccd-mini.html?country=AU&language=en&mas-io-url=https%3A%2F%2F14257-merchatscale-npeltier.adobeioruntime.net%2Fapi%2Fv1%2Fweb%2FMerchAtScale


prod: 
Screenshot 2026-05-22 at 11 44 05

afmicka (approved):

PR #894, opened 2026-05-20
MWPW-195919: ship new OST behind ?ost=new flag (legacy default)

Axelcureno:
Resolves https://jira.corp.adobe.com/browse/MWPW-195919

Splits the OST monorepo work from [#882](https://github.com/adobecom/mas/pull/882) per @npeltier's review — this PR ships the new Lit OST behind a `?ost=new` flag with the legacy tacocat bundle as the default. Production behavior is unchanged on merge.

Parent ticket: [MWPW-195761](https://jira.corp.adobe.com/browse/MWPW-195761) · Sibling follow-up: [MWPW-195920](https://jira.corp.adobe.com/browse/MWPW-195920) (flip default + delete legacy)

## What is in here

- All 26 commits from PR #882 (workspace rename, NALA fixes, behavior fixes, controller inlines, tacocat-core dedup, sp-tabs swap, Options grid, etc.) carry forward unchanged
- **New in this PR:**
  - `studio.html` inline loader flipped: legacy bundle loads by default, `?ost=new` or `#ost=new` opts into the new Lit OST
  - `ost/src/swc.js` rewritten with a scoped `customElements.define` guard so the new OST can co-exist with Studio's own SWC registry without throwing `NotSupportedError`
  - NALA specs that exercise the OST get `ost=new&` prepended to their hash params (the new OST's `data-testid` selectors are what NALA targets)
  - Small dead-code cleanup in `ost-code-output.handleUse` (dropped a no-op `clearTimeout` inside its own firing callback)

## Smoke-tested locally

| URL | Loaded bundle | `window.ost` | `ost-app` |
|---|---|---|---|
| `studio.html` | `studio/ost-legacy/index.js` | defined | not defined (correct — React) |
| `studio.html?ost=new` | `studio/ost/index.js` | defined | defined |
| `studio.html#ost=new` | `studio/ost/index.js` | defined | defined |

No console errors in any mode after the swc.js guard fix.

## Verification

- `npm test --workspace=ost` → 296 / 296 pass
- `npm run build --workspace=ost` → 1.01 MB IIFE, copied to `studio/ost/index.js`
- Manual Chrome smoke at `http://localhost:3028/studio.html` (see table above)

## Test URLs

- After: https://mwpw-195919--mas--adobecom.aem.live/

## Checklist

- [x] C1. Cover code with Unit Tests — 20 test files, 296 cases
- [ ] C2. Add a Nala test (existing NALA suite covers OST via `individuals_save`, `individuals_edit_and_discard`, `variations`; all updated to use `?ost=new`)
- [ ] C3. Verify all Checks are green
- [ ] C4. PR description contains working Test Page link
- [ ] C5. Ready to do a demo from Test Page
- [ ] C6. Re-read Jira to validate all ACs are addressed


ost/src/components/ost-filter-bar.js
```diff
@@ -0,0 +1,305 @@
+import { LitElement, html, css } from 'lit';
+import { store } from '../store/ost-store.js';
+import { HELP_TOOLTIPS } from '../data/help-content.js';
+import './ost-help-icon.js';
+
+const PLAN_TYPES = [
```
yesil, line 6: I see there are still duplicates here, but in order not to block new OST, I propose to raise a new ticket to deprecate tacocat-core these tacocat core related constants that should be moved in web-components folder.
Only Studio related labels etc should stay here. 
The ask is not code change but track with JIRA.

ost/dev/debug.html
yesil, line 1: not sure if any of these dev files needed, but if yes, why not just have one?

ost/src/components/ost-live-preview.js
```diff
@@ -0,0 +1,223 @@
...
+        }
+
+        placeholderOptions.workflow = 'UCv3';
+        placeholderOptions.marketSegment = store.aosParams.marketSegment || 'COM';
+
+        return { type, placeholderOptions, service };
+    }
+
+    renderPreview() {
+        const container = this.shadowRoot?.querySelector('.placeholder-container');
```
yesil: I don't know why we need such dom manipulations in an application with Lit.

ost/src/components/ost-offer-card.js
```diff
@@ -0,0 +1,230 @@
...
+            <span class="cell" @click=${this.handleClick}>
+                ${planType ? html`<sp-badge size="s" variant="${badgeVariant}">${planType}</sp-badge>` : ''}
+            </span>
+            <span class="cell" @click=${this.handleClick}>
+                ${offer.offer_type
+                    ? html`<sp-badge size="s" variant="neutral">${offer.offer_type}</sp-badge>${getTrialDays(offer) !== null
+                              ? html`<span class="trial-days">${getTrialDays(offer)}d</span>`
+                              : ''}`
+                    : ''}
+                ${offer.__landscape
```
yesil: Please avoid using _ or __ as variable prefixes, this is not a pattern we follow in this repository.

ost/src/components/ost-offer-detail-focused.js
```diff
@@ -0,0 +1,388 @@
...
+
+function formatPeriod(offer) {
+    const term = offer?.term;
+    const commitment = offer?.commitment;
+    if (commitment === 'YEAR' || term === 'ANNUAL') return '/yr';
+    if (commitment === 'MONTH' || term === 'MONTHLY') return '/mo';
+    return '';
+}
+
+const PLAN_BADGE_VARIANT = {
```
yesil: this is a duplicate of PLAN_TYPE_COLORS in ost-offer-card.js

ost/src/controllers/.gitkeep
yesil, line 1: remove empty folders please

ost/src/store/ost-store.js
```diff
@@ -0,0 +1,482 @@
...
+    displayRecurrence: true,
+    displayPerUnit: false,
+    displayTax: false,
+    forceTaxExclusive: false,
+    displayOldPrice: true,
+};
+
+const VALID_FLOWS = ['single', 'tryBuy', 'bundle', 'consult'];
+
+export class OstStore extends EventTarget {
```
yesil, line 85: this is a custom store implementation, why not follow existing Store pattern?

ost/src/global.js
```diff
@@ -0,0 +1,86 @@
+import './swc.js';
+import './components/ost-app.js';
+import { store } from './store/ost-store.js';
+
+function openOfferSelectorTool(options) {
```
yesil, line 5: I see this is the same pattern we had in tacocat.js for which I'm the culprit.
I wonder if we can improve this rather then copying the concept from tacocat.js repo to here.

ost/.npmrc
```diff
@@ -0,0 +1,4 @@
+registry=https://registry.npmjs.org/
+@spectrum-web-components:registry=https://registry.npmjs.org/
+@pandora:registry=https://artifactory.corp.adobe.com/artifactory/api/npm/npm-pandora-release
+//artifactory.corp.adobe.com/artifactory/api/npm/npm-pandora-release/:_authToken=${NPM_TOKEN}
```
yesil: we should not have any ref to artifactory.corp.adobe.com, is this really needed?

ost/vite.config.js
yesil, line 1: We use esbuild for bundling, but for the new OST, I don't think it is needed and we want to bundle.
Please follow as we do in Studio, meaning deliver unbundled vanilla js files.

yesil (changes requested):

ost/src/global.js
npeltier: can we make this a fix in the second PR?

npeltier:
@yesil don't you think we should move all blocking PR reviews to the other PR and let this one in? for me it's mostly a move PR here, that is hidden behind a feature flag, so i'm fine getting it in (once we validate it does not change anything to the main behaviour)

ost/src/global.js
yesil: yes, @Axelcureno I don't have hard blockers, but these things should at least be tracked for fixing in the future.

yesil:
@npeltier I thought the other PR was just about turning the FF on.
I'm fine, I'll discard my request change, but will ask to see this changes in the other PR.

yesil:
@Axelcureno OST doesn't work at https://mwpw-195919--mas--adobecom.aem.live/studio.htm

please check console errors.

ost/src/global.js
Axelcureno: @yesil These are all great observations you made. I'd rather fix them first before merging if that's okay.

ost/dev/debug.html
3ch023: yeah we can cleanup once its in mas

Axelcureno:
@3ch023 @yesil @npeltier 
---

Round 2 review addressed (5 commits since the last review).

Fixed in-PR:
- B1 (OST not loading): reproduced — stale browser cache after `32cd5a2f9` dropped `studio/ost-legacy/`. Hard-refresh resolves; verified clean both modes.
- 2eba9f37a: dev/ dedupe, `__landscape` → `landscapeSource`, PLAN_TYPE_COLORS extracted, dead .gitkeeps, .npmrc + @pandora dropped
- afde205bb: ost-live-preview off DOM ops, onto Lit templates (willUpdate + node interpolation)
- 8d6a8b26b: OstStore now backed by ReactiveStore primitives (same as Studio); back-compat getters/setters keep the 17 existing consumers working
- 07d7359d5: coverage 87% → 92% (+58 tests). Also fixes a real bug surfaced by the new tests — the C8 migration mis-stored `onSelect`/`onCancel`/`onMultiSelect`/`onBundleSelect` callbacks because `ReactiveStore.set(fn)` treats functions as updaters. Reverted those four to plain fields.

Tracked per your explicit "JIRA" / "fix in second PR" asks in the threads:
- MWPW-195988 — drop Vite, ship unbundled like Studio
- MWPW-195989 — promote ReactiveStore to a shared package (out of `studio/`)
- MWPW-195991 — deprecate tacocat-core duplicate constants from `ost-filter-bar.js`
- MWPW-195992 — redesign `global.js` init away from tacocat.js style

All 312 OST unit tests pass.

ost/src/components/ost-filter-bar.js
Axelcureno: Tracked as MWPW-195991.

ost/dev/debug.html
Axelcureno: Done in 2eba9f37a — kept ost/dev/index.html, deleted debug.html and module-test.html (neither was wired into any script or CI).

ost/src/components/ost-live-preview.js
Axelcureno: Fixed in afde205bb. Replaced `updated()` + `renderPreview()`'s `while`/`removeChild` + `appendChild` + `document.createElement` with `willUpdate()` building a node + a Lit template interpolating `${this.#placeholderNode ?? nothing}`. The external factory (`createInlinePrice` / `createCheckoutButton`) still returns a live custom element, but Lit handles insertion/removal directly. File is structurally cleaner (~30 imperative lines gone) and all tests pass.

ost/src/components/ost-offer-card.js
Axelcureno: Fixed in 2eba9f37a. Renamed `offer.__landscape` → `offer.landscapeSource` across `ost-offer-card.js` + `ost-product-detail.js` (we own the field — it's set at line 376 of `ost-product-detail.js` to tag offers with which AOS landscape returned them).

ost/src/components/ost-offer-detail-focused.js
Axelcureno: Fixed in 2eba9f37a. Extracted to `ost/src/data/plan-type-colors.js`; both consumers import from the same source. The duplicate was named `PLAN_BADGE_VARIANT` here but held the identical map — picked the more general `PLAN_TYPE_COLORS` name since the map could feed more than badges.

ost/src/controllers/.gitkeep
Axelcureno: Fixed in 2eba9f37a. Deleted `ost/src/controllers/.gitkeep` and 4 other dead `.gitkeep` files in folders that already had real content (`components/`, `data/`, `store/`, `utils/`). The `controllers/` folder drops automatically since nothing else was tracked there.

ost/src/store/ost-store.js
Axelcureno: Fixed in 8d6a8b26b. OstStore now backs every slice with the canonical `ReactiveStore` from `studio/src/reactivity/reactive-store.js`. To avoid breaking the 17 existing consumers, kept the ergonomic `store.foo` getter/setter (`Object.defineProperty` proxies to `slice.get()`/`set()`); the underlying ReactiveStore is reachable via `store.stores.foo` so future consumers can attach a `StoreController`. Multi-mutation methods use a small batch primitive so legacy `store.subscribe(handler)` sees one event per logical op (preserving prior behavior). All 312 tests pass.

Cross-workspace import wasn't clean (the `ost` workspace can't reach into `studio/` without breaking npm boundaries), so the three reactivity files are copied verbatim. Promotion to a shared package tracked in **MWPW-195989**.

Follow-up note: the migration also surfaced and fixed a bug — `ReactiveStore.set(fn)` treats functions as updater functions, which mangled the `on*` callback fields. Reverted those four (`onSelect` / `onCancel` / `onMultiSelect` / `onBundleSelect`) to plain class fields; `init()` still forwards them from config. Fix is in 07d7359d5, caught by the coverage tests.

ost/src/global.js
Axelcureno: Tracked as MWPW-195992 per your "track for future" reply on this thread.

ost/.npmrc
Axelcureno: Fixed in 2eba9f37a. Confirmed nothing in `ost/src/` imports from `@pandora/*` — the registry config and the four `@pandora/*` `optionalDependencies` were inherited from the legacy tacocat React OST. Deleted `ost/.npmrc` entirely and removed the four `optionalDependencies` from `ost/package.json`. `npm install` from repo root succeeds cleanly.

ost/vite.config.js
Axelcureno: Tracked as MWPW-195988. On investigation, this is bigger than swapping the build: Studio's importmap only declares `lit`, while OST also imports ~40 `@spectrum-web-components/*` paths plus `@dexter/tacocat-core/src/promotion.js` and `wcsUtils.js` — none of which would resolve unbundled today. The ticket captures three implementation options (extend importmap; slim deps-only pre-bundle; drop the 607KB CSS separately).

npeltier (approved):
i'm good merging this in almost blindly assuming we fix everything and thoroughly review the "enabling" PR

yesil (approved):
Thanks @Axelcureno for tracking JIRAs.

nala/studio/acom/plans/individuals/specs/individuals_edit_and_discard.spec.js
```diff
@@ -9,7 +9,7 @@ export default {
             data: {
                 cardid: '45e50a68-9bd7-4fc2-9665-12f39140a1be',
             },
-            browserParams: '#page=fragment-editor&path=nala&fragmentId=',
+            browserParams: '#ost=new&page=fragment-editor&path=nala&fragmentId=',
```
afmicka: @Axelcureno is this really needed in all these tests cases in nala spec files? does "ost" param always show when user is in the editor or just when opening the ost? These here is for the fragment url when in editor, please don't change if not needed. Especially if all these changes are not going to be active after the merge.

afmicka:
@Axelcureno Nala is failing due to the changes in authorization. Please check, the test are not executing.

nala/studio/acom/plans/individuals/specs/individuals_edit_and_discard.spec.js
Axelcureno: Good call — removed `ost=new` from all three specs (4cdc1f848), not just these editor URLs. Since the new OST ships behind the flag with legacy as the default, the NALA suite should track default behavior. The param only ever mattered at initial page load anyway. New-OST E2E coverage will come back in MWPW-195920 when we flip the default and delete legacy.

Axelcureno: Not needed for this PR, I have removed it. Thanks @afmicka

nala/libs/auth.setup.cjs
```diff
@@ -45,8 +45,15 @@ setup('authenticate, @mas-studio', async ({ page, baseURL, browserName }) => {
         // Passkey dialog may not appear — continue
     }
 
-    await page.waitForURL(`${baseURL}/studio.html#page=welcome&path=sandbox`);
-    await expect(page).toHaveURL(`${baseURL}/studio.html#page=welcome&path=sandbox`);
+    const escapedBaseURL = baseURL.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
```
afmicka, line 48: i feel like these changes are unnecessary and also they remove the indirect check that the studio indeed loads with the default surface (which is sandbox). 

I don't mind having this but don't see the need of touching this file in this context.

afmicka:
@Axelcureno regression wise this looks ok and we can merge this PR from my end.

The new OST has bunch of issues. I have not check all yet but some of what i have seen:
1. promo code is not populated nor applied in the preview because it is not stored under promoCode but as an stored override (i guess). This may cause a lot of issues to the authors on existing authored cards
Screenshot 2026-05-27 at 14 08 35

2. Strikethrough price preview contains text that is not really displayed on the card
Screenshot 2026-05-27 at 14 09 42

3. there should be discount percentage for the price with promo, this one shows 30% in the old OST
Screenshot 2026-05-27 at 14 11 08


4. Legal disclaimer is not showing preview
Screenshot 2026-05-27 at 14 11 32

5. Country menu does not indicate it is a dropdown, looks like a text input field

Screenshot 2026-05-27 at 14 11 58

6. Options in the top nav does not seem to work
Screenshot 2026-05-27 at 14 12 44

7. if you double click on the price, it opens the selected offer and shows the price tab. If you click on "Back" it looses all the info. The offer is not selected nor it reads the OSI from the fragment (As current behavior in OST). It just gives the blank starting OST window

@npeltier @3ch023 if the decision is to merge this and fix everything in the PR that turns it on, feel free to merge. I'll keep this here as a record not to forget on these in the follow-up

PR #895, opened 2026-05-20
MWPW-195920: ship new Lit OST behind ?ost=new (legacy stays default) + 7 authoring fixes

Axelcureno:
Resolves https://jira.corp.adobe.com/browse/MWPW-195920

Ships the new Lit OST behind a flag (legacy stays the default) and fixes the 7 authoring regressions afmicka flagged on #894. Continuation of merged #894 (MWPW-195919).

## Flag direction (per @3ch023 review)
- Legacy tacocat OST is the **default** in Studio; `?ost=new` / `#ost=new` loads the new Lit OST.
- Both bundles load from the **same deployment** (no cross-origin CDN fetch).
- Legacy bundle (`studio/ost/index.js`) is **kept**, not deleted.

## Build
- New OST is built by Studio's existing esbuild pipeline (`studio/build.mjs`) into `studio/ost/ost-new.js` — the standalone vite `build`+`cp` in `ost/package.json` is removed. One `npm run build`, no second toolchain.

## Bug fixes (new OST → match legacy behavior)
- **Bug 1** — preview applies the context `promotionCode` (new `effectivePromoCode` getter; typed override still wins).
- **Bug 2** — strikethrough preview hides recurrence + VAT labels not shown on a real card.
- **Bug 3** — discount % auto-shows for a promo offer with no manual reference OSI.
- **Bug 4** — legal disclaimer renders (legal type's overrides merged into options).
- **Bug 5** — country selector is an `sp-picker` dropdown (was a text-input-looking field); removed ~105 lines of custom dropdown.
- **Bug 6** — Country/Landscape top-nav options reach the preview.
- **Bug 7** — Back from a deep-linked price preserves the offer/OSI.

## Style
- Selected product row uses a solid grey background instead of the blue left-border accent.

## Verification
- OST unit tests: 639 pass.
- OST bundle rebuilt (`studio/ost/ost-new.js`) from merged source.

## Test URLs
- After: https://mwpw-195920--mas--adobecom.aem.live/ (append `?ost=new` for the new OST)

## Screenshots
- Add before/after for: new-vs-legacy OST, country dropdown, strikethrough, discount %, legal, selected-row style.

## Checklist
- [x] Unit tests cover the changes (639 pass)
- [ ] NALA studio E2E green
- [ ] Demo from Test Page
- [ ] Milo follow-up PR linked


3ch023 (changes requested):
Hi @Axelcureno , i changed description of jira to keep old bundle behind ?ost=legacy flag

it should both in studio and Milo ost.
e.g. by default studio ost and milo ost will have 'new' implementation.
if user adds ?ost=legacy, then studio and milo would load legacy bundle.
we will keep legacy bundle for a few weeks till we are sure there are no blockign issues in new ost.

https://main--milo--adobecom.aem.page/tools/ost?maslibs=mwpw-195920&token=

you will need to get the token to open page above (run `copy(adobeIMS.getAccessToken().token) `in devtools on any acom live pages)
https://git.corp.adobe.com/wcms/tacocat.js/wiki/Testing-OST-Milo

Image

nala/studio/acom/plans/individuals/specs/individuals_save.spec.js
```diff
@@ -133,10 +133,10 @@ export default {
             data: {
                 cardid: 'c72789db-f4c0-4b72-a6ba-3b73b05ae91a',
                 osi: 'r_JXAnlFI7xD6FxWKl2ODvZriLYBoSL701Kd1hRyhe8',
-                legalDisclaimer: 'par licenceAnnual, billed monthly',
-                cardLegalDisclaimer: 'par licence',
-                legalDisclaimerTax: 'par licenceTTC. Annual, billed monthly',
-                cardLegalDisclaimerTax: 'par licenceTTC',
+                legalDisclaimer: 'licenceAnnual, billed monthly',
+                cardLegalDisclaimer: 'Annuel, facturé mensuellement',
+                legalDisclaimerTax: 'licenceHT. Annual, billed monthly',
```
npeltier: is this supposed to be french? :)

npeltier (commented):
i know @3ch023 asks for similarity with old OST but:

Image 
1. that section to me is not part of the selection of offers, but rather context change, i would stick it to the top. Obviously my word is not the final one and we should discuss it.

2. i'm not sure i understand what "Authoring Mode" is supposed to be. Is there any docs / specs i can read around those changes?

3. the fact that we have 240px high templates, all mixed in one pane, is kind of a UX regression to me on smallish screens (well my 16inch laptop at least):
Image
suppose i want a checkout link, i was one click away before, now i need to scroll down which is a bit longer & confusing to me

yesil:
@Axelcureno 

Opening an OSI(price) with a double click is causing several fetch calls for the same resource.
this is pointing a potentially endless loop issue with maybeReapplyDeepLink.

image


Please review and validate that double click on a price opens OST and in the network tab there is only one offer_selectors/ call.

npeltier:
nit: when clicking on help, i get that 2, not sure what the problem is
image
i guess i heard in stand up you wanted to created a doc, you could just point to there

talking about help :) how do i add a second offer in a soft bundle? i added one in soft bundle mode, and couldn't add the other.

thanks for moving options to bottom, if no big deal, moving promo to top as a sticky banner COULD help but it's fine for now.

seanchoi0 (commented):
Tested the `select()` → `#handleOstEvent` flow locally by adding console logs to `ost-app.js` and `rte-field.js`.

**Double-fire confirmed.** When a user picks an offer, `ost-app.select()` dispatches an `ost-select` event before calling `store.onSelect`, so `rte-field.js`'s document-level listener fires twice per selection:

1. First fire — `detail` carries `{osi, type, offer, options, promoOverride, country}` (no `data-wcs-osi`) → handler passes the `ostRteFieldSource` guard and inserts a malformed ProseMirror node
2. Second fire — `onPlaceholderSelect` dispatches with the correct attributes (`data-wcs-osi`, `is`, etc.) → handler inserts the correct node

For inline prices the empty-link cleanup at `rte-field.js:1573` catches the first node (no content, no `data-wcs-osi`), so the end result looks fine. For checkout CTAs with text the first node has content and survives the cleanup, leaving a phantom link in the editor.

A minimal fix would be to remove the `dispatchEvent` call from `ost-app.select()` — `onPlaceholderSelect` already dispatches the correctly-shaped event, so the first one is redundant.

One other thing noticed during the review: `autoResolveOsi` and `autoFillBaseAndTrial` in `ost-store.js` don't have direct tests yet. The existing `loadOffers` tests exit early via `fake-` offer types and never reach those async paths — would be great to add at least branch-level coverage before this lands.

Overall the 7 bug fixes look solid and the new locale/close restore logic is clean — nice work on this one!

seanchoi0 (changes requested):

Axelcureno:
@seanchoi0 thanks — both points verified.

**autoResolveOsi / autoFillBaseAndTrial:** added direct tests (fake-offer short-circuit, AOS resolution, error path; non-tryBuy / not-one-BASE / base-fill / already-filled). Good catch — `loadOffers` bailed early on `fake-` types.

**Double-fire:** confirmed both dispatch `ost-select`, so `#handleOstEvent` fires twice. But I could not reproduce a surviving phantom node on this branch: fire #1 has `attributes.is` undefined → `content` is `null` (rte-field.js:1547) → empty link the cleanup (rte-field.js:1573) removes, even for checkout CTAs with a range selection. A guard no-oping the malformed fire passes with or without it, so I held off shipping a non-discriminating fix. Could you share the exact repro/revision where the phantom survived? If there is a selection/marks state where fire #1 leaks, I will add a failing test + guard.

seanchoi0 (commented):
You're right — I missed that `content` is gated on `attributes.is === CUSTOM_ELEMENT_CHECKOUT_LINK` (line 1540), so fire #1 always produces `content = null` regardless of `ctaText`, and the cleanup catches it. My phantom-survival claim was wrong. The double-fire is real but effectively contained by that guard. No repro on my side — thanks for tracing it through carefully.

seanchoi0 (approved):
looks good

3ch023 (changes requested):
hi @Axelcureno , 

there is a mas defaults regression
open [current OST](https://main--mas--adobecom.aem.live/studio.html#fragmentId=b8934796-68a0-41a9-b273-ed7175db2bfe&locale=de_DE&page=fragment-editor&path=sandbox) and for a price in DE you will see tax label
on [PR branch it is missing](https://mwpw-195920--mas--adobecom.aem.live/studio.html#fragmentId=b8934796-68a0-41a9-b273-ed7175db2bfe&locale=de_DE&page=fragment-editor&path=sandbox), for both new and legacy.

it seems to happen also for legacy ost, not sure how that one got impacted

Image

Axelcureno:
> hi @Axelcureno ,
> 
> there is a mas defaults regression open [current OST](https://main--mas--adobecom.aem.live/studio.html#fragmentId=b8934796-68a0-41a9-b273-ed7175db2bfe&locale=de_DE&page=fragment-editor&path=sandbox) and for a price in DE you will see tax label on [PR branch it is missing](https://mwpw-195920--mas--adobecom.aem.live/studio.html#fragmentId=b8934796-68a0-41a9-b273-ed7175db2bfe&locale=de_DE&page=fragment-editor&path=sandbox), for both new and legacy.
> 
> it seems to happen also for legacy ost, not sure how that one got impacted
> 
> Image

Good catch @3ch023 this is addressed: https://mwpw-195920--mas--adobecom.aem.live/studio.html#fragmentId=b8934796-68a0-41a9-b273-ed7175db2bfe&locale=de_DE&page=fragment-editor&path=sandbox (added 4 unit tests to catch a potential regression in the future)

3ch023 (changes requested):
to wrap up before shut down, maybe we can flip back to legacy ost by default and merge this PR and continue post shut down with fixed. wdyt @Axelcureno ?

feel free to flip default, dismiss my requested changes and ask @Roycethan to run regression on Studio & Milo ost and we can merge it


Findings:
1) when i search for '040DCF1A11122A9096E4756473D186AF' filter are set to wrong values in new ost

Image

2) on AU, 'incl. GST' is missing from legal disclaimer
https://mwpw-195920--mas--adobecom.aem.live/studio.html#fragmentId=c08dd1e7-86d8-419e-bdef-c8000567b6a0&locale=en_GB&region=en_AU&page=fragment-editor&path=sandbox&ost=legacy

Axelcureno:
> 1. 040DCF1A11122A9096E4756473D186AF

Both items are addressed @3ch023. We can proceed with QA and merging. @Roycethan.

nala/studio/acom/plans/individuals/specs/individuals_edit_and_discard.spec.js
```diff
@@ -403,7 +403,7 @@ export default {
                     original: 'yIcVsmjmQCHKQ-TvUJxH3-kop4ifvwoMBBzVg3qfaTg',
                     updated: 'r_JXAnlFI7xD6FxWKl2ODvZriLYBoSL701Kd1hRyhe8',
                 },
-                legalDisclaimer: 'per licenseAnnual, billed monthly',
+                legalDisclaimer: 'per license',
```
3ch023: @Axelcureno here 'per licenseAnnual, billed monthly' needs to be valdiated.
if it would be just 'per license' then it means you have a bug.
OST works correctly, test is passing because 'per licence' is there. but you need to revert to check for the whole thing, including 'Annual, billed monthly' to be there.

eslint.config.mjs
```diff
@@ -12,7 +12,6 @@ const ignoresConfig = {
         '**/libs/**',
         '**/dist/**',
         'web-components/docs/**',
-        'studio/ost/index.js',
```
3ch023, line 15: still needed

3ch023 (approved):

afmicka:
@Axelcureno regression-wise this looks good to me. Legacy OST should not be broken. 
The New OST has issues and those will list separately in the follow-up jiras. We will decide what needs to be fixed before the switch (if anything). 
cc: @3ch023

afmicka (approved):

PR #906, opened 2026-05-26
MWPW-192042: m@s Express: Badge value can not be removed in IN grouped variation

mirafedas:
The 'badge' field in a variation can now persist an intentional empty override using the value `explicit_empty`, so clearing inherited content (parent badge text) no longer falls back to "inherit parent" after save/reload.

The following applies to the badge field only:
- When the parent fragment has text, and the user clears the badge field on the variation, instead of saving [''] we save `['explicit_empty']` (see `prepareVariationForSave` in `fragment.js`);
- When we reload the merch-fragment-editor, we can now distinguish that the badge has been overwritten by the user to store an empty value, because we receive the `['explicit_empty']` in the badge field.
- To show an empty text instead of 'explicit_empty' string, we replace the 'explicit_empty' string with an empty string in three places: `customizeTree` in `customize.js`, `getResolvedFragment` in `preview-fragment-store.js`, and in `transformAuthorData` in `aem-fragment.js`.
- In `simplified-pricing-express` and `full-pricing-express` templates, we now hide the badge slot if it contains no text, even if the border color is selected.

This PR is scoped to the *badge field only*, but the code is written to make it easier to extend this to other variation fields. Here is the Jira for the rest of the fields: https://jira.corp.adobe.com/browse/MWPW-197427 

Resolves https://jira.corp.adobe.com/browse/MWPW-192042

## 🧪 Nala E2E Tests

### To run Nala tests again:

- Remove the `run nala` label

## Test URLs:

Before:
- Parent: https://main--mas--adobecom.aem.live/studio.html#fragmentId=2714e130-9263-47b1-83b7-a05f75d0d2ec&page=fragment-editor&path=sandbox
- Variation: https://main--mas--adobecom.aem.live/studio.html#fragmentId=dbeb9200-f31c-41d6-b7b1-f2af9ba6a483&page=fragment-editor&path=sandbox

After: 
- Parent: https://mwpw-192042--mas--adobecom.aem.live/studio.html#fragmentId=2714e130-9263-47b1-83b7-a05f75d0d2ec&page=fragment-editor&path=sandbox
- Variation: https://mwpw-192042--mas--adobecom.aem.live/studio.html#fragmentId=dbeb9200-f31c-41d6-b7b1-f2af9ba6a483&page=fragment-editor&path=sandbox


npeltier (commented):
claude has a concern :)   

Critical concern: explicitEmpty round-trip is unverified

  This is the core design risk and I can't find a test that addresses it.

  The flag is a client-only invention. saveFragment in aem.js:236 does:

  const fieldsWithType = fields.map((field) => ({ ...field, type: field.type || 'text' }));

  That spreads explicitEmpty into the PUT body. Whether AEM persists it depends on Sites CF API behavior with unknown field properties. Three possibilities:

  1. AEM strips unknown properties. On reload, field comes back as {values: [''], ...} without explicitEmpty → reverts to "inherit from parent" → the user's clear is silently undone on next session.
  2. AEM rejects unknown properties → save fails.
  3. AEM stores it. Round-trip works.

what is suggested: 
 Concretely:
  1. Manual test: clear a promo code on a variation, save, hard-reload, confirm it's still cleared.
  2. If that fails, fall back to a sidecar-field design or coordinate with Odin team.
  3. Tighten the badge-render condition to avoid the unintended parent-card behavior change.

3ch023 (changes requested):
@mirafedas

1.  in description it says 'Persist [''] with explicitEmpty: true on the field', do you see this to be the case? on test fragment i see empty badge HTML
2. variation preview i can still see 'space' for badge, even though text is empty. Likely related to reason above.
I assume the desired view without the badge would be similar to cards on https://www.adobe.com/express/ (see 2nd screenshot)

Image

Image

studio/src/aem/fragment.js
```diff
@@ -202,6 +242,19 @@ export class Fragment {
...
+        const ownValues = ownField?.values || [];
+        const parentValues = parentFragment?.getFieldValues(fieldName) || [];
+        const isMultipleField = ownField?.multiple === true || parentFragment?.getField(fieldName)?.multiple === true;
+
+        if (
+            isVariation &&
+            parentFragment &&
+            isExplicitSingleValueEmptyOverride(ownField, ownValues, parentValues, isMultipleField)
+        ) {
+            return '';
```
3ch023, line 302: tried to debug it and i couldn't reach this branch

npeltier:
@mirafedas what you could do is manage a value for string / string array that set explicit empty. 'explicit_empty' or something. 
We "just" need to do a change on io side too that if value is that string / contains that string, then be empty

io/www/src/fragment/utils/explicit-empty.js
```diff
@@ -0,0 +1,154 @@
+/**
```
npeltier, line 1: i wonder if we shouldn't rather move the whole merge feature here, and call it merge-util.
Also please consider / sync with @yesil 's PR comment here https://github.com/adobecom/mas/pull/846#discussion_r3342746151 where new way of merging is there.

I guess the explicit_empty config could be in that MERGE_CONFIG object too

studio/src/reactivity/preview-fragment-store.js
```diff
@@ -3,12 +3,21 @@ import { FragmentStore } from './fragment-store.js';
...
+} from '../../../io/www/src/fragment/utils/explicit-empty.js';
 
 export const INHERITED_SETTINGS_FIELDS = new Set(['addon', 'showPlanType', 'showSecureLabel', 'quantitySelect']);
 
 export function serializePreviewFields(fields = []) {
     return fields.reduce((result, field) => {
         const values = field.values || [];
+
+        if (fieldValuesArePersistedExplicitEmpty(values)) {
+            return result;
```
Axelcureno: **Blocker (preview-only):** for a persisted `explicit_empty` this `return result` **drops the field from the preview payload** rather than sending an empty value. With no badge key on the variation, IO's deepMerge inherits the parent badge, and `mergeResolvedPreviewFields` then overwrites the original `['explicit_empty']` with the parent's value — so a cleared badge still shows the parent badge in Studio preview. (Published output is fine; this is preview fidelity only.)

Fix: for `FIELDS_ALLOWING_EXPLICIT_EMPTY`, emit an explicit empty instead of returning early, e.g. `result[field.name] = field.multiple ? [] : ''` so the resolver gets an empty override rather than "no opinion → inherit".

studio/src/aem/fragment.js
```diff
@@ -202,7 +263,18 @@ export class Fragment {
...
+        const ownField = this.getField(fieldName);
+        const ownValues = ownField?.values || [];
+        const parentValues = parentFragment?.getFieldValues(fieldName) || [];
+        const isMultipleField = ownField?.multiple === true || parentFragment?.getField(fieldName)?.multiple === true;
+
+        if (isVariation && parentFragment && isExplicitEmptyVariationValue(ownValues, parentValues, isMultipleField)) {
+            return '';
+        }
         const ownValue = this.getFieldValue(fieldName, index);
+        if (isExplicitEmptySentinel(ownValue)) {
```
Axelcureno: This guard fires for **any** single-value field whose stored value is literally `"explicit_empty"`, not just `badge` — the write path is badge-scoped but this read path isn't. Low real-world odds, but worth scoping with `fieldAllowsExplicitEmpty(fieldName)` to match. Same applies to the `getFieldState` branch below (~line 344).

studio/src/aem/fragment.js
```diff
@@ -228,14 +300,23 @@ export class Fragment {
         // For [""] (single empty string):
         // - For multi-value fields (multiple: true): explicit clear sentinel → return empty array
         // - For single-value fields (multiple: false): AEM initializes empty fields this way → inherit from parent
+        if (fieldValuesArePersistedExplicitEmpty(ownValues)) {
```
Axelcureno: Coverage: the `getEffectiveFieldValues`/`getFieldState` test tables have no `EXPLICIT_EMPTY_SENTINEL` row, so nothing asserts these getters return `''` rather than the literal `"explicit_empty"` string to the author — the single most important user-facing guarantee of the feature. One sentinel row in each table would close it. The util-layer tests are otherwise thorough.

Axelcureno (changes requested):
Nice work on the sentinel design — defined once in `explicit-empty.js` and imported everywhere (no string drift), and I confirmed the cross-package import bundles correctly into `dist/mas.js` via esbuild, so no runtime breakage. The published/hydrate path is solid: sentinel → `''` → `appendSlot` skips the empty slot → express variants return `nothing`, so the old "badge reserves a blank space" issue is genuinely fixed for rendered cards.

Requesting changes on one real bug (preview-only) plus two smaller things — details inline:

- **[blocker] Preview drops the cleared badge instead of sending `''`** → Studio preview inherits the parent badge. This is what @3ch023 saw on May 27, and it's still live. (The PR description says serialize "replaces explicit_empty with an empty string" — it actually deletes the field.)
- **[medium]** Read-path sentinel guards in `fragment.js` aren't scoped to `badge`.
- **[medium]** Coverage gap: nothing asserts the display getters return `''` rather than the literal `"explicit_empty"`.

Note: @npeltier and @3ch023's reviews predate the sentinel commits (Jun 2–3), so some of that is already addressed — but the preview drop below is not.

io/www/src/fragment/utils/odinSchemaTransform.js
npeltier, line 19: there should not be any change needed here any more. We should actually remove this, now we are not using odin content directly anymore

io/www/src/fragment/utils/explicit-empty.js
```diff
@@ -0,0 +1,154 @@
...
+
+        if (field.multiple === true && values.some(isExplicitEmptySentinel)) {
+            return {
+                ...field,
+                values: values.map((value) => (isExplicitEmptySentinel(value) ? '' : value)),
+            };
+        }
+
+        return field;
+    });
```
npeltier: there should not be any reference to authoring in here. This is mas/io main pipeline, we happen to reuse for previewing content

npeltier (changes requested):
i think there should be no authoring specific code in io/www, which should be slimmed down.
please remove odinSchemaTransform altogether.

also check Ilyas' soon to be merged compchart, there is mergeConfig we should reuse here as well.

io/www/src/fragment/utils/explicit-empty.js
mirafedas: I propose to address this in the scope of the [follow-up Jira](https://jira.corp.adobe.com/browse/MWPW-197427), since Ilyas's PR is not merged yet, and I also want to avoid this PR getting bigger

afmicka: @mirafedas  Ilays's comp-chart PR has been merged already, few hours ago. Please check on conflicts too

studio/src/aem/fragment.js
mirafedas: Here is how to reach this branch:

1. Create a parent and set a badge text
2. Create a variation and make the badge text empty
3. Reload the variation in the editor

io/www/src/fragment/utils/odinSchemaTransform.js
mirafedas: Ok, I removed my changes from odinSchemaTransform.js 👍

io/www/src/fragment/utils/explicit-empty.js
mirafedas: I made this function universal (it works with both author and publish data formats)

studio/src/aem/fragment.js
```diff
@@ -142,20 +182,28 @@ export class Fragment {
...
+                encodedValues = encodedValues.map((v) => (isExplicitEmptySentinel(v) ? '' : v));
+            }
+        }
 
-        // [''] is the explicit clear sentinel for multi-value fields
         const isSingleEmptyString = encodedValues.length === 1 && encodedValues[0] === '';
-
-        // Determine if this is a multi-value field based on the .multiple property
-        const isMultiple = parentField?.multiple === true || existingField?.multiple === true;
+        const allowVariationInheritEmptyWrite =
```
npeltier, line 229: **[CRITICAL]** `allowVariationInheritEmptyWrite` is computed from `isSingleEmptyString`/`parentValuesHaveContent` **without** an `isExplicitEmptyField(fieldName)` guard, unlike every other new branch in this function. Because `resolveVariationStoredValues` already converts `['']` → `[EXPLICIT_EMPTY_SENTINEL]` for badge *before* this line runs, the flag ends up **always false for badge** (dead for the field it appears to be built for) and **always live for every other single-value field** (title, description, subtitle, promoText, etc.).

Reproduced directly against the built `Fragment` class:
```js
// main (before PR): correctly a no-op
variation.updateField('description', [''], parent) // → false, hasChanges: false, values: []

// this PR (after): silently persists an override
variation.updateField('description', [''], parent) // → true, hasChanges: true, values: ['']
```
(parent has content, variation's own `description` field entry exists with `values: []` — the exact combination the comment above this code calls out as "RTE initialization sends [''] for empty fields", i.e. this fires on ordinary load of any untouched inheritable single-value field on a variation, not an exotic edge case.)

Effect: spurious "unsaved changes" state and a persisted `['']` override written to unrelated fields on save, even though this PR's description and tests scope the feature to badge only. All 87 studio + 88 io/www unit tests still pass because this exact combination (existingField present, empty values, non-badge field, parent with content) has no test coverage — confirmed by re-running the full affected suites against this branch in an isolated worktree.

**Fix:** gate this the same way every other new branch here is gated: `isExplicitEmptyField(fieldName) && ...`. At that point it becomes dead code (since `resolveVariationStoredValues` already handles badge via the sentinel earlier in the function), so it may be simplest to just delete it rather than gate it — but gating is the minimal safe fix. Please add a test for `updateField(<non-badge field>, [''], parentWithContent)` where the variation's existing field has `values: []`, mirroring the existing RTE-initialization test but with a `parentFragment` passed in.

*Reviewed with ❤️ by [Claude Code](https://claude.ai) 🤖 and [Scout](https://github.com/Adobe-AIFoundations/scout)*

npeltier (changes requested):
## PR Review: MWPW-192042 — Badge value can not be removed in IN grouped variation

### Summary
Introduces an `explicit_empty` sentinel (`<explicit_empty/>`) so a variation can persist an intentional "clear the badge" override instead of falling back to the parent's badge text. Scope is declared as badge-only. History is effectively one substantive commit (`f0c049845`) plus 10 no-op merge commits from `main`. All prior reviewers' blockers (the preview-drop blocker, the read-path scoping gap, the missing sentinel test rows, the `odinSchemaTransform.js` removal request) are already addressed in the current head. Found one new regression that was missed: an unscoped guard that silently corrupts dirty-state tracking for *any* single-value variation field, not just badge (see inline comment).

### Change Surface
- 23 files, 45 symbols changed. Scope matches the description (badge-only) for the sentinel-allowed-fields list, except the regression flagged inline, which is unscoped.

### Blast Radius
- `Fragment.updateField`/`getEffectiveFieldValue`/`getFieldState`/`prepareVariationForSave`: all callers reviewed, contracts preserved.
- `normalizeExplicitEmptyInFields`: 3 call sites (aem-fragment.js, preview-fragment-store.js, customize.js), all consistent and idempotent for non-sentinel data.
- No dangling references; no new dead code introduced by this PR.

### Test Gaps
- Missing regression test: `updateField(nonBadgeFieldName, [''], parentWithContent)` where the variation's existing field entry has `values: []`. A test mirroring `fragment.test.js` "does not mark changes for single-value field going from [] to [\"\"]" but *with* a `parentFragment` argument would catch this. All other explicit-empty paths are well covered.

### Affected-Test Run
Ran the directly affected suites in an isolated worktree: `studio/test/aem/fragment.test.js`, `studio/test/reactivity/preview-fragment-store.test.js`, `studio/test/reactivity/create-preview-data-with-parent.test.js` → 87 passing. `io/www/test/explicit-empty.test.js`, `io/www/test/fragment/customize.test.js` → 88 passing. ESLint on all changed source files: clean. Did not run the full web-components affected-tests fan-out; nothing there touches the flagged code path.

### Security
No issues found. Pure client-side state/merge logic; no new external-input trust boundary.

### Verdict
**Request changes** — one CRITICAL finding below is a silent, easily-triggered regression outside the PR's stated scope (badge-only) that the existing suite doesn't catch. One-line gate fix + one test, then good to merge.

---
*Reviewed with ❤️ by [Claude Code](https://claude.ai) 🤖 and [Scout](https://github.com/Adobe-AIFoundations/scout)*

npeltier (commented):
Follow-up to my earlier review: proposed failing unit test for the CRITICAL finding.

studio/test/aem/fragment.test.js
```diff
@@ -362,6 +429,28 @@ describe('Fragment', () => {
...
+                const parent = new Fragment(
+                    createFragmentConfig({ fields: [{ name: 'badge', values: ['<merch-badge>Sale</merch-badge>'] }] }),
+                );
+                const variation = new Fragment(
+                    createFragmentConfig({ fields: [{ name: 'badge', values: [EXPLICIT_EMPTY_SENTINEL] }] }),
+                );
+
+                const prepared = variation.prepareVariationForSave(parent);
+                expect(prepared.getFieldValues('badge')).to.deep.equal([EXPLICIT_EMPTY_SENTINEL]);
+            });
```
npeltier, line 453: Proposed regression test for the CRITICAL finding above — fails on this branch (`result: true`, `hasChanges: true`, `values: ['']`, all wrong), passes on `main`, and would pass again once `allowVariationInheritEmptyWrite` is gated by `isExplicitEmptyField(fieldName)`.

```suggestion
            });

            it('does not mark changes for a non-badge single-value field going from [] to [""] when parent has content', () => {
                // Regression test: the badge-scoped explicit-empty feature must not change
                // behavior for fields outside EXPLICIT_EMPTY_ALLOWED_FIELDS. This mirrors
                // the RTE-initialization no-op test above, but with a parentFragment whose
                // field has content — the combination that exposes the bug.
                const parent = new Fragment(
                    createFragmentConfig({ fields: [{ name: 'description', values: ['Parent description'], multiple: false }] }),
                );
                const variation = new Fragment(
                    createFragmentConfig({ fields: [{ name: 'description', values: [], multiple: false }] }),
                );

                expect(variation.updateField('description', [''], parent)).to.be.false;
                expect(variation.getFieldValues('description')).to.deep.equal([]);
                expect(variation.hasChanges).to.be.false;
            });
```

*Reviewed with ❤️ by [Claude Code](https://claude.ai) 🤖 and [Scout](https://github.com/Adobe-AIFoundations/scout)*

npeltier (approved):
thanks! lgtm

Axelcureno (commented):
Re-reviewed — my three June findings are all addressed, and @npeltier's CRITICAL is genuinely fixed: I pulled the guard out locally and his regression test failed exactly as predicted, then passed on restore.

The preview path I blocked on checks out too — the sentinel is now a defined value, so deepMerge overwrites the parent badge instead of inheriting. `<explicit_empty/>` beats the bare string.

Green locally: 95 IO, 67 studio. Leaving @npeltier's architectural comments to him.

Axelcureno (approved):

afmicka:
@mirafedas these changes are affecting card alignment for Simplified cards when one of the cards does not gave a badge

prod: https://main--da-express-milo--adobecom.aem.page/express/pre-launch/mas/homepage-poc?tab=2
Screenshot 2026-08-13 at 14 23 16

your branch: https://main--da-express-milo--adobecom.aem.page/express/pre-launch/mas/homepage-poc?maslibs=mwpw-192042&tab=2

Screenshot 2026-08-13 at 14 31 27


Note that full pricing is not affected:
- https://main--da-express-milo--adobecom.aem.page/express/pre-launch/mas/pricing-poc-2?tab=3
- https://main--da-express-milo--adobecom.aem.page/express/pre-launch/mas/pricing-poc-2?%3Fmaslibs=mwpw-192042

mirafedas:
> @mirafedas these changes are affecting card alignment for Simplified cards when one of the cards does not gave a badge
> 
@afmicka this change was requested by @3ch023 in this comment: https://github.com/adobecom/mas/pull/906#pullrequestreview-4371136396

afmicka:
@mirafedas your newest changes are bringing back the original issue. Badge is not being removed from variation in your PR description example

https://mwpw-192042--mas--adobecom.aem.live/studio.html#fragmentId=dbeb9200-f31c-41d6-b7b1-f2af9ba6a483&page=fragment-editor&path=sandbox

Screenshot 2026-08-17 at 14 15 57

afmicka (approved):

PR #909, opened 2026-05-28
MWPW-196392: Allow bulk publish for placeholder-only projects with target locales

Axelcureno:
Publish button was permanently disabled on placeholder-only bulk-publish projects because `allAlreadyPublished` checked source en_US status while the publish actually targets configured locales.

- `allAlreadyPublished` now returns `false` when `this.locales.length > 0` — per-locale state is judged server-side by the `/bulk-publish` resolver.
- Unit tests cover both branches (locales configured, locales empty).

## Additional fixes bundled in

- **Publish now enables on existing projects with unsaved changes.** Editing a field (e.g. the title) previously left Publish disabled until a manual save. `publishBlockedReason` only returns UNSAVED for *new* projects now; `publish()` auto-saves first when `hasChanges` is true so the project record matches what gets published.
- **Bulk publish locale cap raised from 50 → 200** in `io/studio/src/bulk-publish/index.js`. The system supports 69 country codes today, and multi-language variants (fr_FR/fr_CA/fr_BE etc.) push the realistic max well past 50. `MAX_RESOLVED` (5000) still bounds total work.

Resolves https://jira.corp.adobe.com/browse/MWPW-196392

- [ ] C2. Add a Nala test
- [x] C3. Verify all Checks are green
- [x] C4. Test Page link in PR
- [ ] C5: Demo ready
- [x] C.6 JIRA ACs validated

Test URLs:

- After: https://mwpw-196392--mas--adobecom.aem.live/

Repro (original bug): `?bulkPublishProjectId=56523407-d833-4e2d-869c-3e450ca61ce1&page=bulkPublishEditor&path=acom`


Axelcureno:
`deploy / deploy` is red because my GitHub user (`Axelcureno`) doesn't have `AIO_STUDIO_{AIO,NS,ENV}_Axelcureno` secrets configured in the repo. Not a code issue.

- Manually deployed `bulk-publish` to my dev workspace (`14257-masstudio-axel`) — successful
- Verified end-to-end: published all locales against prod AEM data via `?io.studio.env=axel`, no longer hits the 50-locale cap
- Other CI failures (Nala E2E) are the same prep-apt install-step flake we've seen on other PRs today; reruns kicked off

yesil (approved):

3ch023 (dismissed):
Hi @Axelcureno ,

could you explain what the fix below is for?
>>>
Publish now enables on existing projects with unsaved changes. Editing a field (e.g. the title) previously left Publish disabled until a manual save. publishBlockedReason only returns UNSAVED for new projects now; publish() auto-saves first when hasChanges is true so the project record matches what gets published.

Axelcureno:
Hi @3ch023 — it's a UX change tied to this PR's flow.

- Placeholder-only / target-locale projects are typically edited (locales, items) and published in one sitting. The old gate disabled Publish on any unsaved change, forcing a manual Save first.
- publishBlockedReason now only blocks isNewProject (no id yet). For existing projects, publish() auto-saves when hasChanges, then proceeds only if the save succeeded (hasChanges cleared) — same save-first guarantee, one click instead of two.

Andrei4226 (approved):
Looks good to me

afmicka:
@3ch023 could you please check if your change request is still valid? 
I assume you want to keep the obligatory saving the project upon change before publishing (thats how translation works too)?

3ch023:
> * The old gate disabled Publish on any unsaved change, forcing a manual Save first.

Hi @Axelcureno ,
>The old gate disabled Publish on any unsaved change, forcing a manual Save first.

I think it is correct.
Imagine a situation, someone accidentally making a change & triggering publish, not saving the project.
As another content author or dev investigating, you woudln't be able to understand how did it happen that bulk project published a card that is not in list of buk project items.

It is expected that changes are saved first, before they can be counted as a part of bulk project.

Axelcureno:
> @3ch023 — the save-first invariant is preserved, just enforced inside `publish()` instead of via the disabled button.

`publish()` auto-saves when there are unsaved changes and only proceeds if the save succeeded (`hasChanges` cleared). A card can never be published without first being written into the saved project's `fragments` list — same guarantee as before.

We moved the gate out of the button because the old `hasChanges` disable-check left Publish stuck disabled after saving (reactive race). Net: same audit guarantee, one click instead of two.

afmicka:
@Axelcureno when i create a porject with placeholder that is published in US but not in FR, and in the project i select only fr_FR locale, i get project published, and the placeholder is published but i get red toast saying content could not have been saved. See screenshot 
Screenshot 2026-06-09 at 12 48 27


@3ch023 this is not blocking for me, if persist we can resolve in a separate jira.

afmicka (approved):

PR #917, opened 2026-05-29
MWPW-181856: Add promo variations

Andrei4226:
Resolves https://jira.corp.adobe.com/browse/MWPW-181856

- "Create promo variation" button in the `⋯` menu from the promotion project
- "Promotion" tab on the default fragment lists all promo variations
- Editor opens the promo variation when navigating from a promotion project
- Publish reminder if promo variations are still unpublished

## 🧪 Nala E2E Tests

### To run Nala tests again:

- Remove the `run nala` label

## Test URLs:

- After: https://mwpw-181856--mas--adobecom.aem.live/


npeltier (changes requested):
hi Andrei! looks great! 
tell me if i'm wrong but it does not seem like you have the variation editor support for promotion variation, where you get "override" link when things have been overriden?

Andrei4226:
> tell me if i'm wrong but it does not seem like you have the variation editor support for promotion variation, where you get "override" link when things have been overriden?

Could you clarify again please? Are you referring to the "Overridden. Click to restore" link in the page editor or a separate "Override content" action in the dots menu?
Promo variations reuse the same editor as regional variations: when you change a field from the parent, you get "Overridden. Click to restore." in the form.
Btw: it's not the final version yet, I want to check one more time some things:D

studio/src/aem/fragment.js
```diff
@@ -345,6 +346,15 @@ export class Fragment {
         return path?.includes(`/${PZN_FOLDER}/`) ?? false;
     }
 
+    /**
+     * Checks whether a path is a promo variation under promotions/{promoName}/.
+     * @param {string} path
+     * @returns {boolean}
+     */
+    static isPromoVariationPath(path) {
```
yesil: is this really needed?

studio/src/promotions/mas-promotions-items-table.js
```diff
@@ -326,14 +354,36 @@ class MasPromotionsItemsTable extends LitElement {
...
+        const existing = new Set();
+        await Promise.all(
+            items.map(async (item) => {
+                const targetPath = buildPromoVariationPathForTag(item.path, promoTag);
+                if (!targetPath) return;
+                const variation = await this.repository.aem.sites.cf.fragments.getByPath(targetPath).catch(() => null);
+                if (variation?.id) existing.add(item.path);
+            }),
+        );
+        this.#existingPromoVariationDefaultPaths = existing;
```
yesil: **[Blocker — race]** This write (and the `requestUpdate()` below) is unconditional, but the caller in `#loadSelected` is abort-guarded. The N-wide `getByPath` fan-out above isn't.

If the selection changes while these are in flight, a superseded run still overwrites `#existingPromoVariationDefaultPaths` and re-renders, leaving stale Create/View menu state for the current selection. Thread the `signal` in and `if (signal.aborted) return;` before the write, mirroring `#loadSelected`.

studio/src/promotions/mas-promotions-items-table.js
```diff
@@ -242,20 +256,34 @@ class MasPromotionsItemsTable extends LitElement {
...
         super();
         this.viewOnlyLoading = false;
         this.viewOnlyFragments = [];
+        this.confirmDialogConfig = null;
+        this.createPromoVariationLoading = false;
         this.getDisplayName = (fragmentData) => fragmentData?.path ?? '';
         this.renderFragmentStatusCell = () => nothing;
     }
 
+    get #promotionTagId() {
```
yesil, line 280: **[Suggestion — reactivity]** Menu visibility (`#canCreatePromoVariation` / `#hasPromoVariationForItem`) flows from `#promotionTagId`, which reads `Store.promotions.inEdit`. But the `ReactiveController` in `connectedCallback` subscribes only to the selection store, not `inEdit`.

If the table mounts before `inEdit` resolves, the Create/View actions won't appear until an unrelated selection change forces a re-render. Please confirm `inEdit` is always populated pre-mount; if not, add it to the controller's store list.

studio/src/promotions/mas-promotions-items-table.js
```diff
@@ -361,12 +411,126 @@ class MasPromotionsItemsTable extends LitElement {
...
+
+        const confirmed = await this.#confirmCreatePromoVariation();
+        if (!confirmed) return;
+
+        try {
+            this.createPromoVariationLoading = true;
+            showToast('Creating promo variation...');
+            const created = await this.repository.createPromoVariation(item.id, promoTag);
+            showToast('Promo variation created', 'positive');
+            this.#existingPromoVariationDefaultPaths = new Set([...this.#existingPromoVariationDefaultPaths, item.path]);
```
yesil: **[Nitpick — reactivity]** `#existingPromoVariationDefaultPaths` is a private field (non-reactive), and unlike `#viewPromoVariation` this reassignment isn't followed by `requestUpdate()`. It works only because navigation immediately unmounts the table. If navigation throws or is short-circuited (discard prompt, router guard), the row keeps showing "Create promo variation" for an item that now has one. Add `this.requestUpdate();` after the reassignment.

studio/src/mas-fragment-editor.js
```diff
@@ -1037,6 +1048,16 @@ export default class MasFragmentEditor extends LitElement {
             return parentData;
         }
 
+        if (Fragment.isPromoVariationPath(fragmentPath)) {
+            parentData = await this.repository.resolveDefaultFragmentForPromoVariation(fragmentPath, this.fragment?.id);
```
yesil: **[Nitpick — stale state]** The success branch below clears `this.groupedVariationOrphanMessage = null`, but the `return null` path (parent unresolved) does not. A grouped-variation orphan warning set earlier can bleed onto a promo-variation editor whose default can't be resolved. Clear the message on the null branch too.

studio/src/mas-promotions.js
```diff
@@ -391,6 +391,16 @@ class MasPromotions extends LitElement {
             showToast('This promotion has ended. Update the dates to publish again.', 'info');
             return;
         }
+        const unpublished = await this.repository.getUnpublishedAttachedPromoVariations(fragment);
```
yesil: **[Blocker — C1 / tests]** This publish-reminder path (unpublished lookup → confirm dialog → early return) has no test: `mas-promotions.js` has no test file, while the *identical* logic in `mas-promotions-editor.js` got a `publish reminder flow` test. The confirm-message string and dialog config are copy-pasted between the two handlers, so the untested copy can drift. Extract one shared helper and cover it.

studio/src/mas-repository.js
```diff
@@ -2282,6 +2302,121 @@ export class MasRepository extends LitElement {
...
+                    (variationPath) => resolveDefaultPathFromPromoVariation(variationPath, promoName) === parentPath,
+                );
+                return { parentPath, promoVariationPaths };
+            }),
+        );
+
+        const variationResults = await Promise.all(
+            parentResults.flatMap(({ parentPath, promoVariationPaths }) =>
+                promoVariationPaths.map(async (variationPath) => {
+                    const variation = await this.aem.sites.cf.fragments.getByPath(variationPath).catch(() => null);
```
yesil: **[Suggestion — perf]** Nested unbounded fan-out on the publish path: one `getByPath` per attached parent, then a second per discovered variation, with no concurrency cap. Grows O(parents + variations) per publish. Consider capping concurrency or batching the lookups.

yesil (dismissed):
please avoid requestUpdate directly. State updates should cause a render properly.

studio/src/aem/fragment.js
Andrei4226: Removed `Fragment.isPromoVariationPath`. Call sites import the util directly.

studio/src/promotions/mas-promotions-items-table.js
Andrei4226: Done.
`#syncExistingPromoVariations(items, signal)` checks `signal.aborted `at entry, in the map callback and before writing `existingPromoVariationDefaultPaths`.

studio/src/promotions/mas-promotions-items-table.js
Andrei4226: Subscribed the table to `Store.promotions.inEdit` so Create/View refresh when the promotion loads, not only when selection changes.

studio/src/promotions/mas-promotions-items-table.js
Andrei4226: Thanks!
Addressed via reactive state: `existingPromoVariationDefaultPaths` is a declared Lit property.
create/sync/view paths assign a new Set so the row updates even if navigation is delayed or cancelled. 
No manual `requestUpdate()` needed.

studio/src/mas-fragment-editor.js
Andrei4226: Done.
The promo branch clears `groupedVariationOrphanMessage` before return null when the parent is unresolved.

studio/src/mas-promotions.js
Andrei4226: Extracted `confirmPublishDespiteUnpublishedPromoVariations` (and related constants) to `studio/src/promotions/promotion-publish-utils.js` with unit tests. 
Both `mas-promotions.js` and `mas-promotions-editor.js` call the shared helper.

studio/src/mas-repository.js
Andrei4226: Good point.
Capped concurrency with existing `processConcurrently` and `VARIATIONS_CONCURRENCY_LIMIT (5)` for both parent and variation `getByPath` calls in `getUnpublishedAttachedPromoVariations`.

studio/src/mas-repository.js
```diff
@@ -2282,6 +2303,127 @@ export class MasRepository extends LitElement {
         return createdFragment;
     }
 
+    /**
```
npeltier: mas-repository bloating is anti pattern:
- can you please move promotions related changes to dedicated promotions model? see https://github.com/adobecom/mas/blob/main/CONTRIBUTING.md#studio-studio
- BONUS: create one too under fragments, and move applyFragmentListFitlers there

studio/src/promotions/promotion-publish-utils.js
```diff
@@ -0,0 +1,29 @@
...
+export async function confirmPublishDespiteUnpublishedPromoVariations(repository, promotionFragment, showDialog) {
+    const unpublished = await repository.getUnpublishedAttachedPromoVariations(promotionFragment);
+    if (!unpublished.length) return true;
+    const message = `This project has ${unpublished.length} attached promo variation(s) that are not published. Publish the project anyway?`;
+    return showDialog(UNPUBLISHED_PROMO_VARIATIONS_DIALOG.title, message, {
+        confirmText: UNPUBLISHED_PROMO_VARIATIONS_DIALOG.confirmText,
+        cancelText: UNPUBLISHED_PROMO_VARIATIONS_DIALOG.cancelText,
+        variant: UNPUBLISHED_PROMO_VARIATIONS_DIALOG.variant,
+    });
+}
```
npeltier: i think we should rather propose to add variations as references of the publication, either not publish at all. So message 
```This project has ${unpublished.length} attached promo variation(s) that are not published. Publish them together with the project?```

File: studio/src/promotions/promo-variation-utils.js [hunk 10]
```diff
@@ -0,0 +1,120 @@
+import { PATH_TOKENS, ROOT_PATH, TAG_PROMOTION_PREFIX } from '../constants.js';
```
npeltier: i would make that whole file the model file tbh, it's not really utils, but one off model method.
promotion-repository.js? promotion-model.js? promotions.js ? 
your call

studio/src/promotions/promo-variation-utils.js
```diff
@@ -0,0 +1,120 @@
...
+
+/**
+ * Extracts promo name from a promo variation path using its mas:promotion/ tag when available.
+ * @param {string} promoPath
+ * @param {string} [promoTagId]
+ * @returns {string|null}
+ */
+export function getPromoNameFromPromoVariationPath(promoPath, promoTagId) {
+    const fromTag = promoTagId ? getPromoNameFromTag(promoTagId) : null;
+    if (fromTag) return fromTag;
```
npeltier: this part is a bit confusing, the method is called getPromoNameFromPromoVariationPath and should just take path as an entry, in caller code, either you have a tag and you use it, either not and call this method

npeltier (changes requested):
we should not be using `variations` field for promo variations we have already all that we need with promo tag (project bind) + fragment path (fragment bind)

studio/src/mas-repository.js
Andrei4226: Done.
Extracted `createPromoVariation`, `getUnpublishedAttachedPromoVariations` and `resolveDefaultFragmentForPromoVariation` into `studio/src/promotions/promotion-variations.js`.
Fragment list filtering moved to `studio/src/fragments/fragment-list-filters.js`.

studio/src/promotions/promotion-publish-utils.js
Andrei4226: Updated: confirm dialog + Publish together publishes project + unpublished promo variations in one call.

studio/src/promotions/promo-variation-utils.js
Andrei4226: Renamed it to `promotion-model.js`.

studio/src/promotions/promo-variation-utils.js
Andrei4226: Fixed, `getPromoNameFromPromoVariationPath` is path only now.

studio/src/promotions/promotion-variations.js
```diff
@@ -0,0 +1,179 @@
...
+    if (!attachedPaths.length) return [];
+
+    const results = await processConcurrently(
+        attachedPaths,
+        async (parentPath) => {
+            const variationPath = buildPromoVariationPathForTag(parentPath, promotionTagId);
+            if (!variationPath) return null;
+            const variation = await aem.sites.cf.fragments.getByPath(variationPath).catch(() => null);
+            if (!variation) return null;
+            if (variation.status === STATUS_PUBLISHED || variation.status === STATUS_MODIFIED) return null;
```
Axelcureno: 🔴 blocker — `STATUS_MODIFIED` shouldn't be excluded here. MODIFIED means the variation was published once but has newer unpublished edits, so treating it as "already published" means Publish-together ships the stale last-published copy and the author's latest edits never go live. And since promo variations aren't AEM references of the project, reference-traversal won't catch them either. Should be `=== STATUS_PUBLISHED` only.

studio/src/mas-repository.js
```diff
@@ -2282,6 +2279,68 @@ export class MasRepository extends LitElement {
...
+            if (parent) {
+                const enriched = promotionVariations.mergePromoVariationReferences(parent, [createdFragment]);
+                parentStore.refreshFrom(enriched);
+                this.#addToCache(parentStore.get());
+            }
+        }
+        return createdFragment;
+    }
+
+    #getPromotionProjectsForProbe() {
```
Axelcureno: Two things here.

🔴 **blocker:** this reads `Store.promotions.list.data` synchronously, but `loadPromotions()` only runs on the Content/Promotions pages — never the editor. On a deep-link or hard reload straight into the editor the list is empty, the probe returns `[]`, and the promo-variations tab shows an authoritative "No promotion variations found" with no self-heal. Awaiting `loadPromotions()` in the merge when the list is empty would fix it.

**SoC** (tying into @npeltier's note): this helper plus `createPromoVariation` / `resolveDefaultFragmentForPromoVariation` / `getUnpublishedAttachedPromoVariations` are promotions-specific and per CONTRIBUTING.md belong in a promotions model, not the shared repository.

studio/src/promotions/promotion-publish-utils.js
```diff
@@ -0,0 +1,83 @@
...
+            const fragments = [promotionWithEtag];
+            for (const path of promoVariationPaths) {
+                const variation = await repository.aem.sites.cf.fragments.getByPath(path).catch(() => null);
+                if (!variation?.id) continue;
+                const variationWithEtag = await repository.aem.sites.cf.fragments.getWithEtag(variation.id);
+                if (variationWithEtag) fragments.push(variationWithEtag);
+            }
+            await repository.aem.sites.cf.fragments.publishFragments(fragments, publishReferencesWithStatus);
+        }
+        showToast(PROMOTION_PUBLISH_SUCCESS_MESSAGE, 'positive');
```
Axelcureno, line 92: 🟡 should-fix — the loop above skips any variation whose `getByPath`/`getWithEtag` comes back falsy (`continue` / no push), then this success toast fires unconditionally. If the user confirmed "publish N variations" and only M < N actually got submitted, they get a green toast with no indication some were dropped. Worth comparing `fragments.length` against `promoVariationPaths.length + 1` and warning on a shortfall.

studio/src/promotions/mas-promotions-items-table.js
```diff
@@ -252,14 +268,29 @@ class MasPromotionsItemsTable extends LitElement {
...
+        const { promotion: promotionTags } = splitPromotionTagsFieldValues(promotion.getFieldValues('tags'));
+        const first = promotionTags[0];
+        return first ? normalizeTagId(first) : null;
+    }
+
     connectedCallback() {
         super.connectedCallback();
         const upper = this.typeUppercased;
-        this.#selectionController = new ReactiveController(this, [getItemsSelectionStore()[`selected${upper}`]]);
+        this.#selectionController = new ReactiveController(this, [
```
Axelcureno, line 293: 🟡 should-fix — controller is created in `connectedCallback` but `disconnectedCallback` never removes it. The Spectrum overlay reparents this table on each Edit-items open/close, so every reopen registers another controller — N opens = N live store subscriptions each firing their own `requestUpdate`. Either construct it once in the constructor, or `removeController` + null it on disconnect and recreate only if absent.

studio/src/promotions/mas-promotions-items-table.js
```diff
@@ -361,12 +417,109 @@ class MasPromotionsItemsTable extends LitElement {
...
+        return this.existingPromoVariationDefaultPaths.has(item.path);
+    }
+
+    async #viewPromoVariation(e, item) {
+        e.stopPropagation();
+        const promoTag = this.#promotionTagId;
+        const targetPath = promoTag ? buildPromoVariationPathForTag(item.path, promoTag) : null;
+        if (!targetPath) return;
+
+        const variation = await this.repository?.aem?.sites?.cf?.fragments?.getByPath(targetPath).catch(() => null);
```
Axelcureno: 🟡 should-fix — `getByPath` throws on *any* non-success (404, 401/403, 5xx, network), so collapsing all of them to `null` means a transient error shows the "may have been removed" message and flips the menu to "Create promo variation" on a variation that's actually fine. Worth distinguishing a true 404/empty from other failures before mutating state or showing the removed message.

Axelcureno (changes requested):
Did a deep multi-pass review (security, reactivity, races, MAS variation conventions, publish/AEM orchestration, navigation, coverage, separation-of-concerns). Solid feature overall — binding promos via tag + deterministic path instead of overloading the parent `variations` field is the right call, and the dedicated `promotion-variations` / `promotion-model` / `promotion-publish-utils` modules are good structure. A few things I'd want addressed before merge.

**Two I'd block on:**

1. `promotion-variations.js:167` — `getUnpublishedAttachedPromoVariations` excludes both `STATUS_PUBLISHED` *and* `STATUS_MODIFIED`. MODIFIED means published once with newer unpublished edits, so "Publish together" silently ships the stale last-published copy and the author's latest edits never go live (and promo variations aren't AEM references of the project, so reference-traversal won't pick them up either). Exclude only `STATUS_PUBLISHED`.
2. The promo-variations tab renders an authoritative "No promotion variations found" on a deep-link or hard reload into the editor. `loadPromotions()` only fires on Content/Promotions (`mas-repository.js:232/249`), but the probe reads the list synchronously (`#getPromotionProjectsForProbe`, ~2303) and never re-runs — the editor page never loads the list at all and doesn't self-heal. Cheapest fix: `await loadPromotions()` in the merge when the list is empty.

**Separation of concerns** — agree with @npeltier's `mas-repository.js` note. The repo picked up ~79 promo-specific lines (`#getPromotionProjectsForProbe`, `#mergePromoReferencesIntoFragmentData`, `resolveDefaultFragmentForPromoVariation`, `getUnpublishedAttachedPromoVariations`, `createPromoVariation`). Most are thin pass-throughs to `promotionVariations.*`, but per CONTRIBUTING.md `common/repository` is generic "repository interaction" while `promotions/` owns its own models — these belong in a promotions model. This is the boundary-erosion we keep hitting: every view adds its methods and `mas-repository` slowly becomes a god-object.

**Should-fix:**
- Partial publish shows a success toast while silently dropping variations (`promotion-publish-utils.js`) — failed `getByPath`/`getWithEtag` results are skipped, then the positive toast fires anyway.
- ReactiveController leaks a subscription per Edit-items modal open (`mas-promotions-items-table.js`) — built in `connectedCallback`, never removed on disconnect; the Spectrum overlay reparents the table each open/close.
- `getByPath().catch(() => null)` treats transient 5xx/network errors as "removed". Distinguish true 404 from other failures.
- `createPromoVariation` guard branches + the tag-strip invariant are untested — a double-tag regression would pass current tests.

**Nits:** no `..`/segment guard on the promo name before folder-create/copy writes (low risk — AEM tag ids can't be `..` — but worth a fail-closed guard); orphan `.tab-content-placeholder` CSS rule to delete; promotion breadcrumb missing when opening a variation from the variations tab; a few untested defensive/translatable branches.

For what it's worth: the `#categorizeVariations` dual-source merge checks out (the `promoPaths` Set dedupes correctly — no drift), and yesil's earlier `requestUpdate` concern looks resolved in the current revision.

studio/src/promotions/promotion-variations.js
Andrei4226: Fixed: `getUnpublishedAttachedPromoVariations` now excludes only `PUBLISHED`.
`MODIFIED` promo variations are included in the batch.

studio/src/mas-repository.js
Andrei4226: Fixed blocker: `getPromotionProjectsForProbe()` awaits `loadPromotions()` when the list is empty, used from `merge/probe` paths and from the variations tab `handleEdit`.
Fixed SoC: moved promo `probe/merge/create` helpers to `promotions-repository.js`

studio/src/promotions/promotion-publish-utils.js
Andrei4226: Fixed: now compare `fragments.length` vs `promoVariationPaths.length + 1` and show an info shortfall toast instead of the green success when some variations were dropped.

studio/src/promotions/mas-promotions-items-table.js
Andrei4226: Fixed: `connectedCallback` now guards with if `(this.#selectionController)` return before creating the `ReactiveController`, so reopening the `Edit-items` overlay doesn’t register duplicate store subscriptions.

studio/src/promotions/mas-promotions-items-table.js
Andrei4226: Fixed. Replaced the `blanket .catch(() => null)` with a shared helper in `promotion-model.js`:

- `isFragmentNotFoundError()` --> treats only 404 / not-found as “missing”
- `getFragmentByPathOrNull()` --> returns null on 404, re-throws 401/403/5xx/network errors

Wired this in `mas-promotions-items-table.js` in three places:

- `#viewPromoVariation`: transient failure → negative toast “Could not verify the promo variation…” and no state change; only a true 404/missing fragment shows the “may have been removed” message and updates the menu.
- `#syncExistingPromoVariations`: on transient error we keep the previous existingPromoVariationDefaultPaths entry so the row doesn’t flip to “Create promo variation” during a blip.
- `#createPromoVariation` same: verify step fails closed with the lookup-failed toast instead of assuming the variation doesn’t exist.

Tests added in promotion-model.test.js for 404 vs non-404 behavior.

studio/src/mas-repository.js
```diff
@@ -2282,6 +2281,48 @@ export class MasRepository extends LitElement {
         return createdFragment;
     }
 
+    /**
+     * Creates a promo variation for a default fragment under promotions/{promoName}/.
+     * @param {string} sourceFragmentId - Default fragment ID to copy from
+     * @param {string} promoTagId - mas:promotion/{promoName} tag from the promotion project
+     * @returns {Promise<Object>} The created promo variation fragment
+     */
+    async createPromoVariation(sourceFragmentId, promoTagId) {
```
npeltier: i still see it :-)

studio/src/mas-repository.js
```diff
@@ -2282,6 +2281,48 @@ export class MasRepository extends LitElement {
...
+            await this.refreshFragment(parentStore);
+            const parent = parentStore.get();
+            if (!parent) return;
+            const enriched = mergePromoVariationReferences(parent, [createdFragment]);
+            parentStore.refreshFrom(enriched);
+            this.#addToCache(parentStore.get());
+        });
+    }
+
+    async mergePromoReferencesIntoFragmentData(fragmentData) {
```
npeltier: same here

studio/src/mas-repository.js
```diff
@@ -2282,6 +2281,48 @@ export class MasRepository extends LitElement {
...
+        return promotionsRepository.mergePromoReferencesIntoFragmentData(this.aem, fragmentData, () => this.loadPromotions());
+    }
+
+    /**
+     * Resolves the default fragment for a promo variation path.
+     * @param {string} promoVariationPath
+     * @param {string} [promoVariationId]
+     * @returns {Promise<Object|null>}
+     */
+    async resolveDefaultFragmentForPromoVariation(promoVariationPath, promoVariationId) {
```
npeltier: and here

studio/src/mas-repository.js
```diff
@@ -2282,6 +2281,48 @@ export class MasRepository extends LitElement {
...
+        return promotionsRepository.resolveDefaultFragmentForPromoVariation(
+            this.aem,
+            promoVariationPath,
+            promoVariationId,
+            () => this.loadPromotions(),
+        );
+    }
+
+    async getUnpublishedAttachedPromoVariations(promotionFragment) {
```
npeltier: and there, are you doing this for simplifying access from the web component?

Andrei4226: Done.
Removed the 4 promo variation wrappers from `mas-repository` and callers now use `promotions-repository` directly with aem.
Still on `mas-repository`: `loadPromotions / getPromotionsPath`, expired auto-unpublish and promotions picker in `searchFragments`.

npeltier (approved):

Axelcureno (approved):
Re-reviewed — all my earlier blockers and should-fixes are addressed, several beyond what I asked. The MODIFIED-status publish bug, the deep-link "no variations" false-empty, the `mas-repository` god-object methods (now in `promotions-repository.js`), the partial-publish false-success toast, and the controller-reparent leak are all fixed and tested. Nice touch turning my throwaway `.catch(() => null)` note into a real `getFragmentByPathOrNull` 404-discriminator with proper per-call-site UX.

One leftover, non-blocking: `getUnpublishedAttachedPromoVariations` (`promotion-variations.js:165`) still uses bare `.catch(() => null)` — a transient 5xx there drops a variation from the publish-together prompt with no shortfall warning. Worth swapping in `getFragmentByPathOrNull` for consistency, but fine as a follow-up.

Approving. 👍

PR #924, opened 2026-06-02
MWPW-194042: fix onboarding scripts

3ch023:
Resolves https://jira.corp.adobe.com/browse/MWPW-194042

Non-functional change.
Updating onboarding scripts for new surfaces.

gen-locales to work based on locales.js
gen-dictionaries to set a correct parent for each dictionary index.

- After: https://mwpw-194042--mas--adobecom.aem.live/


npeltier (approved):

Axelcureno (changes requested):
Nice — this fixes the placeholder leak by having surface-defaults inherit from acom. Couple things before merge: the parent path is built as `acom/${localeName}` from the raw locale, which only works because every current surface-default also exists as an acom folder — if a future surface has a default locale acom only serves as a region, you'd get a dangling parent and the placeholder bug returns. Safer to resolve via `getDefaultLocaleCode('acom', localeName)` with a `null` fallback. Also `validate-dictionary-parent.mjs`'s `--dry-run` is a no-op (script is read-only) — drop it so it doesn't imply a write mode. And since the whole bug is parent-resolution logic, a small fixture test over a few locales would lock it down. Rest is solid.

yesil (commented):
agreed with @Axelcureno's points.

Given the importance of this logic, adding unit tests would help validate it and lock it down against regressions.

Axelcureno (approved):
All three points addressed — `getParentReference` resolves through acom's locale table and I walked the null path downstream: it lands as an empty parent field, never a dangling ref. Tests import the real exports and run in CI via the new workspace, nice.

Two small things, neither blocking:
- README repair-dictionary-entry section now says "Use `--live` to publish" but the script parses `--publish` (`--live` belongs to the validator) — it'd get silently ignored. The "validate published fragments on odin" comment above the repair example looks copy-pasted from the validator section too.
- The one branch still without a test is the null fallback itself — something like `getParentReference('express', 'xx_XX') === null` would lock down exactly the dangling-parent case.

Approving 👍

PR #929, opened 2026-06-03
MWPW-196031 - [m@s studio io] referencedBy API call returns cross-surface parents

honstar:
Resolves https://jira.corp.adobe.com/browse/MWPW-196031

- filter out parent references returned by `referencedBy` call to match target surface
- log a warning for cross-surface parent reference
- add unit-test

## 🧪 Nala E2E Tests

### To run Nala tests again:

- Remove the `run nala` label

## Test URLs:

- After: https://mwpw-NUMBER--mas--adobecom.aem.live/


yesil (approved):

Axelcureno (approved):
LGTM — clean, well-scoped fix. Reusing `PATH_TOKENS` to pull the surface and prefix-filtering `referencedBy` results is the right call, and the no-surface fallback preserving the old behavior is a nice touch. The test is genuinely good — asserting against `stub.getCalls()` instead of `callCounts` (with the comment explaining why the latter would be vacuous) actually proves the filter works rather than just passing alongside it. Confirmed locally: 39 passing.

One thing before merge: Prettier check is red on `project-start.test.js` — `npx prettier --write` on it should clear it.

honstar:
Thanks @yesil and @Axelcureno for the review 👍 I pushed the prettier fix, should be green now.

afmicka (approved):

PR #931, opened 2026-06-03
MWPW-190466: Store mnemonic tooltip text as text content for GLaaS translation

seanchoi0:
Resolves https://jira.corp.adobe.com/browse/MWPW-190466

Mnemonic tooltip text stored in `mnemonic-text` attribute was not being translated by GLaaS. This PR moves to text content inside `<mas-mnemonic>` — a standard text node that GLaaS translates normally — and adds `alt` as a tooltip fallback for icons without text.

## Changes

**`studio/src/rte/rte-field.js`**
- `toDOM`: serialize `mnemonicText` as text content (`['mas-mnemonic', attrs, text]`) instead of `mnemonic-text` attribute
- `parseDOM` (`mas-mnemonic` rule): read `textContent` first, fall back to `mnemonic-text` attribute for backward compatibility with existing fragments

**`web-components/src/mas-mnemonic.js`**
- `effectiveContent` getter: add `textContent` and `alt` to fallback chain (`tooltipText → mnemonicText → content → textContent → alt`)

**`studio/scripts/migrate-mnemonic-text.mjs`** *(new)*
- One-shot migration script to convert `mnemonic-text` attributes in existing AEM fragments to text content
- Dry-run by default; `--apply` to save; `--fragment <id|path>` to test on a single fragment; `--debug` to inspect raw field types

```bash
# Dry-run (report only, no changes)
node studio/scripts/migrate-mnemonic-text.mjs --host https://author-xxx.adobeaemcloud.com --token $TOKEN

# Test on a single fragment
node studio/scripts/migrate-mnemonic-text.mjs --host ... --token ... --fragment <id-or-aem-path>
node studio/scripts/migrate-mnemonic-text.mjs --host ... --token ... --fragment <id-or-aem-path> --apply

# Debug: inspect raw field names/types
node studio/scripts/migrate-mnemonic-text.mjs --host ... --token ... --fragment <id-or-aem-path> --debug

# Apply to English + specific locales
node studio/scripts/migrate-mnemonic-text.mjs --host ... --token ... --apply --locale de,fr,ja

# Apply to all locales
node studio/scripts/migrate-mnemonic-text.mjs --host ... --token ... --apply --locale all
```

> **Note:** Run migration after this PR is deployed. Running before deployment would leave tooltip text empty on live pages.

## Test URLs:

- After: https://mwpw-190466--mas--adobecom.aem.live/

## Content test:
- Before: https://main--cc--adobecom.aem.page/jp/drafts/seanchoi/express-test
- After: https://main--cc--adobecom.aem.page/jp/drafts/seanchoi/express-test?maslibs=MWPW-190466

## Checklist:

- [x] `mnemonicText` serialized as HTML text content → GLaaS translates it
- [x] Existing `mnemonic-text` attribute format still parsed (backward compat)
- [x] `alt` value shown as tooltip fallback when text content is absent
- [x] All existing tests pass
- [x] Migration script tested on single fragment (dry-run + apply)


studio/scripts/migrate-mnemonic-text.mjs
```diff
@@ -0,0 +1,387 @@
...
+        result.skipped = true;
+        return result;
+    }
+
+    if (!APPLY) return result; // dry-run
+
+    // Apply: fetch fresh ETag then PUT
+    try {
+        const fresh = await getFragmentById(id);
+        const toSave = { ...fresh, fields: updatedFields };
```
Andrei4226: Re-fetch fields before transform, not just ETag
On `--apply`, `updatedFields` is built from the search snapshot. 
You re-fetch the fragment for a fresh ETag, but still PUT the old transformed fields:

```
const fresh = await getFragmentById(id);
const toSave = { ...fresh, fields: updatedFields }; // stale
```

If the fragment was edited between search and PUT, those changes get silently overwritten. The ETag retry has the same issue, it refreshes the ETag but reuses the same stale `updatedFields`.

Fix: apply `transformHtml` to `fresh.fields` right before save (and in the retry block 277-278), so you're always transforming the latest content:

```
const fresh = await getFragmentById(id);
const freshUpdated = fresh.fields.map(/* same transform logic */);
const toSave = { ...fresh, fields: freshUpdated };
```

Andrei4226 (changes requested):

studio/scripts/migrate-mnemonic-text.mjs
```diff
@@ -0,0 +1,393 @@
...
+
+const LOCALES_ARG = args.locale;
+let searchLocales = [null]; // null = default path (no locale prefix = en)
+if (SINGLE_FRAGMENT) {
+    searchLocales = [null]; // ignored when --fragment is set
+} else if (LOCALES_ARG === 'all') {
+    searchLocales = [null, ...ALL_LOCALES];
+} else if (LOCALES_ARG) {
+    searchLocales = [
+        null,
```
Andrei4226: When `--locale` is passed (e.g. `--locale de`), `searchLocales` is always `[null, ...requested]`, so the script searches both `/content/dam/mas (EN)` and `/content/dam/mas/de`. Is that intentional?

seanchoi0: Yes, intentional. null maps to the EN base path (/content/dam/mas), which is the source locale for all variations. --locale de means "EN + German." Already-migrated EN fragments will be skipped
  since they won't contain mnemonic-text= anymore.

Andrei4226: Ok, thanks for the clarification!
LGTM 👍 
(please resolve the conflicts)

Andrei4226 (approved):

studio/scripts/migrate-mnemonic-text.mjs
```diff
@@ -0,0 +1,393 @@
...
+            console.log(`  — No mnemonic-text found`);
+        } else {
+            stats.converted++;
+            const action = APPLY ? '✓ Saved' : '~ Would update';
+            console.log(`  ${action}: fields [${result.fieldsChanged.join(', ')}]`);
+        }
+    } else {
+        // Bulk mode: search all matching fragments per locale path
+        for (const locale of searchLocales) {
+            const path = locale ? `${BASE_CONTENT_PATH}/${locale}` : BASE_CONTENT_PATH;
```
honstar: The support for scoping a specific `locale` is broken and is essentially a no-op since `/content/dam/mas/<locale>` does not exist.
Also, running a migration with a default full scope of ALL surfaces and locales is too broad imo and should rather be broken down by surface to keep the blast radius of a potential rollback isolated to a surface.

I'd recommend adding a `surface` parameter in addition to `locale` and do multiple migration runs instead of an all-in one.

studio/scripts/migrate-mnemonic-text.mjs
```diff
@@ -0,0 +1,393 @@
...
+                        const action = APPLY ? '✓' : '~';
+                        console.log(`  ${action} [${result.id}] ${result.title} — fields: ${result.fieldsChanged.join(', ')}`);
+                    }
+                }
+            }
+        }
+    }
+
+    console.log(`
+=== Report ===
```
honstar, line 446: In order to have a paper trail of what exact fragments have been modified, it'd be great to pipe the output to a timestamped log file including surface and locale in the filename (or at least document how to pipe it to a log file).

honstar (changes requested):

studio/scripts/migrate-mnemonic-text.mjs
seanchoi0: Thanks for catching this. Fixed in the latest commit:

- Added `--surface` parameter — path is now built as `/content/dam/mas/<surface>[/<locale>]` so locale scoping actually works.
- Updated the recommended usage in the doc comment to run one surface at a time (e.g. `--surface acom`, `--surface express`) rather than a single all-in sweep, matching your rollback isolation suggestion.

That said, `--locale` is mostly there as a safety net — the current plan is to only run this migration against the original EN (US) content. Localized strings go through GLaaS translation after the source fragments are fixed, so running the script against locale paths shouldn't be necessary.

studio/scripts/migrate-mnemonic-text.mjs
seanchoi0: Done. In `--apply` bulk mode the script now automatically writes a timestamped log file to the current directory, e.g.:
```
mnemonic-migration-acom-en-2026-06-13T12-00-00.log
```
The filename includes surface and locale so runs are easy to tell apart. For dry-run output, the doc comment notes `2>&1 | tee run.log` as the manual option.

web-components/src/mas-mnemonic.js
```diff
@@ -358,7 +359,14 @@ export default class MasMnemonic extends LitElement {
     }
 
     get effectiveContent() {
-        return this.tooltipText || this.mnemonicText || this.content || '';
+        return (
+            this.tooltipText ||
+            this.mnemonicText ||
+            this.content ||
+            this.textContent?.trim() ||
+            this.alt ||
```
honstar: This is a site-wide behavioral change, not just a translation fix! Previously, a `mas-mnemonic` with no tooltip returned '' -> `render()` produced an icon-only span with no tooltip.
Now any `mas-mnemonic` carrying an `alt` attribute (very common, set purely for a11y) will render a hover/tap tooltip. This affects every mnemonic instance across all surfaces, not just the Express simplified cards in the ticket. Please confirm this is intended.

Also note the `alt` path doesn't actually fix translation: `alt` is still an attribute, so GLaaS won't translate it. So locales without migrated text content will show untranslated English `alt` as a tooltip —arguably worse than the previous "no tooltip," because it now displays wrong-language text rather than nothing.

honstar (changes requested):

web-components/src/mas-mnemonic.js
seanchoi0: Thanks for pointing it out. When I originally planned the fix for this ticket, adding `alt` as a fallback felt like the right direction — but re-reading your comment made me realize it was a mistake.

Separately, I do think `alt` should actually be translatable, but this question in [#channel](https://adobedotcom.slack.com/archives/C092YP5CBPH/p1780060702768069) never got a response. For now I've removed the `alt` fallback.

honstar (approved):

3ch023 (approved):

PR #935, opened 2026-06-04
MWPW-197465 bumping #containers and their memory

npeltier:
also saving changes to the scripts

Resolves https://jira.corp.adobe.com/browse/MWPW-197465

- [] C1. Cover code with Unit Tests

## Test URLs:

quick regression on mas-io-url

### DIFF
--- a/io/www/app.config.yaml
+++ b/io/www/app.config.yaml
@@ -12,6 +12,9 @@ application:
                         function: src/fragment/pipeline.js
                         web: 'yes'
                         runtime: nodejs:22
+                        limits:
+                            concurrency: 20
+                            memory: 1024
                         annotations:
                             require-adobe-auth: false
                             final: true
--- a/scripts/k6/plans-fragment-fetches.js
+++ b/scripts/k6/plans-fragment-fetches.js
@@ -1,15 +1,52 @@
 import { check, sleep } from 'k6';
 import http from 'k6/http';
+import { SharedArray } from 'k6/data';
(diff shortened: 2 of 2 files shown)


afmicka (approved):

yesil (approved):

Andrei4226 (approved):
The default limit from Adobe I/O Runtime is 256 MB and a concurrency of 200, right? (https://developer.adobe.com/app-builder/docs/guides/runtime_guides/system-settings). 
And now we have added our own limit by increasing memory and decreasing concurrency in the container for high traffic, correct?

io/www/app.config.yaml
```diff
@@ -12,6 +12,9 @@ application:
                         function: src/fragment/pipeline.js
                         web: 'yes'
                         runtime: nodejs:22
+                        limits:
+                            concurrency: 20
+                            memory: 1024
```
honstar, line 17: Is there a reason why this was bumped to 1024mb memory? Since you're already lowering the `concurrency` to max 20 action invocations on the same container, the memory consumption should be less than before with the default value of 200, no?
Also, did you confirm with the app builder team that containers with 1024mb is a default container setting and already pre-warmed in the pool of available containers? IIRC, non-default container settings are bound to cause a bit of delay due to cold-starts.

honstar: nm, just saw that you added details and reasoning in the corresponding jira ticket.

PR #940, opened 2026-06-05
MWPW-197395: Bulk publish reports partial-success and rolls out missing locales

Axelcureno:
Bulk-publish projects that partially succeed now report accurate status and counts instead of falsely showing "all published" (or staying Draft). Locales where a card doesn't exist yet are reported as failures (not silently dropped), so authors know to translate/localize first.

Resolves https://jira.corp.adobe.com/browse/MWPW-197395

## What changed

- **Async dispatcher/worker** (`io/studio/src/bulk-publish/`): the `bulk-publish` action is now a thin dispatcher that returns 202 and fires a new `bulk-publish-worker` action (mirrors the translation `project-start-dispatcher/worker` pattern; avoids the IO timeout on large runs).
- **Reporting**: the worker is the single writer of the project fragment's `status` / `lastResult` / `publishedAt` / `publishedBy`. `lastResult` carries counts, a per-reason map, and a capped (100) failures list. New `Partially published` / `Failed` statuses.
- **Not-localized failures**: a card not found in a selected locale is reported as `not-localized` (publish only publishes what exists — it doesn't create or translate). Makes the "translate first, then publish" workflow explicit instead of failing silently.
- **Studio UI**: store defers terminal status to the worker; editor reads worker-written status; success banner gains a partial-success variant with a failures breakdown.

> **Scope note:** an earlier revision auto-created missing locale fragments via `/bin/localeSync`. Removed — incompatible with the MAS locale model (fallback locales like en_QA serve en_US at read-time and need no fragment; translation locales like de_DE need translated content, not an English copy). Publishing reports the gap; it doesn't fill it.

## Testing

- Unit: io/studio bulk-publish suite + studio bulk-publish suite, all green.
- Validated end-to-end against real Odin (sandbox throwaway project): publish → CF write of status/lastResult/publishedAt/publishedBy, for full-success and partial-success cases. Fixtures cleaned up.
- Model dependency confirmed: `lastResult`/`publishedAt`/`publishedBy`/`lastError` already exist on the prod `bulk-publish-project` content model.

## Test URLs

- After: https://mwpw-197395--mas--adobecom.aem.live/studio.html?io.base.url=https://14257-merchatscale-axel.adobeioruntime.net/api/v1/web/MerchAtScaleStudio#bulkPublishProjectId=0b2730a3-3d21-4ad9-b664-499612c07485&page=bulkPublishEditor&path=acom

> **Verified end-to-end against a personal IO dev deployment** (not reviewer-accessible): deployed the `bulk-publish` + `bulk-publish-worker` actions to `14257-merchatscale-axel` and ran a real bulk publish through local Studio at
> `http://localhost:3039/studio.html?io.base.url=https://14257-merchatscale-axel.adobeioruntime.net/api/v1/web/MerchAtScaleStudio#page=bulkPublish`
> Confirmed: dispatcher → worker → Odin publish → terminal `Published`/`Partially published` status written to the project CF, banner persists on reload, revert available. Two runtime-only bugs found and fixed here: `odinEndpoint` package-binding collision, and a 412 etag-conflict retry on the terminal CF write.


3ch023 (changes requested):
>>Adds a missing-locale rollout step so cards publish across selected locales that weren't yet rolled out.

this needs to be checked with Lucy, not sure whats the benefit of automatic rollout

Axelcureno:
> > > Adds a missing-locale rollout step so cards publish across selected locales that weren't yet rolled out.
> 
> this needs to be checked with Lucy, not sure whats the benefit of automatic rollout

removed from this PR, not a valid use case. @3ch023

Axelcureno:
Ready for review

3ch023 (approved):

honstar:
@Axelcureno : the test link you provided: https://mwpw-197395--mas--adobecom.aem.live/studio.html?io.base.url=https://14257-merchatscale-axel.adobeioruntime.net/api/v1/web/MerchAtScaleStudio#bulkPublishProjectId=0b2730a3-3d21-4ad9-b664-499612c07485&page=bulkPublishEditor&path=acom

There is no `io.base.url` request parameter. Instead, you only set the name of your workspace via `io.studio.env=axel`, this sets `io-base-url` in the `<head>`.

io/studio/src/bulk-publish/bulk-publish-worker.js
```diff
@@ -0,0 +1,125 @@
...
+
+    const resolved = resolve(paths, locales);
+    const details = await publish(resolved, odinEndpoint, authToken, logger);
+    relabelNotLocalized(details);
+
+    const result = buildResult({ details, startedAt, finishedAt: now().toISOString() });
+    const status = terminalStatus(result);
+    const finalSnapshots = removePendingMarker(snapshotEntries);
+
+    await updateProject(odinEndpoint, projectId, authToken, {
```
honstar, line 96: ### No recovery if the worker crashes between PUBLISHING and the terminal write

Between these 2 project writes:
```
await updateProject(odinEndpoint, projectId, authToken, { status: PROJECT_STATUS.PUBLISHING, ... });
// ... publishResolved + relabelNotLocalized + buildResult ...
await updateProject(odinEndpoint, projectId, authToken, { status, snapshots, publishedAt, ... });
```
If publishResolved throws, OpenWhisk times out (1 h), or the final `updateProject` exhausts its 3 retries, the project sits in `Publishing` forever. The 412 retry mitigates the common case but not OW crashes or pathological conflicts.
Worth a follow-up ticket for reconciliation or a manual "Reset to Draft" affordance.

studio/src/bulk-publish/bulk-publish-store.js
```diff
@@ -10,24 +11,37 @@ function patchProjectStore(project, fields) {
...
-        repository.refreshFragment(project).catch(() => {});
-        return result;
+        await fn({ ioBaseUrl, projectId: project.id, publishedBy, token });
+        for (let i = 0; i < maxPolls; i++) {
+            await repository.refreshFragment(project).catch(() => {});
+            const statusField = project.get()?.fields?.find((f) => f.name === 'status');
+            const status = statusField?.values?.[0];
+            if (terminalStatuses.has(status)) break;
+            await new Promise((resolve) => setTimeout(resolve, pollIntervalMs));
+        }
```
honstar, line 50: ### UI polling cap (5 min) < worker timeout (1 h)

With `maxPolls = 150, pollIntervalMs = 2000` → 5-minute cap. After that the local publishing flag clears silently even if the worker is still legitimately running. Large publishes will end with the UI in stale state.

Options: lengthen the cap, switch to exponential backoff after some threshold, or surface a "still publishing — check back later" affordance when we hit the cap with a non-terminal status.

io/studio/src/bulk-publish/bulk-publish-worker.js
```diff
@@ -0,0 +1,125 @@
...
+
+function removePendingMarker(entries) {
+    return entries.map((e) => {
+        const { publishComplete, ...rest } = JSON.parse(e);
+        return JSON.stringify(rest);
+    });
+}
+
+function terminalStatus(result) {
+    if (result.published === 0) return WORKER_STATUS.FAILED;
```
honstar, line 47: A project with no fragments now publishes 0, fails 0, and gets `Failed` with an empty failures list. The previous code returned 400 'Project has no fragments'. Either guard on `paths.length === 0` before dispatching/publishing,
Or have terminalStatus distinguish "zero work" from "zero successes."

studio/src/bulk-publish/bulk-publish-utils.js
```diff
@@ -7,3 +9,20 @@ export function getProjectFieldList(project, name) {
...
     return data.getFieldValues?.(name) ?? data[name] ?? [];
 }
+
+export function itemTypeFromPath(path) {
+    return path?.includes('/dictionary/') ? 'placeholder' : 'fragment';
+}
+
+export function itemTypeFromFragment(fragment) {
+    if (fragment?.model?.path === COLLECTION_MODEL_PATH) return 'collection';
+    if (fragment?.model?.path?.includes('/dictionnary')) return 'placeholder';
```
honstar: Use a constant for the dictionary model, analogous to `COLLECTION_MODEL_PATH` above. Using a string literal here increases the risk of misclassifying placeholders as fragments should we ever rename this model.

io/studio/src/bulk-publish/bulk-publish-worker.js
```diff
@@ -0,0 +1,125 @@
...
+    PROJECT_STATUS,
+    readProjectFragment,
+    updateProjectFragment,
+    getProjectPaths,
+    getProjectLocales,
+    getProjectTitle,
+    getProjectSnapshots,
+} = require('./project.js');
+
+const WORKER_STATUS = { PUBLISHED: 'Published', PARTIALLY_PUBLISHED: 'Partially published', FAILED: 'Failed' };
```
honstar, line 17: ### WORKER_STATUS and BULK_PUBLISH_STATUS define the same string table on opposite sides of the IO/UI boundary
In [studio/src/constants.js](https://github.com/adobecom/mas/pull/940/changes#diff-d2b76632716240f51f032ff159ad0741d2956713f8ed2cebaf902bd6933bf90bR291), the same strings are defined again. Since we don't really have a good way to share these, add a comment in both files at least to anchor the contract.

honstar (changes requested):

io/studio/src/bulk-publish/bulk-publish-worker.js
Axelcureno: Agreed — out of scope for this PR. Filed MWPW-198151 to track a reconciliation pass for stale `Publishing` projects (or a manual "Reset to Draft" affordance). The 412-retry stays as the common-case mitigation. (a2e1137a3)

studio/src/bulk-publish/bulk-publish-store.js
Axelcureno: Fixed in a2e1137a3. `startPublishing` now uses exponential backoff (2s → ×1.5 → 30s cap) instead of a fixed 2s interval, and returns `{timedOut}`. When the window elapses with a non-terminal status, the editor shows a "Still publishing — check back later" toast instead of silently clearing.

io/studio/src/bulk-publish/bulk-publish-worker.js
Axelcureno: Fixed in a2e1137a3. `terminalStatus` now guards `total === 0 → Published` before the `published === 0 → Failed` check, so a zero-fragment project reports vacuous success rather than a false failure. Added unit tests for the empty / all-failed / all-published / partial cases.

studio/src/bulk-publish/bulk-publish-utils.js
Axelcureno: Fixed in a2e1137a3. Added `DICTIONARY_MODEL_PATH` to constants.js (alongside `CARD_MODEL_PATH`/`COLLECTION_MODEL_PATH`) and used it here and in the duplicate at `render-utils.js`. Kept `.includes()` since paths can carry a suffix.

io/studio/src/bulk-publish/bulk-publish-worker.js
Axelcureno: Done in a2e1137a3. Added cross-referencing comments anchoring the contract in both `bulk-publish-worker.js` (`WORKER_STATUS`) and `constants.js` (`BULK_PUBLISH_STATUS`).

honstar (approved):

Roycethan (commented):
@Axelcureno  Can you plz resolve the conflict ?

PR #943, opened 2026-06-05
MWPW-197685: Split Nala studio suite into 3 runners at once

afmicka:
- Created the 4th self-hosted runner that is dedicated only for Nala Docs test suite
- Other 3 existing runner are dedicated to Nala Studio test suite and are moved to different regions in order to have different IPs
- The suite is split into 3 shards (matrix defined in the workflow) and each is being executed as a separate job resulting in being allocated to a separate runner.
- Each shard has assigned different UA
- Docs and the 3rd shard are sharing the IP so 1 worker is assigned to Docs and 3 workers to 3rd shard.
- Set the max to 45 rps per worker regardless of how many workers are used
- Added throttle to ODIN requests capping to max 20 RPS per UA
- Added EDS throttle to studio tests

Resolves https://jira.corp.adobe.com/browse/MWPW-197685

## 🧪 Nala E2E Tests

### To run Nala tests again:

- Remove the `run nala` label

## Test URLs:

- After: https://mwpw-197685--mas--adobecom.aem.live/


npeltier (approved):

nala/libs/eds-throttle.js
```diff
@@ -2,22 +2,32 @@
...
-        return Number.isFinite(v) && v > 0 ? v : 0;
+        if (Number.isFinite(v) && v > 0) return v;
+        console.warn(
+            `[NALA] NALA_EDS_MAX_RPS="${process.env.NALA_EDS_MAX_RPS}" is not a positive integer — ignoring, falling back to worker-derived default.\n`,
+        );
     }
-    return process.env.CI === 'true' ? DEFAULT_CI_EDS_MAX_RPS : 0;
+    const workers = Number.parseInt(process.env.NALA_WORKER_COUNT ?? '1', 10);
+    const n = Number.isFinite(workers) && workers > 0 ? workers : 1;
+    return Math.min(Math.floor(EDS_SAFE_TOTAL_RPS / n), EDS_MAX_RPS_PER_WORKER);
```
Axelcureno, line 30: Worth a comment that the 45 cap is load-bearing, not belt-and-braces: Noida runs 3 studio workers + 1 docs worker on one IP. Without the cap the studio shard alone computes `floor(180/3)=60` → 180, and docs adds another 180. The `min()` is what holds that IP at 180.

.github/workflows/run-nala.yml
```diff
@@ -26,24 +26,60 @@ jobs:
...
+                  if [ -n "${{ matrix.dirs }}" ]; then
+                    echo "SHARD_DIRS=${{ matrix.dirs }}" >> $GITHUB_ENV
+                  else
+                    # Shard 3 catch-all: every studio dir/file not claimed by shards 1-2.
+                    # New directories are automatically included here — no workflow change needed.
+                    CLAIMED="nala/studio/acom nala/studio/ahome nala/studio/ccd nala/studio/commerce"
+                    DIRS=""
+                    for path in nala/studio/*/; do
+                      dir="${path%/}"
+                      echo "$CLAIMED" | grep -qw "$dir" || DIRS="$DIRS $dir"
```
Axelcureno, line 75: `grep -qw` treats `-` as a non-word char, so a CLAIMED entry like `nala/studio/acom-v2` would also match a real `nala/studio/acom` dir and silently drop it from every shard. Safe with today's names, but `grep -qxF` on a newline list would keep it that way. Silent test loss is a nasty failure mode.

Axelcureno (commented):
Not in the description: local runs are now throttled. Old `resolveEdsMaxRps` returned 0 unless `CI === 'true'`; now local resolves workers=3 → 45 rps/worker. Probably a good thing given branch `.aem.live` 429s, but it slows every local nala run and nothing flags it. Intentional?

afmicka:
@Axelcureno thanks for the feedback. This PR has DO NOT REVIEW label as it is nowhere close to be done. The proposed solution does not work yet. Will check on the comments when i am back to this and the PR will close itself probably anyway.

As for throttling local, i did not come yet to that case as i am trying first to make it work on github, but thinking atm, i think we should throttle local too. If the local execution hits the limits, it will fail too and becomes useless. We can think of a mechanism that suits better for local though. Will check on that after successful throttling on VMs

PR #945, opened 2026-06-05
MWPW-197564: Match settings overrides by request country

yesil:
AU users on non-English locales (e.g. French) kept AU pricing but lost the annualized ABM price, which AU law requires wherever an ABM price is shown. `displayAnnual` was gated by an `en_AU`-keyed settings override, so a `fr_FR` + `country=AU` request never matched and the flag fell back to `false`. `resolveSettingEntry` now also matches an override when the request country equals the country segment of one of its locale codes (`en_AU` → `AU`). Locale is left intact so locale-specific content overrides keep resolving correctly.

Resolves https://jira.corp.adobe.com/browse/MWPW-197564

- [x] C1. Cover code with Unit Tests (4 new tests, full suite 100% coverage)
- [x] C3. Verify all Checks are green (unit tests)
- [ ] C5. ready to demo from Test Page in PR
- [ ] C6. read Jira once more to validate all AC's addressed

Test URLs:

- After: https://mwpw-197564--mas--adobecom.aem.live/


yesil:
Closing — reopening from the adobecom/mas branch directly instead of the fork.

PR #946, opened 2026-06-05
MWPW-197564: Match settings overrides by request country

yesil:
AU users on non-English locales (e.g. French) kept AU pricing but lost the annualized ABM price, which AU law requires wherever an ABM price is shown. `displayAnnual` was gated by an `en_AU`-keyed settings override, so a `fr_FR` + `country=AU` request never matched and the flag fell back to `false`. `resolveSettingEntry` now also matches an override when the request country equals the country segment of one of its locale codes (`en_AU` → `AU`). Locale is left intact so locale-specific content overrides keep resolving correctly.

Resolves https://jira.corp.adobe.com/browse/MWPW-197564

- [x] C1. Cover code with Unit Tests (4 new tests, full suite 100% coverage)
- [x] C2. Add a Nala test (double check with #fishbags if nala test is needed)
- [x] C3. Verify all Checks are green (unit tests)
- [x] C5. ready to demo from Test Page in PR
- [x] C6. read Jira once more to validate all AC's addressed

Test URLs:

- Before: https://main--mas--adobecom.aem.live/web-components/docs/ccd-mini.html?country=AU&locale=fr_FR
- After: https://mwpw-197564--mas--adobecom.aem.live/web-components/docs/ccd-mini.html?maslibs=mwpw-197564--mas--adobecom&country=AU&locale=fr_FR&mas-io-url=https://14257-merchatscale-yesil.adobeioruntime.net/api/v1/web/MerchAtScale/fragment


seanchoi0 (approved):

npeltier (commented):
1. are we sure we never have a country in acom requests? and that acom have "lingo ready" settings?  because this could change actual behaviour..i quickly check and it seems to me we should be good, but still worth checking.
2. there is in common matchesGeo that for now is used by promotions & customize that for now looks at locale & country tags for a match. I wonder if we could extract the tag -> country/locale process and reuse same helper

Axelcureno (approved):

yesil:
@npeltier I refactored as follow: https://github.com/adobecom/mas/pull/946/changes/9511a4801f3c04a20ee687e3b6c6d3b31ded722e
wdyt?

yesil:
@npeltier 
>are we sure we never have a country in acom requests?
we do on express: https://www.adobe.com/fr/express/pricing?country=lu

yesil:
@npeltier here is a test page:

https://main--da-express-milo--adobecom.aem.live/express/pricing?country=au&maslibs=mwpw-197564--mas--adobecom&mas-io-url=https://14257-merchatscale-yesil.adobeioruntime.net/api/v1/web/MerchAtScale/fragment

npeltier:
@yesil as discussed offline, the whole https://www.adobe.com/fr/* site is lingo, hence sends country parameter to mas/io (among approved list of french speaking countries). Please integrate that in testing of that PR

3ch023:
Hi @yesil ,
I checked https://mas.adobe.com/studio.html#page=settings&path=nala and it seems like ABM is enabled only en_AU, not whole 'AU' country.
if its enabled on en_AU then it shouldn't be showing up for fr_AU combination (your test link)

I think we should give GWP ability to create a setting for AU country, without limiting it to a specific language.

however it looks like your fix is breaking settings by showing abm where it shouldn't show up

3ch023 (changes requested):
discussed shortly on slack, we need to add ability for settings to be country specific, not just locale specific. similar to how it works for grouped variations

yesil:
@npeltier actually most of settings if not all are for countries and not for locale, am I wrong? Should be update the locale dialog to country dialog or ask GWP to consider those locales as countries for specific settings like display annual price.
cc: @3ch023

yesil:
@3ch023 your request is about settings management, this PR in its current form allows to toggle annual price for AU.
We need to address tagging of settings per locale outside the scope of this PR IMO.

yesil:
closing for now

PR #952, opened 2026-06-08
MWPW-197420: Show Studio path with locale in bulk publish items

Axelcureno:
Bulk publish item rows showed only the breadcrumb path. Prepend the fragment locale so users get full context — e.g. `[en_US] merch-card: SANDBOX / Individual / com / Creative Cloud Individual` — instead of the internal fragment path.

- Persist `fragment.locale` on each validated item.
- `itemLabel()` renders `[locale] authorPath`, falling back to the raw URL when no Studio path resolves.

Resolves https://jira.corp.adobe.com/browse/MWPW-197420

- [ ] C5: ready to do a demo from Test Page in PR
- [x] C.6 read your Jira one more time to validate that you've addressed all AC's

## Test URLs:

- Before: https://main--mas--adobecom.aem.live/studio.html#page=bulkPublish
- After: https://mwpw-197420--mas--adobecom.aem.live/studio.html#page=bulkPublish

## Screenshots:

Bulk publish items now render `[en_US] merch-card: SANDBOX / Individual / com / Creative Cloud Individual` (verified locally on :3040).


honstar:
I did a quick test and found the following issue: when opening any existing bulk publish project on the feature branch, the internal fragment paths are still shown, e.g. https://main--mas--adobecom.aem.live/studio.html#bulkPublishProjectId=1778f214-891b-46bf-8bb2-228e8327ad08&page=bulkPublishEditor&path=sandbox

Then I thought this only applies to new projects and created one: after pasting a fragment URL into the `Items` field, it resolved to the studio path as expected. I then saved the project, went back to the list view and reopened the project later. Instead of studio paths, the internal paths are displayed again. Looks like loading an existing project is not yet supported fully it seems.

honstar:
In addition to that, a fallback to internal path should not produce invalid links.

Axelcureno:
Good catch — confirmed and fixed. Item labels (`authorPath`/`locale`) were only computed during URL validation, never recomputed when reopening a saved project, so the load fell back to the raw fragment path.

Added `reEnrichItems()` which re-resolves each valid item's fragment on load and recomputes the Studio path + locale. Fails safe if a fragment no longer resolves; doesn't dirty the project.

Added 4 unit tests covering the re-enrichment (valid items, error items, unresolvable fragment, no-dirty).

Pre-existing projects need one reopen to backfill.

studio/src/bulk-publish/mas-bulk-publish-editor.js
```diff
@@ -569,6 +570,36 @@ class MasBulkPublishEditor extends LitElement {
...
+                        authorPath: authorPath || item.authorPath || null,
+                        locale: fragment.locale || item.locale || null,
+                        href: href || item.href || null,
+                    };
+                } catch {
+                    return item;
+                }
+            }),
+        );
+        this.localItems = enriched;
```
honstar, line 614: ### Potential race with validate() can drop user-typed URLs from the UI
`connectedCallback` fires `reEnrichItems()` without awaiting (mas-bulk-publish-editor.js:115). `reEnrichItems()` snapshots `this.items` at entry, then asynchronously fetches and finally does `this.localItems = enriched`. If the user starts typing URLs and triggers validate() while re-enrichment is still in flight, `validate`'s write to `localItems` (with the new pending items) gets overwritten when re-enrichment completes — its enriched array is from the pre-validate snapshot, so the pending items disappear.

Proposed fix: adopt the existing `#validateId` cancellation pattern, or capture a snapshot id at the top of `reEnrichItems()` and bail before the final assignment if it's stale. Worth doing — the surrounding code already uses this pattern for `validate()`.

honstar, line 614: ### No signal.aborted check before writing to localItems
If the user navigates away during enrichment, `disconnectedCallback` aborts the controller but `reEnrichItems()` keeps running and writes to a disconnected element. Lit tolerates this, so it's not a crash — just inconsistent with the careful aborting in `connectedCallback`. Low priority.

studio/src/bulk-publish/mas-bulk-publish-editor.js
```diff
@@ -569,6 +570,36 @@ class MasBulkPublishEditor extends LitElement {
         });
     }
 
+    async reEnrichItems() {
+        const items = this.items;
+        if (!items.some((i) => i.path && !i.authorPath)) return;
+        const { generateCodeToUse } = await import('../utils.js');
+        const surface = Store.search.get()?.path;
+        const enriched = await Promise.all(
```
honstar: This could be an issue with a lot of items in a bulk publish project: `Promise.all` fires N concurrent fragment fetches on load which can cause 429 responses from odin. Please consider adding a small concurrency cap to avoid this. Otherwise, you're penalized with a 1m waiting time before requests are handled properly again.

honstar (changes requested):

Axelcureno:
Thanks @honstar — all four addressed in e3ee4cc1f:

- **Race with validate():** `reEnrichItems()` now captures the `#validateId` generation at entry and bails before writing `localItems` if a later `validate()` or disconnect bumped it. Reuses the existing cancellation pattern.
- **signal.aborted:** covered by the same guard — `disconnectedCallback` bumps `#validateId`, so a stale enrichment won't write.
- **429s from Odin:** replaced `Promise.all` with a concurrency cap of 4 (`mapWithConcurrency`).
- **Invalid fallback links:** rows now render plain text instead of a broken `<a>` when no real http(s) href resolves.

Added 4 unit tests (cancellation, concurrency cap, link/plain-text rendering).

honstar (approved):

3ch023 (approved):

studio/src/bulk-publish/mas-bulk-publish-editor.js
```diff
@@ -41,6 +41,21 @@ const PUBLISH_BLOCKED_REASON = {
     ALL_ITEMS_PUBLISHED: 'All items are already published',
 };
 
+const ENRICH_CONCURRENCY = 4;
```
honstar: A bulk publish project with 100+ items takes multiple seconds to load. I'd suggest to increase the concurrency and double it to 8.

afmicka (approved):

studio/src/bulk-publish/mas-bulk-publish-editor.js
Axelcureno: Good callout. Increased to 8.

PR #966, opened 2026-06-11
MWPW-190535 [M@S Usability] M@S Studio Search by Fragment Title

bozojovicic:
Resolves https://jira.corp.adobe.com/browse/MWPW-190535

Test pages https://mwpw-190535--mas--adobecom.aem.live/studio.html#page=content&path=acom&query=CC+Catalog+Merch+Card%3A+Creative+Cloud+Pro%3A+Individuals%3A+Base+default

https://mwpw-190535--mas--adobecom.aem.live/studio.html#page=content&path=sandbox&query=CC+%22Catalog%22+Collection%3A+CHILD%27s+%28PDF+%26+E-signatures%29

Screenshot 2026-06-12 at 14 25 47

Improvements :
- The most important : it works now. Users often search by the fragment title. Title of ACOM fragments is very often full of special characters like `: & ( )` and we used to encode them `text: encodeURIComponent(query)` so if you search for `Collection:` we will send in the search query `Collection%3A` and no results is displayed. I removed that encoding and now it works also with special characters.
- If you search for a text with more than one word, MAS studio used to search for the longest word and then go through all results to match the right fragments. This can take a lot of time to get the first results. Now I first send one search request with the exact text that user entered, so if we have a fragment with exactly that text it will be displayed right away and it will be displayed on the top. After that it will do what it also did before (search for the longest word and filter ...)
- If you search for a text with more than one word, MAS studio used to display only the results that contain that text in exactly that order. Now the order does not matter. It will display all fragments that have all words of the query text in at least one field. 

## 🧪 Nala E2E Tests

### To run Nala tests again:

- Remove the `run nala` label

## Test URLs:

- After: https://mwpw-190535--mas--adobecom.aem.live/


yesil (approved):
I'm a bit skeptical to client side interpretation of query results, but since it is already broken, I'm fine with this improvement.

TsayAdobe (approved):

Blainegunn (approved):

PR #972, opened 2026-06-16
MWPW-198613: Fix empty promotions tab and surface picker filter for promotions

Andrei4226:
Resolves https://jira.corp.adobe.com/browse/MWPW-198613

**Problem:**

When editing a promotion project that has a specific surface (e.g. "Nala"), the fragment picker incorrectly uses the surface from the global navbar instead of pulling it from the promotion project's own surfaces field. As a result, fragments added to the promotion end up with wrong paths (e.g. merch-card: SANDBOX instead of merch-card: NALA).
Additionally, searching for a fragment by UUID (via the copy-code deep link) was not rendering the promotion variation in the Promotions column of the fragments table, the tab appeared empty.

**Fix applied:**

- The surface picker is now initialized from the promotion project's surfaces field, not from the global navbar picker.
- Surface sync works correctly when navigating to a parent fragment or a promotion fragment.
- The promotions tab now populates correctly when searching by UUID.

## 🧪 Nala E2E Tests

### To run Nala tests again:

- Remove the `run nala` label

## Test URLs:

- After: https://mwpw-198613--mas--adobecom.aem.live/


honstar:
@Andrei4226 , please resolve the conflict, thanks!

mirafedas (approved):

studio/src/common/utils/render-utils.js
```diff
@@ -13,7 +13,12 @@ import { isPromoVariationPath } from '../../promotions/promotion-model.js';
  * @returns {string}
  */
 export function getStudioFragmentDisplayPath(fragment) {
-    return generateCodeToUse(fragment, Store.search.get().path, Store.page.get())?.authorPath || '';
+    const page = Store.page.get();
+    const path =
+        page === PAGE_NAMES.PROMOTIONS_EDITOR
+            ? (Store.promotions.itemPickerSurface.get() ?? extractSurfaceFromPath(fragment?.path) ?? Store.search.get().path)
```
honstar: This could end up in `path=''` since you're using nullish coalescing operators. Consider using `||` instead to guard against empty values as well.

studio/src/promotions/mas-promotions-items-table.js
```diff
@@ -749,6 +751,8 @@ class MasPromotionsItemsTable extends LitElement {
         e.stopPropagation();
         if (!item?.id || !item?.path) return;
         Store.promotions.promotionId.set(null);
+        const surface = extractSurfaceFromPath(item.path);
```
honstar: DRY: consider moving this to a helper function in `utils.js` since the exact same 2 lines are duplicated 3 times.

honstar (changes requested):

studio/src/promotions/mas-promotions-items-table.js
Andrei4226: Thanks!
Extracted in `applySearchSurfaceFromPath` function.

studio/src/common/utils/render-utils.js
Andrei4226: Done.

studio/src/mas-fragment-editor.js
```diff
@@ -1300,10 +1300,13 @@ export default class MasFragmentEditor extends LitElement {
 
     async navigateToLocaleDefaultFragment() {
         if (!this.localeDefaultFragment) return;
-        const parentLocale = extractLocaleFromPath(this.localeDefaultFragment.path);
+        const parentPath = this.localeDefaultFragment.path;
+        const parentLocale = extractLocaleFromPath(parentPath);
         // Reset changes to avoid discard dialog since we're navigating to the parent
         Store.editor.resetChanges();
         Store.promotions.promotionId.set(null);
+        const surface = extractSurfaceFromPath(parentPath);
```
honstar: One left-over, please use the new `applySearchSurfaceFromPath` helper function.

honstar (changes requested):
One remaining minor issue, looks good otherwise 👍

studio/src/mas-fragment-editor.js
Andrei4226: Thanks, I missed this replace.
I added it now 👍

honstar (approved):

npeltier (dismissed):
QA nico speaking :-)  could we land on "active promotions" tab per default?

PR #983, opened 2026-06-17
MWPW-195255: Fix fragments table select-mode checkbox bugs

seanchoi0:
Part of [MWPW-195255](https://jira.corp.adobe.com/browse/MWPW-195255). This PR focuses on the **select-mode checkbox bugs** in the MAS Studio fragments table: parent row checkboxes disappearing after filters/search reload, and variation rows disappearing or breaking when selected.

## Summary

- Re-applies parent `sp-table` checkbox cells after filter/search/expand row changes via `refreshTableSelection()` (`selects` toggle pattern)
- Splits parent vs variation selection so parent table changes do not wipe variation IDs from `Store.selection`
- Strips parent-injected `sp-table-checkbox-cell` from nested variation rows; variation checkboxes are manual in the locale name cell
- Keeps PROD variation styling (light DOM + global CSS); extracted helpers to `mas-content-table-selection.js`

## Why `mas-content-table-selection.js`?

The fix needs three distinct selection concerns in one table UI, and `mas-content.js` was already handling table rendering, filters, pagination, and personalization grouping. Keeping all of that inline made the select-mode logic harder to follow and harder to test.

This module isolates the **parent vs variation selection contract**:

| Helper | Role |
|--------|------|
| `getParentRowSelection` | Binds only top-level fragment IDs to the parent `sp-table` `selected` prop |
| `mergeParentTableSelection` | On parent checkbox change, merges parent `selectedSet` with existing variation IDs instead of replacing `Store.selection` |
| `stripNestedVariationSelectControls` | Removes checkbox cells/state that Spectrum injects into nested variation `sp-table-row` elements in light DOM |

Extracting these as pure functions also lets us unit-test the merge/strip behavior without mounting the full `mas-content` tree (which avoids flaky async/update-loop issues in integration tests).

## Test plan

- [ ] Open MAS Studio → Fragments → table view
- [ ] Click **Select** — parent row checkboxes appear
- [ ] Apply a filter or search — checkboxes remain on parent rows
- [ ] Expand a fragment with locale variations
- [ ] Select variation checkboxes — rows stay visible, blue row styling matches PROD
- [ ] Selection panel count updates for parent + variation selections
- [x] Unit tests: `npm run test:ci -- --files test/mas-content-table-selection.test.js test/mas-content.test.js test/mas-fragment.test.js test/mas-fragment-table.test.js test/mas-fragment-variations.test.js` (70 passed)

Resolves https://jira.corp.adobe.com/browse/MWPW-195255

## Test URLs

Studio (local):

- Before: `http://localhost:3000/studio.html` (main)
- After: `http://localhost:3000/studio.html` on branch `MWPW-195255-1` worktree

Milо preview (if applicable):

- After: https://mwpw-195255-1--mas--adobecom.aem.live/


studio/src/mas-fragment-table.js
```diff
@@ -84,8 +84,12 @@ class MasFragmentTable extends LitElement {
     }
 
     update(changedProperties) {
-        if (changedProperties.has('fragmentStore')) {
-            this.#reactiveControllers.updateStores([this.fragmentStore]);
+        if (changedProperties.has('fragmentStore') || changedProperties.has('nested')) {
+            const stores = [this.fragmentStore];
+            if (this.nested) {
```
Andrei4226, line 90: Subscribing nested rows to `Store.selection` means every `mas-fragment-table` instance with `nested=true` re-renders on each selection change. If a parent fragment has 10 variations expanded, checking one variation triggers 10 re-renders. 
Maybe we can optimize by passing i`sSelected` as a boolean property from `mas-fragment-variations` down to each `mas-fragment-table` row. That way only `mas-fragment-variations` subscribes to `Store.selection` (one subscriber instead of n), computes `isSelected` per variation and passes it down (O(1) re-renders).

Andrei4226 (approved):

Andrei4226 (commented):
As a follow‑up ticket/suggestion for the future: I think it will be clearer if a list appears on the right side when you are in table view and you want to select multiple fragments by checkbox. Why? Because if I select 2 fragments and after that I want to apply a filter, I can’t see which fragments I selected before (in the floating navbar below I can see only the number of selected items, not which ones they actually are).
I think it would be clearer.

honstar:
@seanchoi0 : can you please confirm that the scope of this PR does not include selection support for both promo and grouped variations? Since on `main`, there are checkboxes present whereas on your branch, they are not.

seanchoi0:
Hi @honstar, thanks for catching that!

The scope should include promo and grouped variations — I've now extended the checkbox fix to cover all three variation types (locale, promotion, grouped). The checkboxes now appear in the correct order (before the expand arrow) and toggle selection via Store.selection directly, bypassing the buggy sp-table selection mechanism that was the root cause of the original issue.

Happy to address any remaining concerns.

studio/src/mas-fragment-table.js
yesil, line 90: **[MEDIUM]** Each nested `mas-fragment-table` subscribes to `Store.selecting` and `Store.selection`. With a parent expanded to 10+ variations, toggling one checkbox re-renders every nested row (@Andrei4226 flagged this). Consider passing `isSelected` from `mas-fragment-variations` so only the parent subscribes — follow-up, not a merge blocker.

*Reviewed with ❤️ by [Claude Code](https://claude.ai) 🤖 and [Scout](https://github.com/Adobe-AIFoundations/scout)*

studio/src/mas-content.js
```diff
@@ -159,11 +191,52 @@ class MasContent extends LitElement {
...
+        if (event.target !== event.currentTarget) return;
+
+        Store.selection.set(mergeParentTableSelection(event.target.selectedSet, this.fragments.value, this.selection.value));
+    }
+
+    /**
+     * Re-applies sp-table checkbox cells after rows are added or removed.
+     * Spectrum sp-table only injects checkboxes when `selects` toggles.
+     */
+    async refreshTableSelection() {
```
yesil, line 209: **[MEDIUM]** `refreshTableSelection` toggles `tableSelects` off/on to work around Spectrum checkbox injection, but only the loading no-op path is tested. A happy-path test asserting `tableSelects` cycles `undefined → 'multiple'` would guard the core fix.

*Reviewed with ❤️ by [Claude Code](https://claude.ai) 🤖 and [Scout](https://github.com/Adobe-AIFoundations/scout)*

studio/src/mas-content.js
```diff
@@ -159,11 +191,52 @@ class MasContent extends LitElement {
         `;
     }
 
+    get parentRowSelection() {
+        return getParentRowSelection(this.fragments.value, this.selection.value);
+    }
+
     updateTableSelection(event) {
-        Store.selection.set(Array.from(event.target.selectedSet));
+        if (this.#refreshInProgress) return;
```
yesil, line 199: **[LOW]** `#refreshInProgress` drops parent `sp-table` `@change` events during refresh. Window is brief; unlikely to lose user input, but worth noting.

*Reviewed with ❤️ by [Claude Code](https://claude.ai) 🤖 and [Scout](https://github.com/Adobe-AIFoundations/scout)*

yesil (approved):
## PR Review: MWPW-195255: Fix fragments table select-mode checkbox bugs

### Summary

Focused fix for MAS Studio fragments table select mode: parent `sp-table` checkboxes disappearing after filter/search reload, and variation rows breaking when selected. The PR extracts selection helpers into `mas-content-table-selection.js`, toggles `sp-table` `selects` to re-inject parent checkboxes, strips Spectrum-injected checkboxes from nested variation rows, and adds custom `sp-checkbox` controls for locale/grouped/promo variations.

**Scope match:** 10 files, ~489 additions — matches the description.

**Overall risk:** Low–medium. Core logic is well-factored and tested; main residual risk is render performance with many expanded variations.

### Change Surface
- ~15 meaningful symbols changed across 2 feature commits
- New module: `mas-content-table-selection.js` with pure helper exports
- Surfaces: `mas-content`, `mas-fragment-table`, `mas-fragment-variations`, `mas-fragment`
- Scope: match

### Blast Radius
- `mergeParentTableSelection` → `updateTableSelection` only — no gaps
- `refreshTableSelection` → `#maybeRefreshTableSelection` + `table-selection-refresh` event — no gaps
- No dangling references or new orphans

### Prior Knowledge Surfaced
- Andrei's inline comment on nested `Store.selection` subscriptions remains the main architectural note (see inline).

### Test Gaps
- `refreshTableSelection` success path (toggle + strip) not covered — only loading no-op tested
- No integration test for filter-change → checkbox re-injection
- Pure helpers, merge/bubble logic, grouped/promo checkbox placement well covered

### Affected-Test Run
- 41 tests in 4 files — all passed

### Security
- No issues found

### Suggestions Worth Supporting
- **Andrei4226 — nested row re-render optimization:** Agreed as follow-up. Current approach is correct; at scale with many expanded variations it may feel sluggish.
- **Andrei4226 — selected-items side panel:** Good UX follow-up ticket; out of scope for this bugfix.

### Verdict
**Approve with nits**

The fix is sound: parent/variation selection is split cleanly, nested `sp-table` injection is stripped, custom checkboxes cover locale/grouped/promo paths, and unit tests cover the critical merge/bubble logic.

---
*Reviewed with ❤️ by [Claude Code](https://claude.ai) 🤖 and [Scout](https://github.com/Adobe-AIFoundations/scout)*

studio/src/mas-fragment-table.js
```diff
@@ -206,20 +219,37 @@ class MasFragmentTable extends LitElement {
...
+                              @change=${this.handleVariationSelect}
+                              @click=${(e) => e.stopPropagation()}
+                          ></sp-checkbox>
+                      </sp-table-cell>`
+                    : ''}
+                ${this.nested && !this.toggleExpand
                     ? ''
                     : html`<sp-table-cell class="expand-cell" @click=${this.toggleExpand}>
                           <button class="expand-button" aria-label="${this.expanded ? 'Collapse' : 'Expand'} row">
+                              <!-- audit-ok: icon-only expand toggle, no Spectrum equivalent for this pattern -->
```
honstar: This is an HTML comment that would render in the DOM for every expandable row. Either remove it or turn it into a JS comment instead.

honstar (changes requested):
One minor issue but otherwise looks good 👍

studio/src/mas-fragment-table.js
seanchoi0: Good catch — removed in cca7b3815.

That `audit-ok` HTML comment was added by our local `studio_audit.py` PostToolUse hook (it flags native `<button>` in `studio/src`). It was meant as a source-level suppress marker, but putting it inside the Lit template was wrong: it renders a DOM comment node on every expandable row, and it wasn't even on the same source line as `<button>` so it wouldn't reliably suppress the hook anyway.

`main` already has this expand toggle without any `audit-ok` comment, so I dropped it entirely rather than moving it to a JS comment.

studio/src/mas-content.js
seanchoi0: Added in e92ba6a5b — `refreshTableSelection toggles tableSelects off then back to multiple` in `test/mas-content.test.js`.

The test mounts `mas-content` in select mode, calls `refreshTableSelection()` directly (not the loading early-return path), asserts `tableSelects` is `undefined` synchronously after the call starts, then `'multiple'` once the async refresh completes.

Nested checkbox stripping remains covered in `test/mas-content-table-selection.test.js` (`stripNestedVariationSelectControls`).

studio/src/mas-fragment-table.js
seanchoi0: Agreed — nested rows each subscribing to `Store.selection` is O(n) re-renders on every toggle. Correct for this bugfix; we'll track passing `isSelected` from `mas-fragment-variations` as a follow-up (same as @Andrei4226's note).

studio/src/mas-content.js
seanchoi0: Acknowledged. The window is brief and mainly covers programmatic refresh after filter/expand; user clicks during that window are unlikely. We can queue changes in a follow-up if we see reports.

honstar (approved):

3ch023:
hi @seanchoi0 i tried toQA it, but coudlnt' select grouped variations on https://mwpw-195255-1--mas--adobecom.aem.live/studio.html#page=content&path=acom&query=37e7d4d8-6983-4e22-a656-5a69b00626a5

seanchoi0:
Hi @3ch023, thanks for QA — confirmed the grouped/promo selection bug on preview.

**Root cause:** parent `sp-table` injects its own checkbox cell into nested variation rows, which conflicted with our manual checkbox column — clicks hit the wrong control or did nothing.

**Fix in [`275ff1b`](https://github.com/adobecom/mas/commit/275ff1b):**
- Checkbox co-located in the expand cell (before the chevron) to match the original layout
- Row-click selection for nested variations (same as parent fragments)
- Injected nested `sp-table-checkbox-cell` hidden via CSS + strip helper

Please re-test on the preview URL once AEM sync completes for `275ff1b`. Grouped and promotion tabs should both select via checkbox or row click. Let me know if anything still fails.

3ch023 (dismissed):
i could select a grouped variation but when clicking on 'copy links' or 'publish' nothing happens

seanchoi0:
Hi @3ch023 — thanks for the re-test and for catching the bulk-action gap.

**Root cause:** Copy links / Publish only looked up fragments in `Store.fragments.list.data` (top-level page rows). Variation IDs live in the parent row's `references` (loaded on expand), so selection worked but bulk actions resolved 0 fragments.

**Fix in [`1511401`](https://github.com/adobecom/mas/commit/151140142):**
- Added `fragment-selection-utils.js` — resolves selection IDs from top-level list *or* parent `references` (same pattern as `editFragment()`)
- Wired into Copy links, Copy Code, Copy to folder, and `bulkPublishFragments()` (with `getById` fallback when etag is missing)
- After publishing a variation, refreshes the parent store so status stays in sync

Please re-test grouped/promo variation selection → **Copy links** and **Publish** once AEM sync completes. Let me know if anything still fails.

PR #998, opened 2026-06-22
MWPW-199014: Dynamic imports in studio.js

mirafedas:
Resolves https://jira.corp.adobe.com/browse/MWPW-199014 .

In studio/src/studio.js — added lazy-load page components
Replaced 14 eager static imports (one per page component) with a #lazyLoad(elementName, importPath) helper called from each page getter. Each getter now fires import() on first render for its page, waits for customElements.whenDefined() to schedule a re-render, then returns the template on the second pass.

Additionally:
In studio/src/translation/translation-utils.js —fixed the import source
getFragmentPartsToUse was imported from editor-panel.js. Because editor-panel.js is now lazy-loaded (only on PAGE_NAMES.CONTENT), keeping that import would pull editor-panel.js into the bundle eagerly, defeating the lazy-load entirely. The import was redirected to utils.js, which already exported the same function. The call signature changed accordingly: utils.js takes (fragment, path) while editor-panel.js took (store, fragment).

In studio/src/editor-panel.js — aligned promotion field with utils.js
This change is needed because redirecting translation-utils.js to utils.js created a silent inconsistency: utils.js used fragment?.getCurrentTagTitle?.(TAG_PROMOTION_PREFIX) for the promotion field, while editor-panel.js (still used by the editor panel itself and mas-fragment-editor.js) used the older fragment?.getTagTitle(TAG_PROMOTION_PREFIX).

The two methods differ: getTagTitle reads from the fragment's original tags array (initial state), while getCurrentTagTitle reads from the tags field values, which reflect in-flight edits the user has made but not yet saved. After the translation-utils.js redirect, the translation table and the editor panel could show different promotion labels for the same fragment — getCurrentTagTitle on one surface, getTagTitle on the other. Aligning editor-panel.js to getCurrentTagTitle?.() makes all three surfaces consistent.

## 🧪 Nala E2E Tests

### To run Nala tests again:

- Remove the `run nala` label

## Test URLs:

- After: https://mwpw-444444--mas--adobecom.aem.live/


yesil:
@mirafedas the followings are still to be deferred because they import other files:

https://mwpw-444444--mas--adobecom.aem.live/studio/src/editors/merch-card-collection-editor.js
https://mwpw-444444--mas--adobecom.aem.live/studio/src/editors/merch-card-editor.js
https://mwpw-444444--mas--adobecom.aem.live/studio/src/common/components/mas-items-selector.js

studio/src/mas-fragment-editor.js
honstar: This should also be refactored to use the `getFragmentPartsToUse` function from `utils.js`. Otherwise statically importing `editor-panel.js` would eagerly load it including its tree of dependencies which would defeat the purpose of reducing the resource fetches via lazy loading.

honstar: In fact, [`mas-promotions-editor:14`](https://github.com/adobecom/mas/blob/main/studio/src/promotions/mas-promotions-editor.js#L14) should also be refactored accordingly to avoid the eager loading of `editor-panel.js`.

studio/src/mas-toolbar.js
```diff
@@ -214,13 +214,20 @@ class MasToolbar extends LitElement {
...
         this.selectedContentType = type;
         this.popover.open = false;
-        this.openCreateDialog();
+        await this.openCreateDialog();
     }
 
-    openCreateDialog() {
+    async openCreateDialog() {
+        if (!customElements.get('mas-create-dialog')) {
+            import('./mas-create-dialog.js')
```
honstar: Shouldn't this `await` the import? Otherwise `this.createDialogOpen = true` is set before the element is defined.

honstar (changes requested):

studio/src/placeholders/mas-placeholders-creation-modal.js
```diff
@@ -2,6 +2,7 @@ import { LitElement, html } from 'lit';
 import { createPlaceholder } from './mas-placeholders-repository.js';
 import { normalizeKey, showToast } from '../utils.js';
 import Store from '../store.js';
+import '../rte/rte-field.js';
```
honstar, line 5: Is this really needed?

studio/src/placeholders/mas-placeholders-item.js
```diff
@@ -8,6 +8,7 @@ import { confirmation } from '../mas-confirm-dialog.js';
 import { showToast } from '../utils.js';
 import { FragmentStore } from '../reactivity/fragment-store.js';
 import { Placeholder } from '../aem/placeholder.js';
+import '../rte/rte-field.js';
```
honstar, line 11: Same as above, why is this `import` needed?

honstar (commented):
Can you explain why the `./rte/rte-field.js` imports are needed?

studio/src/placeholders/mas-placeholders-creation-modal.js
mirafedas: Yes, previously it worked without this import because the rte-field was imported in studio.js. I moved the import to files where the rte-field is used, and removed from studio.js to not load it when we are on the Fragments page (because rte-field in not used there).

studio/src/placeholders/mas-placeholders-item.js
mirafedas: Same as in the mas-placeholders-creation-modal.js comment :)

studio/src/mas-toolbar.js
```diff
@@ -214,14 +214,22 @@ class MasToolbar extends LitElement {
...
         this.selectedContentType = type;
         this.popover.open = false;
-        this.openCreateDialog();
+        await this.openCreateDialog();
     }
 
-    openCreateDialog() {
-        this.createDialogOpen = true;
+    async openCreateDialog() {
+        if (customElements.get('mas-create-dialog')) return;
```
honstar: This is in fact a regression: once `mas-create-dialog` has been defined (globally, permanently, after the first create), every subsequent `selectContentType` → `openCreateDialog` hits the `customElements.get(...) return` and never sets `createDialogOpen = true` - so the create dialog silently fails to open on the 2nd+ attempt in a session.

Suggest to separate "load" from "open":

```suggestion
    if (!customElements.get('mas-create-dialog')) {
        try {
            await import('./mas-create-dialog.js');
            this.requestUpdate();
        } catch {
            Events.toast.emit({ variant: 'negative', content: 'Failed to load create dialog' });
            return;
        }
    }
    this.createDialogOpen = true;
```

studio/test/mas-toolbar.test.js
```diff
@@ -0,0 +1,71 @@
...
+    afterEach(() => {
+        sandbox.restore();
+    });
+
+    it('returns early without modifying state when mas-create-dialog is already registered', async () => {
+        sandbox.stub(customElements, 'get').returns(class extends HTMLElement {});
+
+        await el.openCreateDialog();
+
+        expect(el.createDialogOpen).to.be.false;
```
honstar: This is in fact wrong: when calling `openCreateDialog`, the `createDialogOpen` property needs to be `true`, the opposite.

studio/src/rte/rte-mnemonic-editor.js
```diff
@@ -1,5 +1,6 @@
 import { LitElement, html, css } from 'lit';
 import { ADOBE_PRODUCTS } from '../constants/adobe-products.js';
+import './rte-field.js';
```
honstar: This causes a new circular import among the RTE modules: `rte-field.js` imports `rte-mnemonic-editor.js` and vice versa. While this doesn't seem to be a major issue at least functionally, it is a latent fragility imo which should be addressed at some point. Not blocking at this point.

honstar (changes requested):

studio/test/mas-toolbar.test.js
```diff
@@ -18,10 +18,8 @@ describe('MasToolbar – openCreateDialog', () => {
 
     it('returns early without modifying state when mas-create-dialog is already registered', async () => {
```
honstar: Nit: the test name is now stale since state _is_ modified now :) Not blocking though.

honstar (approved):

3ch023 (approved):

afmicka:
@mirafedas The collection editor is not loading Default and Selected icon fields.
https://mwpw-444444--mas--adobecom.aem.live/studio.html#page=content&path=sandbox&query=342a9da6-2ad9-499f-9b28-49f7025aa8cf&tags=mas%3Astudio%2Fcontent-type%2Fmerch-card-collection

Screenshot 2026-07-08 at 12 56 30

see prod: 
Screenshot 2026-07-08 at 12 56 54


i get these timeout errors on your branch.
Screenshot 2026-07-08 at 13 17 19

mirafedas:
@afmicka thanks, fixed 👍

afmicka (approved):

PR #987, opened 2026-06-18
MWPW-199041: bulk-edit find IO action

yesil:
## Summary

Adds the **bulk-edit-find** Adobe I/O Runtime API — the Find stage of M@S bulk Find & Replace. Studio (MWPW-191300) and operators can start a cross-fragment text search, poll job progress, download match exports (JSON + CSV), upload an edited CSV subset, and remove the upload to restore full results. **Replace** lands separately in [MWPW-199042 / PR #999](https://github.com/adobecom/mas/pull/999).

Search targets **card and collection** fragments only — dictionary entries are excluded.

Two-stage search: Odin CF Search API does **recall** (`path` + `fullText` + optional `tags`/`status`, cursor-paged via `fetchOdin`), then the find worker does **precision** matching via `SCOPE_FIELDS`. `*` scans scoped fields only (not arbitrary CF fields). Fragment **title** is searchable (`fragmentTitle` scope or `*`); fragment **name** is never matched.

## API

**Base URL:** `https://14257-masstudio-yesil.adobeioruntime.net/api/v1/web/MerchAtScaleStudio/bulk-edit-find`

Auth: `Authorization: Bearer <mas-studio IMS token>` (see `io/studio/.env` → `MAS_TOKEN`)

| Method | Path / body | Purpose |
|--------|-------------|---------|
| `POST` | JSON `{ find, replace, surface, … }` | Start or reuse an async find job → `202 { jobId, reused }` |
| `GET` | `?jobId=` | Poll progress; when `exportReady`, body includes presigned `exports.json` + `exports.csv` URLs |
| `POST` | `?jobId=` + CSV body (multipart or `text/csv`) | Upload edited CSV; filters exports to uploaded rows |
| `DELETE` | `?jobId=` | Remove uploaded CSV; restore full exports |

### POST find parameters

- **Required:** `find`, `replace`, `surface` (`replace` is captured into the job identity but **not applied** in this PR — the Replace stage performs it in [PR #999](https://github.com/adobecom/mas/pull/999))
- **Optional:** `searchIn` (`*` default, or `prices`, `ctas`, `calloutText`, `productDescription`, `promoText`, `subtitle`, `fragmentDescription`, `fragmentTitle`, `tags`, or array of scopes), `locale` (string or array), `tags`, `status`, `matchCase`, `forceRefresh`, `supersedes` (cancels a prior RUNNING job)

### GET poll response

```json
{
  "jobId": "abc…",
  "type": "find",
  "status": "DONE",
  "done": true,
[...]

PR #999, opened 2026-06-22
MWPW-199042: Bulk-edit CSV replace worker

yesil:
Adds CSV-driven bulk-edit to M@S Studio's bulk-edit IO action: export find results to CSV, re-upload an edited CSV, and run an async replace worker that applies the changes (with dry-run + report). Builds on the bulk-edit find action (MWPW-199041, #987).

Full API, job lifecycle, CSV format, and logging/Splunk details: [`io/studio/src/bulk-edit/README.md`](https://github.com/adobecom/mas/blob/MWPW-199042/io/studio/src/bulk-edit/README.md)

Resolves https://jira.corp.adobe.com/browse/MWPW-199042

Test URLs:

- After: https://mwpw-199042--mas--adobecom.aem.live/


io/studio/src/bulk-edit/bulk-edit.js
```diff
@@ -449,52 +633,134 @@ async function handleGet(params) {
...
+                const body = parseOwBody(params);
+                body.allowedClientId = params.allowedClientId;
+                body.odinEndpoint = params.odinEndpoint;
+                body.__ow_headers = params.__ow_headers;
+                if (mode === 'find') {
+                    body.type = 'find';
+                    return await handlePost(body);
+                }
+                body.type = 'replace';
+                return await handleReplacePost(body, getBearerToken(params));
```
honstar, line 691: There is a security issue here: calling `handleReplacePost` goes through unauthenticated. Please add the same `isAllowed` auth check to `handleReplacePost` that has already been added to all the other `handle*` functions.

io/studio/src/bulk-edit/bulk-edit.js
```diff
@@ -271,19 +310,97 @@ async function handlePost(params) {
     return { statusCode: 202, body: { jobId, reused: false } };
 }
 
-function isTerminalJob(job) {
+const TERMINAL_REPLACE_STATUSES = new Set(['DONE']);
+
+async function handleReplacePost(params, authToken) {
+    if (!params.odinEndpoint) return errorResponse(400, 'missing parameter(s) odinEndpoint', logger);
```
honstar, line 318: As mentioned above, add the `isAllowed` check here:
```
if (!(await isAllowed(authToken, params.allowedClientId))) {
    return errorResponse(401, 'Authorization failed', logger);
}
```

io/studio/src/bulk-edit/replace-worker.js
```diff
@@ -0,0 +1,308 @@
...
+
+    let results = job.results || [];
+    let cursor = job.cursor || 0;
+    logger.info(JSON.stringify({ event: 'bulk-edit-replace-worker-start', jobId, total: items.length, dryRun, cursor }));
+
+    try {
+        for (let i = cursor; i < items.length; i += batchSize) {
+            const batch = items.slice(i, i + batchSize);
+
+            const batchResults = await Promise.all(
```
honstar, line 255: Consider using the existing `processBatchWithConcurrency` helper function from `common.js` that respects `rpsLimit` in addition to `batchSize`. You're already passing `rpsLimit` as [action input parameter](https://github.com/adobecom/mas/pull/999/changes#diff-6a102ac1ec16973f90199650b10fe4557969f92c0020f85e3d54ca2a0bbd97dbR77) but not using it here, it's basically dead config.
This is a more defensive approach rather than a reactive one where you handle 429s using the existing backoff mechanism. A single rate-limited PATCH request that is persistently rate-limited could stall the whole batch for many minutes.

honstar (dismissed):

io/studio/src/bulk-edit/replace-worker.js
yesil: Good catch on the dead config — fixed. I added the proactive `rpsLimit` throttle, but inlined the formula rather than swapping in `processBatchWithConcurrency`.

Reason: that helper is a run-to-completion map (`for i=0…end`, no early exit). This worker loop is resumable + interruptible — it starts at `cursor`, returns early on cancel, and re-invokes a sibling action when the soft budget is hit (IO Runtime wall-clock limit). The helper's only hook is `onBatchCompleted`, a fire-and-forget callback that can't break or resume, so using it would mean re-PATCHing already-processed fragments on a continued run.

So I reused the invariant (the rate formula), not the loop:

```js
const minBatchMs = rpsLimit ? (batchSize / rpsLimit) * 1000 : 0;
// ...after the stop/soft-budget checks, only when the loop actually continues:
if (minBatchMs > 0) {
    const wait = minBatchMs - (Date.now() - batchStart);
    if (wait > 0) await sleep(wait);
}
```

If the duplicated formula bothers us, happy to extract `computeMinBatchMs(batchSize, rpsLimit)` into `common.js` and have both call sites share it.

io/studio/src/bulk-edit/replace-worker.js
```diff
@@ -0,0 +1,319 @@
...
+
+function etagMismatchResult(item, csvEtag, serverEtag) {
+    return buildResult(item, 'SKIPPED', {
+        reason: 'etag_mismatch',
+        csvEtag,
+        serverEtag,
+    });
+}
+
+async function patchOrThrow(odinEndpoint, id, authToken, patchBody, etag) {
```
honstar: Odin is using a static 60s cooldown once rps limits are exceeded and 429 responses are sent for a user-agent. The backoff implemented here (1/2/4/8s) will always land in the 60s cooldown window and thus get re-rejected on retry. While `fetchOdin` already respects the 60s cooldown (and in fact using 65s as safety buffer), this outer layer would only burn attempts and exhaust them.
There is no shared cooldown gate, this function is called with a 10-wide batch that runs concurrently: a single 429 freezes the whole client but all ~10 in-flight requests 429 together, all sleep ~65s in `fetchOdin` then wake and fire simultaneously again, potentially tripping the lockout again.
Instead, there should be a worker-wide cooldown - on _any_ 429, set a cooldown of 60s + some jitter and have every `processFragment` await it before issuing a `GET` or `PATCH`, and drop the inner loop in this function to have a single place where 429 is handled.

honstar (changes requested):

io/studio/src/bulk-edit/replace-worker.js
yesil: Fixed in ae2e486c2 — replaced patchOrThrow's per-fragment retry loop with a worker-wide cooldown gate:

- `createRateLimitGate()`/`tripRateLimitGate()`/`awaitRateLimitGate()` in replace-worker.js hold a single `{ cooldownUntil }` shared across one worker invocation.
- Both `getFragmentWithEtag` and `patchToOdin` calls in `processFragment` now go through `runGated(gate, fn)`: await any active cooldown first, and on a 429 trip a fresh 60s+jitter cooldown that every other fragment (this batch and later ones) will see and wait out.
- Dropped the old 1/2/4/8s exponential backoff entirely — you're right that it always landed inside Odin's 60s window and just burned attempts. A persistent 429 now fails that one fragment immediately (same as before once retries exhausted) instead of retrying pointlessly, but the shared gate means the *next* fragment doesn't independently rediscover the same 429 and potentially re-trip the lockout.

Added a deterministic test (`batchSize: '1'` to force sequential processing) that asserts the second fragment waits ≥60s via the gate before Odin is hit again, plus unit tests on the gate mechanics directly.

io/studio/app.config.yaml
```diff
@@ -59,6 +68,18 @@ application:
                         annotations:
                             require-adobe-auth: false
                             final: true
+                    bulk-edit-replace-worker:
+                        function: src/bulk-edit/replace-worker.js
+                        runtime: nodejs:22
+                        inputs:
+                            odinEndpoint: $ODIN_ENDPOINT
```
honstar: This is already defined on package level and can be reused in all actions.

io/studio/app.config.yaml
honstar: I didn't catch this in the bulk-find PR: this is already defined on package level above and doesn't need to be redefined on action level, that is redundant.

honstar (changes requested):

honstar:
I noticed that all requests to odin sent eventually via `fetchOdin` are defaulting to the `mas-translation-project` user-agent. I don't think we should be doing that and instead, use a separate `mas-bulk-edit` for example for all odin requests.

I'd suggest to log a follow-up story to extend both `patchToOdin` and `getFragmentWithEtag` to allow passing in an optional `user-agent` string. There are probably additional odin request helper functions that would need to be extended.

yesil:
Thanks @honstar, i address your feedback in the commit: https://github.com/adobecom/mas/pull/999/changes/beb1ededbdc844f173cafef595ba07e080f965f7

honstar (approved):
LGTM, thanks for bearing with me :)

3ch023 (approved):

PR #1005, opened 2026-06-25
MWPW-199352: Support pasting fragment links in translation project

Andrei4226:
Resolves https://jira.corp.adobe.com/browse/MWPW-199352

Import via URL — dedicated panel for adding fragments by pasting Studio deep links.
Figma: https://www.figma.com/design/GDDRLo3S7fz0SMRefpJOpx/M-Studio?node-id=9975-159933&t=5vG9XXCIBGdtcUf0-1

A new "Import via URL" button appears in the item selector header.
Clicking it opens a dedicated panel with a textarea for pasting URLs and a table showing each URL's resolution status.

The table displays the fragment display name, a status chip (Loading / Validated / Error) and a per-row Remove button. A footer row shows the total count and a Remove All action. The right panel continues to show the current selection in real time.

Supported URL format: Studio deep links (https://mas.adobe.com/studio.html#content-type=...&query=<uuid>).
E.g: https://mas.adobe.com/studio.html#content-type=merch-card&page=content&path=sandbox&query=c075f0c7-a08d-4e42-8d45-03a5169dd877

What each editor accepts:
- Translation editor: merch-card, merch-card-collection, mas-compare-chart — restricted to fragments whose surface matches the active project's surface
- Compare Chart editor: merch-card only (max 4) — restricted to fragments whose surface matches the active search surface

Status:
- Validated (green): fragment found, surface allowed and added to selection
- Error (red):
  - Fragment not found.: fetch failed / fragment doesn't exist
  - Type not allowed here.: URL content-type not accepted in this editor
  - {Surface} not allowed here.: fragment belongs to a different surface than the current project (e.g. pasting a Nala fragment into a Sandbox translation project shows "Nala not allowed here.")
  - Unsupported fragment type.
  - Max 4 fragments reached.
- Loading: fetch in progress

When at least one item is invalid, a warning banner appears above the URL input:
`"N item(s) are invalid and won't be added. Cancel to review the links, or continue to add only the valid items."
`
Toasts:
- "Fragment added" / "N fragments added": all resolved and added successfully
- "Fragment not found" / "N fragments not found": failed for any error reason above
- "N added, N not found": mixed result
- "No valid URLs found": nothing parseable was pasted (bad format, missing UUID, unsupported content-type param)
- "Already added" / "N already added" — pasted URL(s) already exist in the current import list

translation project:
image

mas-compare-chart:
image

## 🧪 Nala E2E Tests

### To run Nala tests again:

- Remove the `run nala` label

## Test URLs:

- After: https://mwpw-199352--mas--adobecom.aem.live/


mirafedas (changes requested):
Looks great, but can we please make the 'No items selected yet' and loading while verifying the links a part of mas-selected-items.js (extend it)?
Also, when I remove the item in mas-selected-items, it does not get removed from the Import from URL table

Andrei4226:
> remove the item in mas-selected-item

Addressed all points:
- Moved "_No items selected yet_" and loading state into `mas-selected-items.js`
- Fixed the sync bug, removing an item via `mas-selected-items `now dispatches a `selected-item-removed` event that updates the Import URL table
- Changed "_Import via URL_" to behave like a tab (as Fragments/Collections/Placeholders) and second click no longer closes the view

mirafedas (approved):

3ch023 (changes requested):
works very well, but one finding is worth addressing.
could you please add validation for pasted resources.
it should prevent pasting of nala fragments into sandbox translation projects.
pls check if we can somehow make this validation configurable, because we might later reuse this import window in promos and there we would need to filter out variations (while in translation projects they are allowed)

please keep invalid items shown in the dialog, with some 'Invalid' red status and suggest to user to "cancel" dialog or "proceed only valid paths"

studio/src/common/components/mas-items-selector.js
```diff
@@ -128,27 +120,214 @@ class MasItemsSelector extends LitElement {
...
+                this.#setImportedUrlStatus(item.fragmentId, 'error', 'Unsupported fragment type.', path, displayName);
+                continue;
+            }
+            const appended = isCard ? this.#appendSelectedCard(display) : this.#appendSelectedCollection(display);
+            if (!appended) {
+                failed++;
+                this.#setImportedUrlStatus(
+                    item.fragmentId,
+                    'error',
+                    `Max ${this.maxSelectedCards} fragments reached.`,
```
yesil, line 284: `#appendSelectedCard` returns `false` both when the path is already in the selection and when `maxSelectedCards` is hit, so this branch always labels the failure "Max N fragments reached." A fragment that is already selected (e.g. picked via search, then its link pasted here — import mode clears `importedUrls` but not `selectedCards`) therefore shows a misleading "Max reached" chip instead of an "already added" state. Consider distinguishing the two cases so the row reports the real reason.

studio/src/common/components/mas-items-selector.js
```diff
@@ -221,6 +400,179 @@ class MasItemsSelector extends LitElement {
...
+        const invalidCount = this.importedUrls.filter((i) => i.status === 'error').length;
+        if (!invalidCount) return nothing;
+        return html`<p class="import-invalid-warning">
+            <sp-icon-alert></sp-icon-alert>
+            ${invalidCount} ${invalidCount === 1 ? 'item is' : 'items are'} invalid and won't be added. Cancel to review the
+            links, or continue to add only the valid items.
+        </p>`;
+    }
+
+    #renderImportPanel() {
```
yesil, line 489: ideally these methods should be getters:

get #importPanel()

yesil (approved):

3ch023 (approved):

PR #1007, opened 2026-06-26
MWPW-189495: [M@S CI/CD] Trivial PR

mirafedas:
Resolves https://jira.corp.adobe.com/browse/MWPW-189495

## 🧪 Nala E2E Tests

### To run Nala tests again:

- Remove the `run nala` label

## Test URLs:

- After: https://mwpw-189495--mas--adobecom.aem.live/


.github/workflows/run-nala.yml
```diff
@@ -27,12 +62,12 @@ jobs:
...
         name: Running Nala Studio E2E UI Tests
-        needs: nala-gate
-        # Run when PR is opened (no label needed) OR when 'run nala' label exists
-        # This covers: opened (always), synchronize (if label exists), labeled (if label exists)
+        needs: [nala-gate, check-paths]
         if: |
-            github.event.action == 'opened' ||
-            contains(github.event.pull_request.labels.*.name, 'run nala')
+            needs.nala-gate.result == 'success' &&
+            needs.check-paths.outputs.readme-only != 'true' &&
```
3ch023: maybe we could use !needs.check-paths.outputs.readme-only instead.
because README.md are not the only trivial files, there are .yml, .claude/*** ones, .gitignore ones. its better if the list of trivial files is extensible and naming is generic

.github/workflows/run-nala.yml
```diff
@@ -129,20 +164,23 @@ jobs:
...
+              run: git fetch origin "${BASE_REF}"
 
             - name: Detect io/www changes
               id: filter
+              env:
+                  BASE_REF: ${{ github.event.pull_request.base.ref }}
               run: |
-                  if git diff --quiet origin/${{ github.event.pull_request.base.ref }}...HEAD -- io/www/ 2>/dev/null; then
-                      echo "io-www=false" >> $GITHUB_OUTPUT
+                  git diff --quiet "origin/${BASE_REF}...HEAD" -- io/www/ 2>/dev/null
```
3ch023, line 167: why removing `if` here?

.github/workflows/run-nala.yml
```diff
@@ -129,20 +164,23 @@ jobs:
                   fetch-depth: 0
 
             - name: Fetch base branch for diff
-              run: git fetch origin ${{ github.event.pull_request.base.ref }}
+              env:
+                  BASE_REF: ${{ github.event.pull_request.base.ref }}
```
3ch023, line 159: this constant doesn't seem to bring any value since its not reused anywhere

3ch023 (changes requested):
in addition to comments on pr:

1. Duplicated README-detection logic. we need to find a way to avoid duplication since it could lead to problems in future.
2. README-detection should be changed to 'trivial changes detection'. README are not the only trivial changes, there will be more files and the list should be extendable.

.github/workflows/run-nala.yml
mirafedas: This constant is needed to avoid the script injection vulnerability

.github/workflows/run-nala.yml
mirafedas: Because the diff result is being saved in the `status`, and then we have the `if else` statements depending on the `status` value

3ch023 (approved):

npeltier (approved):

npeltier:
looks like a fail @mirafedas 
image

PR #1010, opened 2026-06-26
MWPW-199670 fix auto deploy

npeltier:
Resolves https://jira.corp.adobe.com/browse/MWPW-199670

### DIFF
--- a/.github/workflows/io-deploy.yaml
+++ b/.github/workflows/io-deploy.yaml
@@ -21,9 +21,9 @@ jobs:
             - name: Setup Node.js
               uses: actions/setup-node@v4
               with:
-                  # Pinned: Node 22.23.0 breaks node-fetch v2 gzip handling in aio-cli,
-                  # causing ERR_STREAM_PREMATURE_CLOSE when minting IMS tokens on deploy.
-                  node-version: '22.22.0'
+                  # Pinned: 22.23.1 fixes the node-fetch v2 gzip / ERR_STREAM_PREMATURE_CLOSE
+                  # regression introduced in 22.23.0.
+                  node-version: '22.23.1'
                   cache: 'npm'
                   cache-dependency-path: io/www/package.json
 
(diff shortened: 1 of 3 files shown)


.github/workflows/io-deploy.yaml
```diff
@@ -21,7 +21,9 @@ jobs:
             - name: Setup Node.js
               uses: actions/setup-node@v4
               with:
-                  node-version: '22.x'
+                  # Pinned: Node 22.23.0 breaks node-fetch v2 gzip handling in aio-cli,
+                  # causing ERR_STREAM_PREMATURE_CLOSE when minting IMS tokens on deploy.
+                  node-version: '22.22.0'
```
honstar: Please pin to 22.23.1 instead as this has the fix for the network issue.

Andrei4226 (approved):

yesil (approved):

yesil:
This pin fixes `io-deploy.yaml` (io/www) but **`io-studio-deploy.yaml` still uses unpinned `node-version: 22`**, so it resolves to 22.23.0 and keeps hitting the same `ERR_STREAM_PREMATURE_CLOSE` on IMS token mint.

Every `io/studio` PR deploy is still red — confirmed on [run 28427335256](https://github.com/adobecom/mas/actions/runs/28427335256/job/84235339136) (Node `v22.23.0`, fails at `Gunzip.<anonymous>` minting the IMS token), and identically on MWPW-194144 / MWPW-199042 / MWPW-194867.

Same one-line fix needed in `io-studio-deploy.yaml` (the deploy job at line ~70 is the one that mints the token; line ~47 tests job for parity):

```yaml
node-version: '22.22.0'
```

Could fold it into this PR so the studio path is covered too.

yesil (changes requested):
please address the above findings.

yesil (approved):

honstar (approved):

PR #1018, opened 2026-07-01
MWPW-199358 promo permissions

npeltier:
- enable promotions for GRP-ODIN-MAS-PROMO-EDITORS members and admin only,
- for the others, make read only UI

Resolves https://jira.corp.adobe.com/browse/MWPW-199358

## Test URLs:

- After: https://MWPW-199358--mas--adobecom.aem.live/studio.html

### DIFF
--- a/studio/src/promotions/mas-promotions-editor.js
+++ b/studio/src/promotions/mas-promotions-editor.js
@@ -244,20 +244,20 @@ class MasPromotionsEditor extends LitElement {
     }
 
     get canManagePromoCodes() {
-        if (!canEditPromotions()) return false;
+        if (!this.canEdit) return false;
         const geos = this.fragment?.getFieldValues('geos') ?? [];
         const hasOffers = Store.promotions.selectedOffers.value.length > 0 || Store.promotions.selectedCards.value.length > 0;
         return hasOffers && geos.length > 0;
     }
 
     get canManagePromoCodesInEmptyState() {
-        if (!canEditPromotions()) return false;
+        if (!this.canEdit) return false;
         const geos = this.fragment?.getFieldValues('geos') ?? [];
         return this.hasSelectedOffers && geos.length > 0;
     }
 
     get canEditPromotionItemsInEmptyState() {
-        return canEditPromotions() && this.hasSelectedOffers;
+        return this.canEdit && this.hasSelectedOffers;
     }
 
     #mapPromotionOfferSelectorToRow(selectorId) {
@@ -805,7 +805,7 @@ class MasPromotionsEditor extends LitElement {
 
     get disabledPromotionQuickActions() {
         const disabled = new Set([QUICK_ACTION.LOCK]);
-        if (!canEditPromotions()) {
+        if (!this.canEdit) {
             PROMOTION_QUICK_ACTIONS.forEach((action) => disabled.add(action));
             return disabled;
         }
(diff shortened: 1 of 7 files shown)


Andrei4226 (approved):
LGTM

- I can't create a new promo project if I don't have access (the "Create promotion project" button is hidden)
- I can open an existing promo project in view mode. The fields are still editable (not disabled), but I can't save so it's fine (a minor UX mismatch)

studio/src/promotions/mas-promotions.js
```diff
@@ -235,10 +236,12 @@ class MasPromotions extends LitElement {
             <div class="promotions-container">
                 <div class="promotions-header">
                     <sp-search size="m" placeholder="Search"></sp-search>
-                    <sp-button variant="accent" @click=${() => this.#handleAddPromotion()} class="create-button">
-                        <sp-icon-add slot="icon"></sp-icon-add>
-                        Create promotion project
-                    </sp-button>
+                    ${canEditPromotions()
```
honstar: In case `Store.users` resolves after this component has been rendered, it can happen that an eligible promo-user won't be able to see this gated component. Consider adding `Store.users` into the reactive dependency list.

studio/src/promotions/mas-promotions-editor.js
honstar, line 193: Add `Store.users` here to enable a re-render in case the async call to fetch users resolves _after_ components are rendered in the editor.

honstar:
The promotions editor itself is not gated behind the `canEditPromotions` check which means fields are still editable but cannot be saved since the corresponding quick actions are disabled. Also, changing any value would trigger the `Discard changes?` prompt.

studio/src/promotions/mas-promotions-editor.js
```diff
@@ -1409,7 +1430,11 @@ class MasPromotionsEditor extends LitElement {
                                       </h2>
                                       <div>
                                           ${this.selectedItemsViewTab === TABLE_TYPE.OFFERS
-                                              ? html`<sp-action-button quiet @click=${this.#openPromotionsOst}>
+                                              ? html`<sp-action-button
+                                                    quiet
+                                                    ?disabled=${!canEditPromotions()}
```
yesil: `render()` already computes `readOnly`/`canEdit` at the top of this method (~L1267-1268). Reuse `readOnly` here instead of re-calling `canEditPromotions()`, which re-runs a linear `Store.users.get().find(...)` + `.map()` scan (`groups.js:26-28`) for a value that's stable for the whole render. Same applies to the getters at L247/L254/L260/L808 and the helper renderers `#renderPromotionOffersEmptyPanel` (L1066) / `#renderPromotionEmptyToolbarActions` (L1107) — promoting to a single `get canEditPromotions()` lets all of them share one computation and keeps the predicate spelled one way.

studio/src/promotions/mas-promotions.js
```diff
@@ -306,6 +310,18 @@ class MasPromotions extends LitElement {
     }
```
yesil, line 315: `renderActionCell` runs `canEditPromotions()` once per promotion row, and each call is a full linear `Store.users` scan + `.map()` (`groups.js:26-28`). For N rows that's N scans per render. Hoist to a single `const canEdit = canEditPromotions()` in `render()` (or a shared getter) and read it here.

yesil (approved):
See small perf related comments.

yesil:
@npeltier as a readonly user I see one can add an offer:

image

PR #1028, opened 2026-07-06
MWPW-199681 and MWPW-192879 Global settings fixes

bozojovicic:
Resolves https://jira.corp.adobe.com/browse/MWPW-199681
and https://jira.corp.adobe.com/browse/MWPW-192879

**MWPW-199681** - When "Secure transaction" is enabled in global settings in default node for some templates, and we enable it in card editor for some card with another template - it did not work. - Now it does.
**MWPW-192879** - When "Secure transaction" is enabled in global settings in default node for some templates and if we create override node, this somehow enabled it for all card templates. - Now it is restricted to selected templates only.

## 🧪 Nala E2E Tests

### To run Nala tests again:

- Remove the `run nala` label

## Test URLs:

- After: https://mwpw-199681--mas--adobecom.aem.live/


io/www/src/fragment/transformers/settings.js
```diff
@@ -166,7 +166,21 @@ export function resolveSettingEntry(fragment, locale, setting) {
     const defaultEntry = setting.default;
     if (!defaultEntry) return null;
     const template = fragment.fields?.variant;
-    if (defaultEntry.templates?.length > 0 && !defaultEntry.templates.includes(template)) return null;
+    if (defaultEntry.templates?.length > 0 && !defaultEntry.templates.includes(template)) {
+        const definition = SETTING_NAME_BY_VALUE.get(defaultEntry.name);
+        const fragmentValue = fragment.fields[definition?.propertyName || definition?.name];
+        if (typeof fragmentValue !== 'undefined') {
+            const isBoolean = 'boolean' === typeof fragmentValue;
```
3ch023: there is that normalizeBoolean method, which might give more flexibility in case fragmentValue will come as string 'true'. but not sure if it ever can be the case

3ch023 (commented):
Hi @bozojovicic , could you double check if new Unit Test enters the if branch for card override.
claude is complaining that tests are checking 'plans' fragment against 'plans' setting, but to enter if we need a template mismatch

bozojovicic:
> Hi @bozojovicic , could you double check if new Unit Test enters the if branch for card override. claude is complaining that tests are checking 'plans' fragment against 'plans' setting, but to enter if we need a template mismatch

Claude is confused. :) In unit tests, in settings-sandbox.json the quantity selector is configured for the product template only. I check it for the plans card and in that plans card the QS value is overridden on the card level. So it enters the IF block.

io/www/src/fragment/transformers/settings.js
bozojovicic: `const isBoolean = 'boolean' === typeof normalizeBoolean(fragmentValue);`

done

3ch023 (approved):

TsayAdobe (approved):

Roycethan (approved):

PR #1030, opened 2026-07-08
MWPW-198125: preserve cancel-context override when saving from new OST

Andrei4226:
Resolves https://jira.corp.adobe.com/browse/MWPW-198125

## Problem
In the new OST (`?ost=new`), clicking "Use" after setting the promo override to `cancel-context` saved `effectivePromoCode` instead of the raw override.
`effectivePromoCode` always resolves to `undefined` when a promo is cancelled, so the cancellation was silently dropped — the price kept applying the context promo and reopening OST showed the override as empty.

## Fix
In `ost-code-output.js`, when the stored override equals the literal `cancel-context` value, persist that value directly instead of `effectivePromoCode`.

## Testing
- Added a test asserting `app.select` receives `promoOverride: 'cancel-context'` when the context promo is cancelled.
- Companion fix for the legacy (default) OST tracked separately in https://git.corp.adobe.com/wcms/tacocat.js 
  PR link: https://git.corp.adobe.com/wcms/tacocat.js/pull/680

Note: the compiled `ost-new.js` bundle is not included in this PR.

I included in this PR the legacy changes (for `JS` and `CSS`) along with the new `index.js` file from `tacocat.js` (in accordance with the changes from https://git.corp.adobe.com/wcms/tacocat.js/pull/680).

Before (without `data-promotion-code` for `cancel-context`):

image

After (with `data-promotion-code` for `cancel-context`):

image

Persistence on saving for `cancel-context`:

image

## 🧪 Nala E2E Tests

### To run Nala tests again:

- Remove the `run nala` label

## Test URLs:

- After: https://mwpw-198125--mas--adobecom.aem.live/


ost/src/components/ost-code-output.js
```diff
@@ -156,6 +156,13 @@ export class OstCodeOutput extends LitElement {
 
         const promoStatus = computePromoStatus(store.storedPromoOverride, store.promotionCode);
 
+        // Persist the literal cancel value
+        // effectivePromoCode would lose it (collapses to undefined)
+        const promoOverride =
```
mirafedas, line 161: If we add 'context-promo' to the price or CTA that does not have any promo code (so nothing to cancel), it will add `data-promotion-code="cancel-context"` on the element. I'd expect it to add nothing, as no promo code was initially set for this price/CTA.

mirafedas (changes requested):

ost/src/components/ost-code-output.js
3ch023: this is the bug yes.
card A has no promos, but later in /mas/io it will get promo decoration.
we need an ability to always set cancel-context so that decoration that happens later doesn't occur on specific prices

3ch023: here is my test page 
https://main--milo--adobecom.aem.page/in/drafts/mariia/testpromomweb?clearCaches.io

3ch023 (approved):

mirafedas (approved):

afmicka (approved):

PR #1031, opened 2026-07-08
MWPW-198960: Create Promo tag automatically based on promo project title

mirafedas:
Resolves https://jira.corp.adobe.com/browse/MWPW-198960

## 🧪 Nala E2E Tests

### To run Nala tests again:

- Remove the `run nala` label

## Test URLs:

- Before: https://main--mas--adobecom.aem.live/studio.html#page=promotions&path=sandbox
- After: https://mwpw-198960--mas--adobecom.aem.live/studio.html#page=promotions&path=sandbox


3ch023 (changes requested):
tried on https://mwpw-198960--mas--adobecom.aem.live/studio.html#page=promotions-editor&path=sandbox&promotionId=4dfd9cb4-b55a-464f-ad28-f00e755649ec but it didn't really create a tag

Axelcureno (commented):
Nice cleanup dropping the manual picker. Two things:

- New promotion + a title that normalizes to an empty slug sets `tags` to `[]`, so save fails with "Please add at least one Promotion tag". That field is disabled, so the user can't act on it — the real problem is the title. Surface a title-specific message?
- @3ch023's "didn't create a tag" looks tested on an existing promotion, where auto-gen is gated on `isNewPromotion`. Confirm on a new one?

3ch023 (approved):

studio/src/aem/aem.js
```diff
@@ -1078,6 +1078,78 @@ class AEM {
...
+        if (response.status !== 404) {
+            throw new Error(`Failed to check tag: ${response.status} ${response.statusText}`);
+        }
+
+        const csrfToken = await this.getCsrfToken();
+        const formData = new FormData();
+        formData.append('jcr:primaryType', 'cq:Tag');
+        formData.append('jcr:title', title);
+
+        const createResponse = await fetch(`${this.baseUrl}${tagPath}`, {
```
npeltier: looks like a nice yet new way of POSTing to AEM from studio code, 
are you sure we need CSRF token here? 
If yes, it looks pretty generic, would you mind factorizing this in one method? you are already duplicating code once, and i bet lot of others could benefit from it.

npeltier (approved):
lgtm, either treat my comment, or create a JIRA :)

studio/src/promotions/mas-promotions-editor.js
```diff
@@ -853,10 +886,13 @@ class MasPromotionsEditor extends LitElement {
             },
         );
         if (!confirmed) return;
+        const tagId = getPromotionTagFromFragment(this.fragmentStore.get());
+        const [tagPath] = tagId ? fromAttribute(tagId) : [];
         try {
             showToast('Deleting promotion campaign...');
             await deleteAttachedPromoVariations(this.repository.aem, this.fragment);
             await this.repository.deleteFragment(this.fragmentStore, { startToast: false, endToast: false });
+            if (tagPath) await this.repository.aem.tags.delete(tagPath);
```
Andrei4226: If `tags.delete` throws here, the snippet is already deleted, but the code jumps to the catch block and displays 'Failed to delete promotion campaign' which is a misleading message

studio/src/promotions/mas-promotions-editor.js
Andrei4226, line 737: Not part of this PR's diff, but the new tag-creation logic above interacts with this pre-existing catch: if `createFragment` throws here, we skip the tag rollback entirely and the tag stays orphaned in AEM
Retrying with the same title then fails with "Tag already exists." 
Worth deleting the tag here too

Andrei4226 (approved):

PR #1034, opened 2026-07-09
MWPW-200064: View promo variations in Promo project fragment table view - part 1 (Select item dialog only)

mirafedas:
This PR covers the Add items dialog. The view-only table will be covered in the follow-up PR (same Jira).
After merging, please move the Jira back to 'In Development'.

Resolves https://jira.corp.adobe.com/browse/MWPW-200064

Included in the scope:
- a fix for cloning the fragment in `replaceFrom` (`fragment.js`) - it used to throw console errors because on UI we are adding fields `offerData` and `groupedVariations` that do not exist in the fragment content model and are not clonable by structuredClone. Now we exclude them from being cloned, as we do not send these fields to the backend anyway.
- props refactoring for `mas-collapsible-table-row`: the `tabs` prop controls which variation tabs we show (covers previously used `hideLocaleTab` and `disableLocaleVariations`), and `nonSelectableVariations` controls which variation types are not possible to select (covers previously used `disableGroupedVariationSelection`, making it more universal). This hides the `Locale` tab in the translations editor, which was not used (and not planned to be used) anyway.

When testing, please check the mini-compare-chart and translation editors as well.

## 🧪 Nala E2E Tests

### To run Nala tests again:

- Remove the `run nala` label

## Test URLs:

- After: https://mwpw-200064--mas--adobecom.aem.live/


studio/src/common/components/mas-items-selector.js
```diff
@@ -34,9 +34,8 @@ class MasItemsSelector extends LitElement {
         maxSelectedCards: { type: Number, attribute: 'max-selected-cards' },
         lockedTemplateFilter: { type: String, attribute: 'locked-template-filter' },
         defaultTemplateFilter: { type: String, attribute: 'default-template-filter' },
-        disableGroupedVariationSelection: { type: Boolean, attribute: 'disable-grouped-variation-selection' },
-        hideLocaleTab: { type: Boolean, attribute: 'hide-locale-tab' },
-        disableLocaleVariations: { type: Boolean, attribute: 'disable-locale-variations' },
+        nonSelectableVariations: { type: Array, attribute: 'non-selectable-variations' },
```
3ch023, line 37: kudos on refactoring.
just a suggestion, but we could standartise and have 'selectableVariations' and 'visibleVariations'

not bound to exact naming, but rather to the fact that in both cases we have 'allow-list'. e.g. what can be seen and what can be selected.

and right now its a bit confusing because for what can be seen we have 'allow-list', but for 'what can be chosen' we have deny list

3ch023 (commented):
left a suggestion

studio/src/common/components/mas-select-items-table.js
```diff
@@ -28,11 +28,9 @@ class MasSelectItemsTable extends LitElement {
         maxSelectedCards: { type: Number },
         getDisplayName: { type: Function },
         renderFragmentStatusCell: { type: Function },
-        disableCardExpansion: { type: Boolean },
-        disableGroupedVariationSelection: { type: Boolean },
-        hideLocaleTab: { type: Boolean },
-        disableLocaleVariations: { type: Boolean },
+        nonSelectableVariations: { type: Boolean },
```
yesil: **[nit]** `nonSelectableVariations` is declared `{ type: Boolean }` here, but it's an `Array` everywhere else in the chain (`mas-items-selector`, `mas-collapsible-table-row`). Harmless today since it's only set via `.property` binding (the type converter is skipped), but should be `{ type: Array }` to avoid confusion.

yesil (approved):
Nice cleanup — swapping those four boolean flags for the tabs + nonSelectableVariations model reads a lot better, and I like that the tab keys are centralized now. Left one small thing inline.

Andrei4226 (approved):
LGTM
Now it's clearer where we can find each promo variation in the Select Items window along with that hyperlink for the specific promo project 👍

studio/test/translation/mas-collapsible-table-row.test.js
```diff
@@ -713,131 +700,6 @@ describe('MasCollapsibleTableRow', () => {
...
-                    .topLevelCard=${topLevelCard}
-                    .isTopLevelExpanded=${true}
-                ></mas-collapsible-table-row>`,
-            );
-            await el.updateComplete;
-            expect(listStub.called).to.be.true;
-        });
-    });
-
-    describe('disableGroupedVariationSelection', () => {
```
honstar, line 785: This deleted test block should be replaced with equivalent tests for `nonSelectableVariations` - the actual "view but don't select" rendering logic this PR introduces is actually untested.

honstar (changes requested):

honstar (approved):

afmicka (approved):

PR #1037, opened 2026-07-10
MWPW-199953: Create geo-specific promo variations

Andrei4226:
Resolves https://jira.corp.adobe.com/browse/MWPW-199953

### Scope: Add ability to create geo-specific promo variations

- Create promo variation now opens a "Select geos" step (with the `geos` selected in promo project), then a confirmation step explaining what will be created
- The functions that resolved the 'parent' path of a promo variation (`resolvePromoVariationParentPath`, `resolveDefaultPathFromPromoVariation`) previously returned a single string (or `null` if they didn't find anything). Now they return a `string[]`,  an ordered list of possible paths, because support for multiple variations on geo introduces paths with suffixes (`my-card-2, my-card-3`) and it can no longer be known in advance which one is correct without checking them.
- You can create more than one promo variation per card, as long as each one uses different `geos` (no overlap allowed)
- Geo tags are shown wherever you'd expect them: the create dialog, the fragment editor and the fragments table under a card's "Promotion" tab
- Deleting a promo variation now takes you back to the promotion project, not the general fragments list
- "Create promo variation" only shows up when the project still has unused `geos` left for that card

## New features added

- **Symmetric unpublish:** Unpublishing a promo project now warns first if it has live promo variations and unpublishes them together with the project — mirroring the existing publish-side warning. Before, only the project's own config was unpublished, silently leaving variations live on the site.
- `unpublishPromotionProject()` (new, `promotion-publish-utils.js`) unpublishes the project + each individual variation using `getWithEtag/unpublish`, with shortfall counting (variations that could not be unpublished) and a differentiated toast message.
- Refactor: `confirmPublishDespiteUnpublishedPromoVariations` and the new `confirmUnpublishAlongsidePromoVariations` are now both generated from an internal generic function `confirmActionAgainstPromoVariations(...)` (code deduplication for `publish/unpublish`).
- **Cascading delete:** Deleting a promo project now also deletes its attached promo variations (the root cause of the bug above), with the confirm dialog now stating how many variations will be removed too.
- **Consistent geo-restriction handling:** Legacy promo variations created without any geo restriction are now correctly treated as covering all geos, blocking new variation creation for that card as expected instead of being silently ignored.
- **Multi-variation awareness:** Viewing, publishing and unpublishing now correctly account for every promo variation a card has, not just the first — fixing stale "variation was deleted" errors and silent gaps in `publish/unpublish` warnings when a card has more than one variation.
- `createKeyedAsyncLoader()` in `utils.js` — memoization helper for reactively triggered async fetches (runs `load()` only if `computeKey()` has changed compared to the last run, otherwise skips). Used by 3 new loaders: `#promotionGeoOptionsLoader`, `#disabledPromoGeoOptionsLoader` (`mas-fragment-editor.js`) and `#promotionGeosFallbackLoader` (`mas-fragment-variations.js`) — avoids redundant geo re-fetches on every re-render.

## Fix: promo variations tab wasn't showing more than one variation per project

`probePromoVariationReferences` only checked the unsuffixed path (e.g. `.../promotions/cyber-monday/my-card`) for each promotion project, so a second or third geo-specific variation for the same card and project (`.../my-card-2`, `.../my-card-3`) was silently ignored — it never showed up under the "Promotion" tab in the fragment list, no matter how long you waited for indexing.

Now it probes every suffixed index per project (same sequential prober already used for `create/view/publish/unpublish`), so all variations for a card under a given promotion tag are correctly discovered and displayed, not just the first one.

## IO changes

- io/www now supports geo-scoped promo variations
- Legacy variations = "all geos" (an old variation, with no saved `pznTags`, is treated as covering all current `geos`of the promo project)
– completely hides the "Create promo variation" button if a legacy variation already exists.
– The display fallback (`mas-fragment-variations.js`) no longer picks the first project found by tag, but also checks that the project actually contains the fragment, so it doesn't show the geos of the wrong project (until the https://github.com/adobecom/mas/pull/1031 will be merged)

!!After this PR, the user can edit `geos` field for promo variation ONLY from promo project.

## Regression

Preview

- non regression on us plans page https://main--cc--adobecom.aem.page/creativecloud/plans?maslibs=MWPW-199953
- non regression on fr plans page https://main--cc--adobecom.aem.page/fr/creativecloud/plans?maslibs=MWPW-199953
- non regression on catalog page https://main--cc--adobecom.aem.page/products/catalog?maslibs=MWPW-199953

mas/io

- non regression on us plans page https://main--cc--adobecom.aem.live/creativecloud/plans?mas-io-url=https://14257-merchatscale-andrei4226.adobeioruntime.net/api/v1/web/MerchAtScale
- non regression on fr plans page https://main--cc--adobecom.aem.live/fr/creativecloud/plans?mas-io-url=https://14257-merchatscale-andrei4226.adobeioruntime.net/api/v1/web/MerchAtScale
- non regression on catalog page https://main--cc--adobecom.aem.live/products/catalog?mas-io-url=https://14257-merchatscale-andrei4226.adobeioruntime.net/api/v1/web/MerchAtScale

### bugfix/feature

Promo project: https://mwpw-199953--mas--adobecom.aem.live/studio.html#page=promotions-editor&path=sandbox&promotionId=049f0402-0720-41e1-8b92-c0e2805d832e

Fragment: https://mwpw-199953--mas--adobecom.aem.live/studio.html#page=content&path=nala&query=146fec18-3d9c-4f93-908b-fd5e4ee76436

Promo variation for en_AE: https://mwpw-199953--mas--adobecom.aem.live/studio.html#fragmentId=ac0e986e-26d5-4b7e-9b68-ed1665be12cb&page=fragment-editor&path=nala

Promo variation for en_BE: https://mwpw-199953--mas--adobecom.aem.live/studio.html#fragmentId=4107382f-b084-4bb3-a13f-857809328d78&page=fragment-editor&path=nala&promotionId=049f0402-0720-41e1-8b92-c0e2805d832e

Promo variation for en_CA: https://mwpw-199953--mas--adobecom.aem.live/studio.html#fragmentId=ca649897-3521-4b95-b3a2-79523ef5b340&page=fragment-editor&path=nala&promotionId=049f0402-0720-41e1-8b92-c0e2805d832e

Link: https://mas.adobe.com/web-components/docs/plans-collection.html

Valid:
for en_AE:
- Before: https://mas.adobe.com/web-components/docs/plans-collection.html?locale=en_AE
- After:https://mwpw-199953--mas--adobecom.aem.page/web-components/docs/plans-collection.html?locale=en_AE&mas.preview=on&maslibs=MWPW-199953
- After: https://mwpw-199953--mas--adobecom.aem.live/web-components/docs/plans-collection.html?locale=en_AE&mas-io-url=https://14257-merchatscale-andrei4226.adobeioruntime.net/api/v1/web/MerchAtScale

for en_BE:
- Before: https://mas.adobe.com/web-components/docs/plans-collection.html?locale=en_BE
- After: https://mwpw-199953--mas--adobecom.aem.page/web-components/docs/plans-collection.html?locale=en_BE&mas.preview=on&maslibs=MWPW-199953
- After: https://mwpw-199953--mas--adobecom.aem.live/web-components/docs/plans-collection.html?locale=en_BE&mas-io-url=https://14257-merchatscale-andrei4226.adobeioruntime.net/api/v1/web/MerchAtScale

for en_CA:
- Before: https://mas.adobe.com/web-components/docs/plans-collection.html?locale=en_CA
- After: https://mwpw-199953--mas--adobecom.aem.page/web-components/docs/plans-collection.html?locale=en_CA&mas.preview=on&maslibs=MWPW-199953
- After: https://mwpw-199953--mas--adobecom.aem.live/web-components/docs/plans-collection.html?locale=en_CA&mas-io-url=https://14257-merchatscale-andrei4226.adobeioruntime.net/api/v1/web/MerchAtScale

image
image

## 🧪 Nala E2E Tests

### To run Nala tests again:

- Remove the `run nala` label

## Test URLs:

- After: https://mwpw-199953--mas--adobecom.aem.live/


npeltier:
@Andrei4226 note sure you have discussed this link when several geos are available
image
it's a bit misleading to me to have "one" variation link going to the first one.
I would rather have "view promo variations" going to the table view, or some modal proposing which geo to go to?

io/www/src/fragment/transformers/customize.js
```diff
@@ -178,15 +178,61 @@ function promoProjectLabel(project) {
...
+        const geo = matchesGeo(pznTags, { regionLocale, country });
+        if (!geo) continue;
+        const score = (geo.region ? 2 : 0) + (geo.country ? 1 : 0);
+        if (score > bestScore) {
+            bestScore = score;
+            best = candidate;
+        }
+    }
+    return best || fallback;
+}
```
npeltier, line 218: rather than ...-1, ...-2, ...-3, ...  variations with tags to solve with complicated scoring, why not relying on a path based resolution?
if i create a variation for en_AE, having one ...-en_AE variation i can check right away?

npeltier (commented):
please answer :)

io/www/src/fragment/transformers/customize.js
Andrei4226: I think that based only on path I can no longer have a single variation that covers multiple geos simultaneously
A single variation can have pznTags: `['mas:country/AE', 'mas:country/SA', 'mas:country/KW']` and `matchesGeo` will return a match for any request coming from AE, SA or KW, using the same fragment
Path based would be simpler to resolve but we would trade that for content duplication (one fragment per locale even when the content is identical across a region)

Andrei4226:
> @Andrei4226 note sure you have discussed this link when several geos are available image it's a bit misleading to me to have "one" variation link going to the first one. I would rather have "view promo variations" going to the table view, or some modal proposing which geo to go to?

Yes, sounds good to me
Should be also in this PR or as a follow-up ticket? @npeltier @3ch023
EDIT: I forgot about this ticket: https://jira.corp.adobe.com/browse/MWPW-200064

io/www/src/fragment/transformers/customize.js
npeltier: i see thanks

npeltier: fine, but then we should probably factorize geo scoring in some geoMatchScore, you can use in both pzn & promo

3ch023 (changes requested):
found a little bug:
probePromoVariationsForFragment stops probing as soon as it hits a missing index. But variations can be deleted individually, so if someone deletes 2nd one, 3d variation becomes invisible.

you can reproduce it here:
https://mwpw-199953--mas--adobecom.aem.live/studio.html#page=content&path=sandbox&query=cae6ca50-cfd7-4870-a875-65906f471945

I deleted en_BE (idx3) variation and now i can't see my en_EG variation.
thought [en_EG](https://mwpw-199953--mas--adobecom.aem.live/studio.html#fragmentId=5c0678cb-23b0-400c-aac8-e4d491c196c5&page=fragment-editor&path=sandbox) still exists and [renders](https://main--milo--adobecom.aem.page/merch/mas/preview?fragment-id=cae6ca50-cfd7-4870-a875-65906f471945&content-type=merch-card&locale=en_EG&clearCaches.io&debug.io&mas.preview=on&maslibs=MWPW-199953), its invisible in studio

npeltier (dismissed):
in addition, to mine, here are claude reviews (i think Mariia's feedback is fixed with 1.)
  Findings (most severe first)

  1. studio/src/promotions/promotion-variations.js:37-58 + :333-346 — sequential probing + swallowed delete errors can permanently orphan variations.
  probePromoVariationsForFragment stops at the first missing index (if (!variation?.id) break;, line 45). deleteAttachedPromoVariations deletes attached
  variations one at a time in a for...of with a per-item try/catch that only logs on failure (line 342-344) — it doesn't retry or abort. If deleting index-1
  succeeds but index-2 fails (transient AEM error), every subsequent probe for that fragment+tag breaks at index-1 and never sees index-2 (or any index beyond
  it) again. That orphan then silently escapes geo-overlap checks on new variation creation, and future "delete/unpublish project" runs stop seeing it too — a
  permanently orphaned, potentially still-published fragment. Confirmed independently by 4 of the 8 finder passes plus my own read of the source.

  2. studio/src/promotions/promotions-repository.js:157-160 vs :47-54 — createPromoVariation skips the store-freshness guard its siblings use.
  resolveDefaultFragmentForPromoVariation and mergePromoReferencesIntoFragmentData both go through getPromotionProjectsForProbe(loadPromotions), which
  force-fetches the promotions list if it hasn't loaded yet. createPromoVariation instead calls readPromotionProjectsFromStore() directly with no such guard. If
  it's invoked before the promotions list has ever been fetched, attachedFragmentPaths and projectGeos come back [], silently disabling both the path-collision
  check and the legacy-sibling geo-overlap fallback. Plausible — depends on whether every call site guarantees the store is warm first (not fully traced).

  3. studio/src/promotions/promotion-variations.js:260-269 (resolveDefaultFragmentForPromoVariation) — candidate tie-break can pick the wrong parent. Candidates
  are sorted by attachedSet.has(...), but when neither candidate is in attachedFragmentPaths, the sort is a no-op and falls back to getCandidateDefaultLeaves's
  insertion order: the suffixed path (e.g. my-card-2) is tried as a literal default fragment before the stripped one (my-card). If a real, unrelated fragment
  happens to be named with a trailing -N, it could be resolved as the "parent" instead of the actual card. Lower-confidence edge case.

  4. studio/src/promotions/promotion-variations.js:333-346 vs :288-301 — inconsistent concurrency in the same file. collectAttachedPromoVariations already uses
  processConcurrently/VARIATIONS_CONCURRENCY_LIMIT, but the new deleteAttachedPromoVariations deletes variations sequentially via for...of + await. Deleting a
  project with 20 attached geo variations costs 20 sequential round-trips instead of bounded-parallel.

  5. Duplicated suffix scheme across runtimes. MAX_PROMO_VARIATIONS_PER_FRAGMENT = 50 and the -N suffix convention are independently reimplemented in
  io/www/src/fragment/transformers/customize.js and studio/src/promotions/promotion-variations.js:18. A future change to the cap or suffix format has to be
  hand-replicated in both, with no shared spec/test.

  6. studio/src/promotions/mas-promo-variation-geos.js reimplements an existing pattern. The search + select-all + indeterminate-checkbox list is near-identical
  to studio/src/translation/mas-translation-languages.js (same getters, same logic, "locale" swapped for "geo") — a generalized list-selector component would
  avoid maintaining two copies of the same a11y/edge-case behavior.

  7. Convention violation — studio/src/mas-fragment-editor.js:1833. Root CLAUDE.md: "Use getters (e.g., get headerTemplate()) for rendering HTML sections, not 
  renderXyz() methods." The new displayPromoVariationGeos(clazz) method renders an HTML section but isn't a getter, unlike the sibling get 
  promoVariationGeoTagsTemplate() added elsewhere in this same PR.

  Not flagged as bugs (checked and ruled out): all real call sites of the string→string[] migration (mas-repository.js, editor-context-store.js,
  promotion-variations.js) correctly iterate the array rather than treating it as a truthy scalar.

npeltier:
Finally, not related to your PR but we should probably not have promotions-model & promotions-repository, as their scope is exactly the same

afmicka (approved):

3ch023 (approved):

PR #1039, opened 2026-07-13
MWPW-197349: Bulk version/revert follow-up — snapshot recording at project save time

seanchoi0:
Record a snapshot (non-translation CF version ID) every time a bulk publish project is saved, so that bulk publish can use it as the revert target instead of the pre-publish state. Translation versions (identified by `createdBy === 'odin-cf-versioning-user'` or `comment.startsWith('Pre-rollout snapshot')`) are skipped when finding the revert target.

Resolves https://jira.corp.adobe.com/browse/MWPW-197349

Test URLs:

- After: https://mwpw-197349--mas--adobecom.aem.live/


io/studio/src/bulk-publish/snapshot.js
```diff
@@ -51,6 +51,25 @@ async function unpublishFragment(odinEndpoint, fragmentPath, authToken) {
     });
 }
 
+function isTranslationVersion(version) {
```
Axelcureno, line 67: **[test gap]** `isTranslationVersion` intentionally treats `Pre-bulk-publish` versions as non-translation, so on save→publish→save the second `recordSnapshot` can pick a prior `Pre-bulk-publish` version as the revert target instead of the last authored/published state. No test pins this behavior. Please add one where version history already contains a `Pre-bulk-publish` entry — it documents the intent and guards against an accidental flip in this predicate.

io/studio/src/bulk-publish/bulk-publish-worker.js
```diff
@@ -31,6 +31,18 @@ function hasPendingSnapshot(entries) {
     }
 }
 
+function hasValidPreRecordedSnapshot(entries) {
```
Axelcureno, line 25: **[test gap]** The worker tests cover pre-recorded / empty / malformed entries, but none exercises an entry with `publishComplete` defined (a finalized prior publish). That's exactly the `publishComplete === undefined` guard — dropping it would make a stale completed snapshot get reused as the revert target and skip re-recording, with no failing test. Please add a case with `publishComplete: true` asserting it falls through to record + createSnapshot.

Axelcureno (commented):
Reviewed the snapshot-at-save flow end to end — all 50 IO tests pass and the serialize/deserialize contract with `revertSnapshot` is consistent. Requesting two tests for scenarios where a future change could silently regress the revert target. Non-blocking.

io/studio/src/bulk-publish/bulk-publish-worker.js
honstar: In addition to what @Axelcureno wrote above, this function only checks per-entry shape, never that recorded fragment IDs cover the current project paths. A fragment added after the last successful snapshot save gets published but has no revert entry - `revertSnapshot` will silently skip it.

io/studio/src/bulk-publish/snapshot.js
```diff
@@ -135,6 +154,30 @@ async function revertSnapshot({ entries, odinEndpoint, authToken }) {
...
+    const createdAt = new Date(timestamp).toISOString();
+
+    logger.info(JSON.stringify({ event: 'record-snapshot-start', count: paths.length }));
+
+    const entries = await processBatchWithConcurrency(paths, FRAGMENT_CONCURRENCY, async (path) => {
+        const fragment = await getFragmentByPath(odinEndpoint, path, authToken);
+        if (!fragment) throw new Error(`Fragment not found at path: ${path}`);
+        const wasPublished = fragment.status === STATUS_PUBLISHED || fragment.status === STATUS_MODIFIED;
+        const versionId = await findNonTranslationVersion(odinEndpoint, fragment.id, authToken);
+        if (!versionId) throw new Error(`No non-translation version found for fragment: ${path}`);
```
honstar: This throws for the whole batch if even one fragment lacks a non-translation version. In the frontend it's just console.error'd (user never told the project has no valid snapshot); in `runWorker`'s fallback branch it's uncaught, propagates to a bare 500, and - unlike `revert-action.js`'s established pattern of writing `lastError` back to the project — never updates project status.

studio/src/bulk-publish/mas-bulk-publish-editor.js
```diff
@@ -555,6 +557,14 @@ class MasBulkPublishEditor extends LitElement {
                     Store.bulkPublishProjects.inEdit.set(new FragmentStore(new Fragment(raw)));
                     this.hasChanges = false;
                     showToast('Project created successfully.', 'positive');
+                    const { saveSnapshot } = await import('./bulk-publish-store.js');
+                    const createdProject = Store.bulkPublishProjects.inEdit.get();
+                    saveSnapshot({
```
honstar: Saving a snapshot should be awaited to avoid a fire-and-forget situation: if a user clicks `Publish` right after `Save`, this can trigger `runWorker` before the fresh snapshot write lands, reading stale/absent snapshot data instead.

studio/src/bulk-publish/mas-bulk-publish-editor.js
```diff
@@ -573,6 +583,13 @@ class MasBulkPublishEditor extends LitElement {
                     if (!saved) throw new Error('Save returned empty response');
                     this.hasChanges = false;
                     showToast('Project saved successfully.', 'positive');
+                    const { saveSnapshot } = await import('./bulk-publish-store.js');
+                    saveSnapshot({
```
honstar: Same as above, please `await` this function call.

honstar (changes requested):

io/studio/src/bulk-publish/bulk-publish-worker.js
seanchoi0: @honstar  Your concern was actually considered during design. The snapshot is recorded on **every project save**, so adding a fragment to a project requires saving the project first, which triggers `saveSnapshot` immediately after. By the time the user can click Publish, the new fragment is already covered by the latest snapshot.

Snapshotting on every save (rather than only on create or only pre-publish) was chosen precisely to cover this case.

io/studio/src/bulk-publish/snapshot.js
seanchoi0: Done — added a test that places a `Pre-bulk-publish` version ahead of a translation version in the history and asserts it gets picked as the revert target.

io/studio/src/bulk-publish/bulk-publish-worker.js
seanchoi0: @Axelcureno I've added a test with `publishComplete: true` that asserts it falls through to the record + createSnapshot path.

io/studio/src/bulk-publish/snapshot.js
seanchoi0: Fixed. `recordSnapshot` now catches per-fragment errors internally and returns `{ entries, failures }` — matching the `revertSnapshot` pattern. `save-snapshot-action` writes `lastError` back to the project fragment when failures occur, and the worker's fallback branch does the same.

studio/src/bulk-publish/mas-bulk-publish-editor.js
seanchoi0: The original intent was to keep the save operation fast — the snapshot write is a background concern and a failure there shouldn't block the user from seeing the 'Project saved' confirmation. That's why I left it fire-and-forget with a `.catch`.

That said, I think your point is more important: if the user clicks Publish immediately after Save, the worker could read stale or absent snapshot data before the write lands. The risk of a silent race outweighs the minor UX cost of waiting.

Fixed in both the create and update paths — now `await`ed inside their own `try/catch` so snapshot errors are still non-fatal to the save flow.

studio/src/bulk-publish/mas-bulk-publish-editor.js
seanchoi0: Same fix applied here.

io/studio/src/bulk-publish/snapshot.js
```diff
@@ -135,6 +154,33 @@ async function revertSnapshot({ entries, odinEndpoint, authToken }) {
...
 }
 
+async function recordSnapshot({ paths, odinEndpoint, authToken }) {
+    const timestamp = Date.now();
+    const createdAt = new Date(timestamp).toISOString();
+
+    logger.info(JSON.stringify({ event: 'record-snapshot-start', count: paths.length }));
+
+    const results = await processBatchWithConcurrency(paths, FRAGMENT_CONCURRENCY, async (path) => {
+        const fragment = await getFragmentByPath(odinEndpoint, path, authToken);
```
honstar: `getFragmentByPath` is only ignoring `404` which means that all other non 2xx responses (500, network error) still throw and abort the whole batch, triggering the original all-or-nothing failure.

io/studio/src/bulk-publish/snapshot.js
```diff
@@ -135,6 +154,33 @@ async function revertSnapshot({ entries, odinEndpoint, authToken }) {
...
+        if (!fragment) return { path, error: `Fragment not found at path: ${path}` };
+        const wasPublished = fragment.status === STATUS_PUBLISHED || fragment.status === STATUS_MODIFIED;
+        const versionId = await findNonTranslationVersion(odinEndpoint, fragment.id, authToken);
```
honstar: Same as above: `findNonTranslationVersion` is only ignoring `404` which means that all other non 2xx responses (500, network error) still throw and abort the whole batch, triggering the original all-or-nothing failure.

io/studio/src/bulk-publish/bulk-publish-worker.js
honstar: This is in fact cancelling any error that might have been recorded just moments earlier above.

honstar (changes requested):

io/studio/src/bulk-publish/snapshot.js
seanchoi0: Fixed — wrapped the entire batch callback in try/catch so unexpected throws from `getFragmentByPath` (500, network error, etc.) are captured as per-fragment failures instead of aborting the whole batch.

io/studio/src/bulk-publish/snapshot.js
seanchoi0: Same fix applied here.

io/studio/src/bulk-publish/bulk-publish-worker.js
seanchoi0: Fixed — introduced a `snapshotError` variable that's set in the else branch (fallback path) and carried through to the final `updateProject` call. In the normal `hasValidPreRecordedSnapshot` flow `snapshotError` remains `''`, so the final update clears `lastError` only when publish completes cleanly — which is the intended behavior.

io/studio/src/bulk-publish/snapshot.js
honstar: Analogous to the fix applied to `recordSnapshot` where a single failure would reject the whole batch of fragments, please wrap this in a `try/catch` as well.

honstar (changes requested):

io/studio/src/bulk-publish/snapshot.js
seanchoi0: Fixed — applied the same `{ entries, failures }` pattern as `recordSnapshot`. Each fragment callback is now wrapped in try/catch so a single failure returns `{ path, error }` instead of aborting the whole batch. `createSnapshot` now returns `{ entries, failures }` and the worker accumulates any failures as a `CREATE_SNAPSHOT:` prefix in `snapshotError`, surfaced in `lastError` the same way `SAVE_SNAPSHOT:` failures are.

honstar (approved):

honstar:
Please resolve conflicts.

studio/src/bulk-publish/mas-bulk-publish-editor.js
```diff
@@ -555,6 +557,18 @@ class MasBulkPublishEditor extends LitElement {
                     Store.bulkPublishProjects.inEdit.set(new FragmentStore(new Fragment(raw)));
                     this.hasChanges = false;
                     showToast('Project created successfully.', 'positive');
+                    const { saveSnapshot } = await import('./bulk-publish-store.js');
```
Andrei4226: Duplicated with the identical block at line 590 (same dynamic import + `try/catch` around `saveSnapshot({...}`), only project and the error message differ). 
Maybe we can extract into a private method (`#recordSnapshotAfterSave(project, context)`) and call it from both places.

io/studio/src/bulk-publish/snapshot.js
```diff
@@ -135,6 +161,37 @@ async function revertSnapshot({ entries, odinEndpoint, authToken }) {
...
+            const wasPublished = fragment.status === STATUS_PUBLISHED || fragment.status === STATUS_MODIFIED;
+            const versionId = await findNonTranslationVersion(odinEndpoint, fragment.id, authToken);
+            if (!versionId) return { path, error: `No non-translation version found for fragment: ${path}` };
+            return [fragment.id, { path: fragment.path, versionId, wasPublished }];
+        } catch (err) {
+            return { path, error: err.message };
+        }
+    });
+
+    const failures = results.filter((r) => r?.error);
```
Andrei4226, line 241: `revertSnapshot` only walks `Object.entries(snapshot.fragments)`
A fragment that failed during `recordSnapshot` (line 183) is never added to `snap.fragments`, so it's absent from entries here too, it ends up in neither failures nor skipped, with no trace at revert time. Consider having `recordSnapshot` persist failed paths alongside the snapshot (or have the caller track them) so revertSnapshot/the UI can surface "N fragments have no revert target" instead of silently skipping them.

.claude/commands/release-mas-js.md
```diff
@@ -8,20 +8,11 @@ Release the `@adobecom/mas` npm package and create a GitHub release at `adobecom
...
-The **published GitHub release** is the source of truth for the current version — not `package.json`, which can drift if a prior release's bump commit (step 10) never landed.
+Read `web-components/package.json` to show the current version.
 
-```bash
-gh release list --repo adobecom/mas --limit 10 --json tagName,publishedAt,name
-```
-
-Take the highest `mas-js-v*` tag as the current version. Then read `web-components/package.json` and compare:
-
-- If `package.json` is **behind** the latest release tag, it drifted (a previous bump commit was never committed/pushed). Warn the user, and treat the release tag — not `package.json` — as the baseline for choosing the next version.
```
yesil, line 21: please do not revert this section. I guess this file is not supposed to be changed in this PR.

io/studio/src/bulk-publish/snapshot.js
```diff
@@ -51,6 +51,25 @@ async function unpublishFragment(odinEndpoint, fragmentPath, authToken) {
...
+        ignoreErrors: [404],
+    });
+    if (!response.ok) return [];
+    const data = await response.json();
+    return data?.items ?? [];
+}
+
+async function findNonTranslationVersion(odinEndpoint, fragmentId, authToken) {
+    const versions = await fetchVersionHistory(odinEndpoint, fragmentId, authToken);
+    const found = versions.find((v) => !isTranslationVersion(v));
```
yesil, line 83: `.find` takes the first non-translation entry, so the revert target assumes Odin returns `/versions` newest-first. If that order isn't guaranteed we could revert to an older version — worth pinning with an explicit sort or a comment noting the assumption.

studio/src/bulk-publish/mas-bulk-publish-editor.js
seanchoi0: Fixed — extracted into a `#recordSnapshotAfterSave` private method shared by both the create and update paths.

io/studio/src/bulk-publish/snapshot.js
seanchoi0: The silent skip is intentional here: if a fragment had no non-translation version at save time, there's no valid prior state to restore — adding it to a `noRevertTarget` list wouldn't change that, it would just re-surface information the user already received. The partial failure is written to `lastError` on the project (by `save-snapshot-action`, visible in the Studio UI) before they ever reach the revert button, so they know the snapshot was incomplete going in. The skip-on-missing-version path is also consistent with how `revertSnapshot` handles 404-deleted fragments (the existing `skipped` return). Happy to revisit if you think the caller needs the information for a different reason.

.claude/commands/release-mas-js.md
seanchoi0: Fixed — restored to match `main`. This file came in via the `.claude/` symlink during a merge; it shouldn't have been part of this PR's diff at all.

io/studio/src/bulk-publish/snapshot.js
seanchoi0: Added a comment documenting the assumption — Odin returns `/versions` newest-first, so `.find` picks the most recent non-translation version. Let me know if you'd prefer an explicit sort instead.

afmicka:
@seanchoi0 could you please resolve conflicts?

afmicka:
@seanchoi0 when i try to revert the bulk publish project i get 404 error. Please take a look
https://mwpw-197349--mas--adobecom.aem.live/studio.html?io.studio.env=seanchoi#bulkPublishProjectId=3fbdb394-48a0-4c72-ad79-ca5f3a133355&page=bulkPublishEditor&path=sandbox

Screenshot 2026-09-01 at 16 01 22

seanchoi0:
@afmicka io/studio 재배포가 필요했고, 지금은 완료됐어요. 이제 동작할 거예요!

seanchoi0:
@afmicka Needed to redeploy io/studio — done now. Should be working!

afmicka (approved):

PR #1045, opened 2026-07-15
MWPW-190304: Contribute M@S Studio Extension

Axelcureno:
Migrates the M@S Studio Extension from its standalone local repo (never published, no remote) into the monorepo as a top-level `extension/` module, following the `edgeworker/` precedent: own `package.json`, deliberately not an npm workspace.

**On the "custom api key" AC — resolved by deletion, not implementation.** `auth/ims-auth.js` hardcoded a placeholder client ID (`mas-studio-extension`) that was never registered with IMS, and the `.env`/`setup.sh` scaffolding around it was never read at runtime. More importantly, no request ever sent an `Authorization` header: fragment data comes from the public M@S IO endpoint via the public `wcms-commerce-ims-ro-user-milo-extension` key. The OAuth flow granted no access while asking for real Adobe credentials and 7 scopes, so it's gone.

- Delete `ims-auth.js`, `token-manager.js`, auth message handlers, `AEMClient.getAccessToken()`
- Drop `identity` permission and `*.adobelogin.com` host/CSP entries — manifest is now `storage` + `tabs`
- Delete dead `AEMClient.validateFragment()`/`detectIssues()`
- `"type": "commonjs"` so the `node:test` suites run under the ESM root; add the missing `test` script (18 tests, previously unrunnable)
- Path-filtered CI at `.github/workflows/extension-pr.yaml`
- Format to repo prettier style; `extension/vendor` (third-party Spectrum CSS) excluded

Root `npm run build`/`npm test` are unaffected — verified, `extension/` is outside `workspaces`.

Not included, deliberately: `utils/locales.js` duplicates concepts in `studio/src/locales.js`, but Studio's is ESM and content scripts are classic non-module with no bundler — sharing needs a build step. Follow-up ticket.

## Test URLs:

- After: https://mwpw-190304--mas--adobecom.aem.live/

Not web-testable — this is a Chrome extension. To verify: load unpacked from `extension/` at `chrome://extensions/`, open a page with merch cards (e.g. https://www.adobe.com/creativecloud/plans.html), confirm badges render and a badge click opens the fragment in Studio.

## Checklist:

- [x] Tests pass locally (18/18)
- [ ] Tested on Before/After URLs — n/a, see above


extension/api/aem-client.js
```diff
@@ -0,0 +1,78 @@
+class AEMClient {
+    constructor() {
+        this.masIOUrl = 'https://www.adobe.com/mas/io';
+        this.wcsApiKey = 'wcms-commerce-ims-ro-user-milo';
```
Axelcureno: need to get a new key

Andrei4226 (commented):
Wow, awesome feature @Axelcureno ! 👏 

My findings:
- the value for the "Path" field appears only when I press the "Additional details" button (it is not an issue, but it always shows "Loading" if I don't press that button)
- I know this feature is for pages, to see details for each card, and to find them in Studio, but with the extension activated on `https://mas.adobe.com/studio.html#page=content&path=sandbox`, it will appear like this (just a mention)

Image

Minor:
- `card-detector.js:62-103` — `localeFromUrl()` reimplements its own hardcoded `langCountryDefaults` (lang→country) lookup table to guess the locale from the URL, even though `utils/locales.js` (loaded earlier in `manifest.json`, exposed globally as `window.MASLocales`) already provides `getDefaultLocaleForLanguage`(language, surface) with canonical per-surface tables (ACOM/CCD/EXPRESS/etc.). Two independent sources of truth for the same thing — they can drift out of sync if one is updated and the other is forgotten.

npeltier (changes requested):
claude review ;) 
 Review: PR #1045 — Contribute M@S Studio Extension
  
  Overview

  Migrates a previously unpublished, standalone Chrome extension into the monorepo as extension/ (mirroring the edgeworker/ precedent — own package.json, deliberately outside
  workspaces). Of the reported 15.8K additions, ~10.7K are vendored third-party Spectrum CSS (extension/vendor/, regenerated by scripts/sync-spectrum.sh, excluded from Prettier)
  — not authored logic. The actual hand-written surface is ~20 files / ~3K lines: a service worker (background.js), content scripts (card-detector.js, overlay.js, content.js),
  a popup, and shared utils (fragment-parser.js, locales.js, validators.js, escape.js, studio-linker.js, icons.js). The auth removal (deleting a dead IMS OAuth flow that never
  actually sent an Authorization header) is a legitimate, well-justified deletion, not a regression.

  Correctness issues found

  1. "Edit in Studio" opens the wrong locale for any non-English card. In background.js::buildStudioUrl, the content view hardcodes both locale and path:
  const params = new URLSearchParams({
      page: 'content',
      query: fragmentId,
      locale: 'en_US',
      path: `/content/dam/mas/en-us/${sur}`,
  });
  No caller (studio-linker.js::openInStudio, overlay.js's edit button, popup.js::openInStudio) ever passes the card's actual locale/country, even though
  cardData.locale/card.locale are available at every call site. I confirmed in studio/src/router.js that path/locale/query are used as search-scoping filters for the content
  list (STORE_SEARCH_HASH_KEYS), not just cosmetic defaults — so for a fragment that actually lives under e.g. /content/dam/mas/fr-fr/acom/..., this deep link searches the en-us
  folder instead and will likely come up empty. Given the extension's core pitch is "one-click deep link," this silently breaks for every non-en_US merch card, which is
  presumably a large share of real usage on adobe.com.

  2. "Parent" locale-variation link never carries a fragment ID. In overlay.js:
  window.MASStudioLinker.openParentInStudio(variationInfo);
  but studio-linker.js declares openParentInStudio(variationInfo, parentFragmentId) — the second argument is never passed, so fragmentId is undefined inside the built message.
  Worse, fragment-parser.js::parseVariationInfo never captures the parent's actual fragment ID anywhere (only a computed localeDefaultPath string) — there's no data available to
  pass even if the call site were fixed. Contrast with sibling openVariationInStudio, which does carry a real id from extractVariations(). So clicking "Parent" degrades to a
  generic locale/surface navigation instead of a precise deep link, unlike every other link in the same panel.

  Test coverage

  18 tests exist, covering only aem-client.js, escape.js, and validators.js. The largest and most logic-heavy files are entirely untested: overlay.js (669 lines),
  card-detector.js (locale-detection heuristics, MutationObserver batching), fragment-parser.js (variation-path regex parsing), background.js's buildStudioUrl (the actual URL
  construction that feeds isAllowedOpenUrl — the one function this bug lives in), popup.js, locales.js. Since this workspace sits outside the root's enforced coverage
  thresholds, nothing blocks merging with this gap, but it's exactly the kind of URL-building logic where a test would have caught issue #1 above.

  Security posture (positive)

  The removal of the dead OAuth flow is the right call, and what replaced it is genuinely careful for a content-script-injecting extension:
  - background.js checks sender.id === chrome.runtime.id before handling any message.
  - Every fetch/message validates UUID/locale/country via strict regexes before touching the network.
  - isAllowedOpenUrl allowlists only https://mas.adobe.com before any chrome.tabs.create — blocks javascript:/data:/arbitrary-host payloads reflected through message data.
  - Consistent, correctly-ordered (& first) HTML escaping with separate text/attribute variants, applied at every interpolation point I checked in overlay.js/popup.js.
  - Manifest CSP is restrictive (script-src 'self', object-src 'none'), permissions trimmed to storage+tabs only. 

  Minor items

  - Debug console.log/console.error left throughout shipped code (aem-client.js logs full request/response details, background.js/content.js log lifecycle messages) —
  non-sensitive (all public data) but noisy on every matched adobe.com page load.
  - popup.js reimplements its own local escapeHtml (DOM-based) rather than reusing utils/escape.js, because popup.html doesn't load that script — harmless duplication, worth
  consolidating later.
  - PR description says fragment data comes via the wcms-commerce-ims-user-prod key; the code and README both actually use wcms-commerce-ims-ro-user-milo-extension — just a
  description/code mismatch, not a functional issue.
  - The new extension-pr.yaml workflow only runs npm test; no lint/format check is wired in for this workspace, so future changes rely on manual npm run format discipline.

  Overall: solid security fundamentals and a justified auth-removal, but I'd fix the two locale/ID deep-linking bugs before merge — they undercut the extension's main selling
  point ("one-click deep link") for exactly the multi-locale content this tool is meant to help manage.

Axelcureno:
@npeltier both bugs are fixed. The card locale now flows into every Studio link, and the Parent link refetches the locale default to get the real parent id, with a fallback if that fails. Pulled buildStudioUrl into utils/studio-url.js and added 7 tests (52/52 passing). Page-load logs are gone. Deferred: popup escapeHtml reuse, the PR description key name (fixed now), and CI lint, which needs an eslint env override for extension globals first.

npeltier (approved):
thx! what would be cool would be a GH workflow for releasing the extension with a version to this repo's releases, but i guess that can be done later

extension/scripts/sync-spectrum.sh
```diff
@@ -0,0 +1,81 @@
+#!/usr/bin/env bash
```
yesil, line 1: Why doesn’t the Spectrum consumption follow the Studio approach?

For example, could we generate a `swc.js` that contains all SWC dependencies?

I’m personally fine with merging this as-is, but if a review is required, I think this approach is diverging from the existing Studio pattern.

extension/api/aem-client.js
```diff
@@ -0,0 +1,70 @@
+class AEMClient {
+    constructor() {
+        this.masIOUrl = 'https://www.adobe.com/mas/io';
+        this.wcsApiKey = 'wcms-commerce-ims-ro-user-milo-extension';
```
yesil: I think the extension should not have its own aem client, but use whatever is available in the page.
thus we guarantee that it is using the same params accurately.

extension/scripts/sync-spectrum.sh
yesil: I noticed that this is using spectrum css, any reason for that? Is it for avoiding conflicts with Studio while registering web components?

extension/utils/card-detector.js
```diff
@@ -0,0 +1,322 @@
...
+        const alias = URL_SEGMENT_ALIASES[langSegment];
+        const lang = alias?.lang || langSegment;
+        const locales = typeof window !== 'undefined' ? window.MASLocales : null;
+        const defaultLocale = locales?.getDefaultLocaleForLanguage(lang);
+        if (!defaultLocale) return null;
+        let country = alias?.country || defaultLocale.split('_')[1];
+
+        const secondSegment = segments[1]?.toLowerCase();
+        if (secondSegment && /^[a-z]{2}$/.test(secondSegment)) {
+            const looksLikeCountry = [
```
yesil: what is the reason for having a subset of country codes here?

extension/utils/locales.js
yesil, line 1: I think, it should use this data from M@S and not have own copy.  It can create gaps as things change with Lingo.

yesil (dismissed):
I think the extension should not have own copy of aem-client, locale mappings etc.
It should try to re-use as much as possible, otherwise we would not sure if we see the data via the extension is actually correct. This can create more support work for us which I would like to avoid.

extension/utils/card-detector.js
Axelcureno: No good reason, it's a hand-typed dupe of the `regions` arrays and `MASLocales` is already loaded there. Mostly dead anyway since `<mas-commerce-service>` wins first. Going away with the locale change.

extension/api/aem-client.js
Axelcureno: Client has to stay, [MV3](https://developer.chrome.com/docs/extensions/develop/migrate/what-is-mv3) forces the split. It runs in the service worker (`background.js:1-3`) which has no DOM, and content scripts can't do the cross-origin fetch. Nothing in the page to reach from there.

But you're right that it shouldn't be guessing params. It hardcodes `masIOUrl = 'https://www.adobe.com/mas/io'` while M@S resolves `?mas-io-url=` -> config -> env (`web-components/src/settings.js:123-126`), so on a branch io deploy the extension reads prod and shows data the card didn't render from. Content script already reads `<mas-commerce-service>` for locale (`utils/card-detector.js:47-60`) and that element exposes `mas-io-url`, `env`, `wcs-api-key`, so I'll grab them there and pass them through. Will fix.

Checked the api key suffix while I was in there, it's harmless. Byte-identical data to `wcms-commerce-ims-ro-user-milo`, and still matches `wcms-commerce-ims-ro.+` so same surface.

extension/scripts/sync-spectrum.sh
Axelcureno: Yeah, conflict avoidance, exactly what you guessed.

`studio/src/swc.js` imports call `customElements.define()` at module eval, and a content script shares `customElements` with the page. On anything running mas.js those collide. We already hit this in `ost/src/swc.js:1-16`, which monkey-patches `CustomElementRegistry.prototype.define` and never restores it. Doing that to a page we don't own isn't great, and if the extension registers first the page's own Spectrum binds to our copy.

Also swc.js builds to ESM loaded with `type="module"`, MV3 content scripts are classic scripts.

The bit that does apply is `studio/src/spectrum.css.js` (raw `@spectrum-css`, plain classes, light DOM), which is what I vendored. So it is the Studio pattern, just the CSS half.

Side effect you made me notice: `@spectrum-css/tokens` puts custom props on `:root`, which leaks into the host page. Scoping that to our container.

Axelcureno:
@yesil fair points, mostly. Dug into all four, three hold up.

Locales copy is the worst one. We already have three copies of these tables (`io/www/src/fragment/locales.js`, `io/studio/src/locales.js`, and mine) and the io/studio one already drifted. `studio/src/locales.js` does it right with a re-export barrel. I just missed that pattern.

One thing I found while testing: io already does the regional fallback server-side. Request `de_AT`, response path comes back `de_DE`. Same for `en_AU` -> `en_GB`, `es_MX` -> `es_ES`. So the extension is redoing that from a stale copy.

Can't delete the tables outright though. Comparing requested vs resolved locale describes the fragment you got served, not the one you asked for, and `isVariationPathInParentLocaleFamily` needs `getRegionLocales` anyway. What's missing from the response is the surface default locale, which is the only bit needed for variation-vs-default. `parsedLocale`, `surface` and `defaultLocale` are all sitting on `context` at serialization (`io/www/src/fragment/pipeline.js:163-194`), just not returned.

Two options:

1. Return those three fields from io. ~3 lines, and the extension drops `locales.js` entirely. Separate ticket though, io/www shouldn't ride along in an extension PR.
2. Generate the extension's copy from `io/www` at package time, like `sync-spectrum.sh` does for Spectrum. Could fold in the io/studio fork too, three copies -> one.

Either way the hand-written copy goes. Preference? cc @3ch023 @npeltier

extension/scripts/sync-spectrum.sh
Axelcureno: Correction on my last comment: the `:root` leak isn't real, I got that wrong. The vendored tokens are scoped to `.spectrum` / `.spectrum--light` / `.spectrum--medium`, no `:root` anywhere in `vendor/spectrum/`, and `content.js:25` only puts those classes on our own badge and panel. Nothing to fix, sorry for the noise.

extension/utils/studio-url.js
```diff
@@ -0,0 +1,68 @@
...
+const VARIANT_SURFACES = {
+    catalog: 'acom',
+    plans: 'acom',
+    'plans-v2': 'acom',
+    'plans-students': 'acom',
+    'plans-education': 'acom',
+    'ccd-slice': 'ccd',
+    'ccd-suggested': 'ccd',
+    fries: 'commerce',
+    'ah-try-buy-widget': 'ahome',
```
mirafedas, line 12: In the constants.js in /studio the surface name is 'adobe-home', not 'ahome', should we maybe keep the naming the same? Also, the list seems to be missing some variants, for example headless or mini-compare-chart - is this intentional?

extension/background.js
```diff
@@ -0,0 +1,57 @@
...
+    if (country && !MASValidators.isValidCountry(country)) {
+        sendResponse({ success: false, error: 'invalid_country' });
+        return;
+    }
+    try {
+        const client = new AEMClient({ masIOUrl, wcsApiKey });
+        const data = await client.fetchFragmentData(fragmentId, locale, country);
+        sendResponse({ success: true, data });
+    } catch (error) {
+        sendResponse({ success: false, error: 'fetch_failed' });
```
mirafedas, line 44: Nit: just a proposal of improvement, feel free to ignore in this PR - we could add better error messaging if we take the error message from the error object we receive here from AEM

extension/popup/popup.js
```diff
@@ -0,0 +1,572 @@
...
+                            updateCardItemInList(card.fragmentId, cardName);
+                        }
+                    }
+                },
+            );
+        }
+    });
+}
+
+function loadCardDetails(fragmentId) {
```
mirafedas, line 343: Seems like loadCardDetails, renderCardDetails, renderVariationSection, formatFieldName are not used anywhere, can you check?

yesil:
I think we can start using this extension and address the issues as they appear instead of trying to predict and fix all of them ahead of time.

afmicka:
@Axelcureno something for further improvement. When you go to id_id plans page (https://www.adobe.com/id_id/creativecloud/plans.html) and you click to edit in studio, it will land you on the en_US fragment. It should probably get you to the id_id one (that was probably translated from the default one)

I will merge this as is.

PR #1047, opened 2026-07-15
MWPW-198151: Recover bulk publish projects stuck in Publishing

Axelcureno:
Bulk-publish projects could sit in `Publishing` forever. The worker wrote `status=Publishing`, ran the publish, then wrote the terminal status — if anything threw in between, `main()` logged and returned 500 without resetting the status.

Resolves https://jira.corp.adobe.com/browse/MWPW-198151

## What changed

- **Worker self-heal** (`bulk-publish-worker.js`): `main()` now writes a terminal status on failure. When publish succeeded and only the status write failed, it reports the real outcome (`Published`/`Partially published`) instead of `Failed` — the cards are live, and a false `Failed` invites a needless revert. The recovery write is itself guarded so it can never escape `main()`.
- **Pending-marker semantics**: preserved on failure so a retry resumes against the interrupted snapshot; stripped on an explicit reset. The marker helpers moved to `project.js`, now shared by the worker and the reset action.
- **`bulk-publish-reset` action + "Reset to Draft"**: covers the crash the in-process handler can't reach (OpenWhisk timeout or hard kill). Sets `Draft`, strips the marker, keeps snapshot entries for revert history. Confirm dialog warns it may abandon a running publish.
- **Status enums consolidated**: one IO-side `PROJECT_STATUS`; `WORKER_STATUS` deleted. `LOCKED` stays UI-only — no IO writer sets it.

## Testing

- Unit: io/studio 509 passing, studio 3099 passing. Lint clean.
- **Verified end-to-end against real Odin** (personal IO deployment, not reviewer-accessible): forced a crash after the `Publishing` write and confirmed the project landed on `status=Failed` with `lastError` set and the `publishComplete:false` marker intact — where it previously stranded in `Publishing`. Also confirmed the happy path still writes `Published` + `publishedAt`/`publishedBy`/`lastResult` with the marker stripped, and that Reset to Draft returns status `Draft` with snapshot entries preserved. Throwaway sandbox fixture deleted after.
- Verified the `bulk-publish-project` model has no field for a publish start time, so no new field is written — every write uses only model-defined fields, now pinned by a test.

> **Note for reviewers:** a staleness guard (grey out Reset until a run is provably dead) is deliberately deferred — it needs a `publishingStartedAt` field and the prod model is locked. The confirm dialog is the v1 guard.

## Test URLs

- After: https://mwpw-198151--mas--adobecom.aem.live/


mirafedas (approved):

npeltier (approved):
lgtm

afmicka (approved):

PR #1051, opened 2026-07-16
MWPW-200909: Fix promo not applied on headless marquee discount price

Axelcureno:
Headless marquee prices now resolve with the promo project's context promo code — server cache prefill and client request both. (Earlier corrector markup-injection approach reverted per review.)

- `wcs.js`: bare markup OSIs fall back to the fragment's `fields.promoCode` (set by `applyPromoCode` from the promo project), so the WCS cache prefills the promo-coded offer.
- `mas-field.js`: mirrors merch-card's `contextPromotionCode` — reflects `fields.promoCode` as `data-promotion-code`, injects `options.promotionCode` when unset. Authored `cancel-context` overrides still win.
- Verified on deployed io (NalaEvergreen project): `fields.promoCode=NICOPROMO`, cache key `<osi>-us-mult-nicopromo` for a bare markup OSI.
- Note: no promo project currently targets the repro marquee fragment — visual demo needs that content authored.
- Tests: io suite 455 passing, 100% coverage; web-components 1531 passing.

Resolves https://jira.corp.adobe.com/browse/MWPW-200909

- [x] C4. PR description contains working Test Page link
- [ ] C5: ready to demo from Test Page
- [ ] C.6 re-read Jira, validate all ACs

Test URLs:

- Before: https://main--da-cc--adobecom.aem.page/ca/drafts/ravneet/dotcom-188566/downloads-test?instant=2027-07-13T05:00:00.000Z
- After: https://main--da-cc--adobecom.aem.page/ca/drafts/ravneet/dotcom-188566/downloads-test?instant=2027-07-13T05:00:00.000Z&maslibs=mwpw-200909&io.studio.env=axel


Axelcureno:
Headless marquees were showing 0% off instead of the real discount (cards were fine). Turns out the marquee repeats the same OSI a few times, and the discount tag was missing the data-promotion-code that a sibling price tag had. No promo code → WCS returns one plain offer → no discount. Cards only work because a card-scoped provider quietly copies the code onto every price — and a marquee has no card.

Fix lives in `corrector.js`: copy the promo code from a same-OSI sibling onto the tags that are missing it. Runs on all surfaces, never clobbers an existing code, no-ops if there's no sibling code.

Why there and not elsewhere: the client already reads the attribute (no web-components change), and wcs.js only builds a cache — it never touches the markup that ships. corrector is the last transformer and already rewrites fields.* HTML, so it's the natural home.

One thing worth a second pair of eyes: the regex runs on the parsed field value (real " quotes), so it skips the &quot; dance wcs.js does. Pretty sure that's right, but ping me if that boundary looks off.

npeltier (changes requested):
- please provide test url with mas-io-url, we can't afford reviewing lots of PRs and do that ourselves,
- adding data-promotion-code in corrector is not a good solution : you will keep having wcs transformer fetching base offer, and client not having cache pre-filled. I would dig down a bit more on 
1. getting mas fields using context promo code, rather than mimicking a promotion code set on the price itself. 
2. check/fix wcs transformer not using context promo code in that case

Axelcureno:
Direction adopted — corrector markup injection reverted.

What's in now:
1. `wcs.js`: bare markup OSIs fall back to the fragment's `fields.promoCode` (the context promo `applyPromoCode` already sets), so the cache prefills the promo-coded offer.
2. `mas-field.js`: mirrors merch-card's `contextPromotionCode` — reflects `fields.promoCode` as `data-promotion-code`, injects `options.promotionCode` when unset. Authored `cancel-context` overrides still win.

Verified on the deployed io (test URL in description, `mas-io-url` included): with the NalaEvergreen project, `fields.promoCode=NICOPROMO` and the cache prefills `<osi>-us-mult-nicopromo` for a bare markup OSI — WCS fetched with `promotion_code`.

Two things for your sign-off:
- An OSI inheriting the context promo no longer prefills its plain (no-promo) entry — a consumer querying without the code live-fetches instead. Mixed-case `noPromoOsis` semantics unchanged.
- No promo project currently targets the repro marquee fragment (publish has only the nala test projects), so the repro page can't visually demo yet. Is the Intro promo project still unpublished, or should QA author one?

io/www/src/fragment/pipeline.js
```diff
@@ -138,7 +138,8 @@ async function mainProcess(context) {
         if (transformer.init) {
             //we fork context to avoid init to override any context property
             const initContext = {
-                ...structuredClone(context),
+                ...structuredClone({ ...context, state: undefined }),
+                state: context.state,
```
npeltier, line 142: can you explain me why that change?

npeltier (changes requested):
thanks, it looks much better to me. There is still that chnage in pipeline init i don't understand.
Also i wonder how your wcs change works with collections, can you add some tests and eventual fix? (most complicated case i have in mind: collection has 2 cards with 2 different promos).

io/www/src/fragment/pipeline.js
Axelcureno: The init loop does `structuredClone(context)` to fork the context per transformer. With the current lockfile, `aio-lib-state` 5.3.1 attaches a winston logger to `context.state`, and `structuredClone` throws `DataCloneError: [winston logger] could not be cloned` — so every request 503'd on a fresh deploy. (Prod only survives because its deployed bundle froze an older lib version; the next prod deploy would hit this.)

The fix clones everything except the state client, then reattaches the original `state` reference by hand:
```js
...structuredClone({ ...context, state: undefined }),
state: context.state,
```
State is a shared client, not per-transformer data, so sharing the reference across forks is correct — we only clone to isolate the plain data. Happy to pull this into its own ticket/PR if you'd rather keep it out of this one.

Axelcureno:
Added collection tests (e5a5d32a0) — including your worst case: collection with 2 cards / 2 different promos.

The fallback is `promoCodeByOsi[baseOsi] ?? context.body.fields?.promoCode`. Each card's promo lives in `promoCodeByOsi` (built from `references`), so for a real card the `??` short-circuits and the top-level `fields.promoCode` is never consulted — no cross-application. And a collection **root** never carries `fields.promoCode` (`applyPromoCode` bails without `fields.osi`, which collection roots don't have), so the fallback only ever fires for a genuine top-level fragment's own bare OSI (the marquee case). So no code change was needed — the behavior was correct but untested; the 3 new tests lock it in:
- 2 cards / 2 promos → each cached with its own promo, neither cross-applied
- no-promo card in a promo'd collection → plain offer cached, not wrongly promoted
- bare OSI with no reference → never inherits another card's promo

io suite 458 passing, 100% coverage.

io/www/src/fragment/pipeline.js
Axelcureno: TL;DR: `aio-lib-state@5.3.1` puts a logger on `context.state`, which `structuredClone` can't clone → every fresh deploy 503s. Fix clones everything except state, then reattaches it by reference (it's a shared client, not per-transformer data).

npeltier: i read it :) 
thanks for explaining, it looked unrelated to the PR context, hence my surprise, ideally we would have dealt with it in a separate context, but i'm cool with it

npeltier (approved):
ok, i guess i'm good with it but for collection we are still not good: 
my worst case is having card a having promo context (promo field set) and card b too. In that case the code is not setting the card's promo context to the card's promo field value.
however it's not the core issue of your PR, will need another ticket/ fix

3ch023 (approved):

afmicka (approved):

Axelcureno:
It's no biggie, I added the test @npeltier

Blainegunn:
Codecov is down
This PR is urgent 
Considering all relevant tests are passing, we'll go ahead and merge.

PR #1054, opened 2026-07-17
MWPW-200064: View promo variations in Promo project fragment table view - part 2: view-only table

mirafedas:
Resolves https://jira.corp.adobe.com/browse/MWPW-200064

## 🧪 Nala E2E Tests

### To run Nala tests again:

- Remove the `run nala` label

## Test URLs:

- After: https://mwpw-200064-1--mas--adobecom.aem.live/studio.html


studio/src/common/components/mas-select-items-table.js
```diff
@@ -296,7 +298,16 @@ class MasSelectItemsTable extends LitElement {
...
-        return TABLE_COLUMNS[this.effectiveType][this.viewOnly ? 'viewOnly' : 'selectable'];
+        const base = TABLE_COLUMNS[this.effectiveType][this.viewOnly ? 'viewOnly' : 'selectable'];
+        const supportsExtraCells = this.viewOnly && [TABLE_TYPE.CARDS, TABLE_TYPE.COLLECTIONS].includes(this.effectiveType);
+        if (!supportsExtraCells) return base;
+        return [
+            ...base,
+            ...(this.effectiveType === TABLE_TYPE.CARDS && this.renderPreviewCell !== NOOP_CELL
+                ? [{ label: 'Preview', key: 'preview' }]
+                : []),
+            ...(this.renderActionsCell !== NOOP_CELL ? [{ label: 'Actions', key: 'actions' }] : []),
```
yesil: Column visibility keys off `renderActionsCell/renderPreviewCell !== NOOP_CELL`, but `mas-collapsible-table-row` defaults these to a separate `() => nothing` rather than the shared sentinel. It works now only because the caller passes fresh arrows. If anyone later passes an equivalent no-op to *hide* a column, this would instead render an empty Actions/Preview header over blank cells. Cleaner to gate on an explicit flag (or default to `null` and check `!= null`) in both components.

studio/src/common/components/mas-select-items-table.js
```diff
@@ -16,23 +16,27 @@ import {
...
 
+const NOOP_CELL = () => nothing;
+
 class MasSelectItemsTable extends LitElement {
     static styles = styles;
 
     static properties = {
         type: { type: String },
         viewOnly: { type: Boolean },
-        viewOnlyLoading: { type: Boolean, state: true },
```
honstar, line 25: Shouldn't these properties remain internal component state properties? Removing `state: true` lets Lit create or observe a corresponding HTML attribute for it which doesn't sound right for these imo.

studio/test/translation/mas-select-items-table.test.js
honstar, line 1: Since the component to be tested is located at `common/components/mas-select-items-table.js`, this corresponding test file should be moved over to `test/common/components` respectively.

studio/src/promotions/mas-promotions-items-table.css.js
```diff
@@ -0,0 +1,425 @@
+import { css } from 'lit';
+
+export const promotionsItemsTableStyles = css`
+    :host {
+        width: 100%;
+        display: flex;
+        min-height: 0;
+    }
+
+    .promotions-view-only .offer-cell {
```
honstar, line 10: This including the following rules for `.mnemonic-icon, .path, .preview-cell, .actions-cell, .promotions-fragments-layout/.promotions-collections-layout` column widths can never apply: the markup now renders inside `<mas-select-items-table` / `<mas-collapsible-table-row>` own shadow roots which this stylesheet can't reach.

honstar (changes requested):

studio/src/common/components/mas-select-items-table.js
mirafedas: I changed them to be external so that parent can pass data it already has, avoiding a redundant fetch.

honstar: Ok, thanks for the explanation 👍

studio/src/promotions/mas-promotions-items-table.css.js
honstar: There is still a visual regression: while the dead code has been deleted, the underlying gap is still there. The commit removed the misleading dead CSS but didn't move the tuned styling anywhere.
Concretely: `mas-promotions-items-table.css.js` still defines `24px .mnemonic-icon` - but that rule now only reaches the Offers table (rendered in the same shadow root). The delegated Cards/Collections tables render their offer/mnemonic markup inside `mas-collapsible-table-row.js`, styled by its own CSS file which independently sets `32px` icons with single-line ellipsis truncation.
Net result: within the same Promotions editor, Offers tab shows 24px icons, Fragments/Collections tabs show 32px icons which is an inconsistency that seems to have been present before this PR already.

mirafedas: Yes, this has been present before this PR already, and anyway the design will change once the Figma for promotion variations is finalized (should be soon), so I'm leaving this unchanged for now.

honstar (approved):

studio/src/common/components/mas-select-items-table.js
```diff
@@ -64,7 +70,7 @@ class MasSelectItemsTable extends LitElement {
         this.dataState.abortController = new AbortController();
```
yesil, line 72: nit: when embedded by `mas-promotions-items-table`, the parent loads fragments in its own `updated()`→`#loadSelected` and passes them down. But the child mounts before that async load resolves, so `viewOnlyFragments` is still empty here and this `!viewOnlyFragments.length` guard still fires a second `loadSelectedFragments` for the same paths on first mount — it only avoids re-loading on later renders. Worth confirming; if it does double-fetch, an explicit "parent owns data" flag would be a cleaner gate than array length.

studio/src/promotions/mas-promotions-items-table.js
```diff
@@ -1362,51 +841,74 @@ class MasPromotionsItemsTable extends LitElement {
         </div>`;
```
yesil, line 841: nit: dropping `#emptySelectionMessage` means Cards/Collections now show the delegated generic "No items found." instead of the previous actionable copy ("No fragments selected. Use Add fragments to pick cards matching your offers." / "No collections selected."). Offers keeps its rich empty state; intentional?

studio/src/translation/mas-collapsible-table-row.js
```diff
@@ -80,6 +84,9 @@ export class MasCollapsibleTableRow extends LitElement {
                 this.isLoadingPromoVariations = false;
             }
         }
+        if (changedProperties.has('promoVariationsFetchedByParent') && this.promoVariationsFetchedByParent) {
```
yesil: nit: setting the reactive `promoVariations` from within `updated()` triggers a follow-up render (the "scheduled an update after an update completed" warnings in the test run). Deriving it in `willUpdate` or via a getter would avoid the extra cycle.

yesil (approved):
Approve with nits. Lint clean, 333 affected unit tests pass, tab-model refactor is consistent across the selector→table→row chain and I confirmed no stale `tab.key`/`nonSelectableVariations` consumers remain. A few non-blocking notes inline.

studio/src/promotions/mas-promotions-items-table.js
mirafedas: I'm keeping it for now as we are waiting for the final Figma, will address on a new Jira once we have it

afmicka:
@mirafedas after i add a fragment that has promo variation to the promotion project, the promo variation is not displayed in the view only table. I understood that this PR should resolve that. Could you please check? 
project: https://mwpw-200064-1--mas--adobecom.aem.live/studio.html#page=promotions-editor&path=sandbox&promotionId=a313968d-739c-4369-af83-183481e6d1e2
fragment: https://mas.adobe.com/studio.html#page=content&path=sandbox&query=bd525f5a-6db1-497c-9a5e-4a45a92da2e1

Screenshot 2026-07-27 at 13 44 33
Screenshot 2026-07-27 at 13 41 42

afmicka:
@mirafedas could you please check one more thing. In the translation project, i have selected both promo variations together with the default fragment. It says that are 3 fragments selected everywhere but in the side panel it displays only 2 fragments (default and one promo). The newly created promo variation from your branch is not being displayed 

project: https://mwpw-200064-1--mas--adobecom.aem.live/studio.html#page=translation-editor&path=sandbox&translationProjectId=9a035ced-733f-485a-a73e-c20075052c70


Screenshot 2026-07-27 at 14 43 32
Screenshot 2026-07-27 at 14 43 08

PR #996, opened 2026-06-19
MWPW-199086 Product Card Updates

Blainegunn:
Resolves 
https://jira.corp.adobe.com/browse/MWPW-199086 
https://jira.corp.adobe.com/browse/MWPW-198614

Product Card Updates: 
- "What's Included" & Short description section now support the info icon
- Short description enabled 
- "Add on" toggle enabled with background color support
- Pricing and percentage values from OST are no longer bolded 
- Asterisk to a price value stays inline 
- Tooltip pops from top on mobile

QA Checklist: https://wiki.corp.adobe.com/display/adobedotcom/M@S+Engineering+QA+Use+Cases

Please do the steps below before submitting your PR for a code review or QA

- [ ] C1. Cover code with Unit Tests
- [ ] C2. Add a Nala test (double check with #fishbags if nala test is needed)
- [ ] C3. Verify all Checks are green (unit tests, nala tests)
- [ ] C4. PR description contains working Test Page link where the feature can be tested
- [ ] C5: you are ready to do a demo from Test Page in PR (bonus: write a working demo script that you'll use on Thursday, you can eventually put in your PR)
- [ ] C.6 read your Jira one more time to validate that you've addressed all AC's and nothing is missing

## 🧪 Nala E2E Tests

Nala tests run automatically when you **open this PR**.

### To run Nala tests again:

1. Add the `run nala` label to this PR (in the right sidebar)
2. Tests will run automatically on the current commit
3. Any future commits will also trigger tests **as long as the label remains**

### To stop automatic Nala tests:

- Remove the `run nala` label

> **Note**: Tests only run on commits if the `run nala` label is present. Add the label whenever you need tests to run on new changes.

## Test URLs:

- Before: https://main--mas--adobecom.aem.live/
- After: https://mwpw-199086--mas--adobecom.aem.live/

PR #1055, opened 2026-07-18
MWPW-200940: Fix ETF text disappearing in mini-compare-chart after full page load

seanchoi0:
When `merch-card-collection` removes and re-inserts `merch-card` elements during initial load ordering, `disconnectedCallbackHook` clears `legalObserver` and `legalResolvedHandler` — but `legalAdjusted` stays `true`. On reconnection, `adjustLegal` is blocked by `legalAdjusted=true` and no observer exists to catch `inline-price` re-renders, causing the ETF text ("Fee applies of half your remaining annual commitment…") to disappear once other components finish loading.

Resolves https://jira.corp.adobe.com/browse/MWPW-200940

Test URLs:

- After: https://mwpw-200940--mas--adobecom.aem.live/

Test page: https://main--da-dc--adobecom.aem.live/drafts/prateeksha/test-placeholder?akamaiLocale=CO&mep=off&maslibs=MWPW-200940


3ch023 (dismissed):
1. there are seem to be some unrelated fixes as well making scope of change bigger. please double check if those are needed and if not, remove them. (existingLegal block, adjustShortDescription, etc)
2. why deleting release-mas-js?

seanchoi0 (commented):
1. Agreed — those changes (existingLegal block, MutationObserver on existingLegal, adjustShortDescription rAF retry / hasIconButton / innerHTML, heading-xs slot) were self-reviewed additions unrelated to the original bug. Will revert them to keep scope focused on the core fix: legalObserver cleanup in disconnectedCallbackHook and the recovery block in connectedCallbackHook.

2. The `.claude/commands/release-mas-js.md` deletion happened during merge conflict resolution — `.claude` is a symlink in this worktree and git rejected the file with "beyond a symbolic link", so `git rm --cached` was used to resolve the conflict. The file will be restored.

web-components/src/variants/mini-compare-chart.js
```diff
@@ -182,7 +208,6 @@ export class MiniCompareChart extends VariantLayout {
 
         const slots = [
             'heading-m',
```
Axelcureno, line 210: Restore `heading-xs` — the AEM mapping still stamps `slot="heading-xs"`, so dropping it from the equalization list loses row-height alignment (#996).
```suggestion
            'heading-m',
            'heading-xs',
```

web-components/src/variants/mini-compare-chart.js
```diff
@@ -832,11 +869,6 @@ export class MiniCompareChart extends VariantLayout {
                 --consonant-merch-card-mini-compare-chart-heading-m-height
             );
         }
```
Axelcureno, line 880: Companion to the slots-array restore — the min-height rule that consumes the equalized height.
```suggestion
        }
        :host([variant='mini-compare-chart']) slot[name='heading-xs'] {
            min-height: var(
                --consonant-merch-card-mini-compare-chart-heading-xs-height
            );
        }
```

web-components/src/variants/mini-compare-chart.js
```diff
@@ -603,18 +642,16 @@ export class MiniCompareChart extends VariantLayout {
             this.shortDescriptionSource = bodyXxs;
             bodyXxs.remove();
         }
-        const source = this.shortDescriptionSource;
-        const text = source.textContent?.trim();
-        const hasIconButton = !!source.querySelector('.icon-button');
-        if (!text && !hasIconButton) return;
+        const text = this.shortDescriptionSource.textContent?.trim();
+        if (!text) return;
```
Axelcureno: Restore the #996 icon-button gate — without it an icon-only `body-xxs` bails out and the tooltip icon is dropped entirely.
```suggestion
        const source = this.shortDescriptionSource;
        const text = source.textContent?.trim();
        const hasIconButton = !!source.querySelector('.icon-button');
        if (!text && !hasIconButton) return;
```

web-components/src/variants/mini-compare-chart.js
```diff
@@ -603,18 +642,16 @@ export class MiniCompareChart extends VariantLayout {
...
         const legalPrice = this.card.querySelector(
             '[slot="heading-m-price"] [data-template="legal"]',
         );
         const planType = legalPrice?.querySelector('.price-plan-type');
         if (!planType) return;
         if (planType.querySelector('em')) return;
         const em = document.createElement('em');
-        em.innerHTML = ` ${source.innerHTML}`;
+        em.textContent = ` ${text}`;
```
Axelcureno: Companion restore — `textContent` flattens icon-buttons/links/sup in ETF copy to plain text. If moving to `textContent` on purpose (it does close an injection sink), better as its own PR with #996 owners + updated tests.
```suggestion
        em.innerHTML = ` ${source.innerHTML}`;
```

Axelcureno (changes requested):
Verified the core fix — collection reordering really does disconnect/reconnect cards (`updated()` → `prepend()` on connected children), and the observer lifecycle is now symmetric across the cycle. Nice isolation of the root cause.

Blocking on two things:

- The last main merge still reverts parts of #996 (MWPW-199086): `heading-xs` slot equalization and the icon-button/`innerHTML` ETF handling (plus its two tests in `merch-card.adjustAddon.test.js`). Left inline suggestions to restore them. If dropping `innerHTML` is intentional — it does close an injection sink — that deserves its own conversation with the #996 folks, with the tests updated rather than deleted.
- The "injects ETF text" test wraps all its assertions in `if (planType)`, so it can pass while asserting nothing — and the fixture's `data-wcs-osi="abm"` has no key in `mocks/offers.json` (only `abm-mult` etc.), which is likely why. A real mock OSI + unconditional assertions would make the new tests actually bite.

Smaller: if `legal.onceSettled()` rejects after the clone is inserted, `legalAdjusted` is already `true`, so the "allow retry" comment in the catch never happens — and a naive retry would duplicate the legal node since it's already in the DOM. Worth cleaning up the inserted clone in the catch. Also the `delay(50)`s / `document.__etfMount` stash could use the file's `checkReady()` pattern, and Nala Gate is red on this PR.

seanchoi0 (commented):
Thanks for the thorough review!

- Restored #996 changes (heading-xs slot, icon-button/innerHTML) that were accidentally reverted — those belonged to MWPW-199086, not this ticket's scope.
- catch block now removes the inserted clone and resets legalAdjusted/legalResolvedHandler/legalElement so retry actually works.
- ETF test: switched to abm-mult (valid OSI in offers.json), replaced delay(50) with card.checkReady(), made planType assertion unconditional, dropped document.__etfMount stash.

Axelcureno (approved):
All good — verified the #996 restore on the branch (`heading-xs` + `hasIconButton`/`innerHTML` back, adjustAddon tests intact), and the catch cleanup handles both failure points correctly: pre-insertion failures leave `legal` undefined, post-insertion failures remove the clone and reset state so the retry actually fires now. Test fixes look right too — `abm-mult` resolves against the mocks, assertions are unconditional, and `checkReady()` replaces the delays.

One thing left: Nala Gate is only failing on the "Require `run nala` label" step — adding the label should turn it green. Non-blocking: the catch-cleanup path has no test yet, fine as a follow-up.

web-components/src/variants/mini-compare-chart.js
```diff
@@ -70,6 +70,30 @@ export class MiniCompareChart extends VariantLayout {
             this.updatePriceQuantity,
         );
 
+        if (this.legalAdjusted && !this.legalObserver) {
+            const legal = this.card.querySelector(
+                '[slot="heading-m-price"] [is="inline-price"][data-template="legal"]',
+            );
+            if (legal) {
```
npeltier, line 77: isn't `[data-template="legal"]` or `[is="inline-price"][data-template="legal"]` enough as a selector? or do you want only legal templates in a given slot? what if for some reason we change that slot?

npeltier (approved):
lgtm

PR #1056, opened 2026-07-19
MWPW-201030: Allow creating a promo variation without geos (inherit promo project's geos)

mirafedas:
The promotion project may have only one promotion variation with no geos specified - this variation automatically applies to all promotion project geos (for those added later as well).
When resolving which promotion variation to show, we first check if the promotion project has the variation for the specific geo. If not, we use the content from the promotion variation with no geos.

Resolves https://jira.corp.adobe.com/browse/MWPW-201030

## 🧪 Nala E2E Tests

### To run Nala tests again:

- Remove the `run nala` label

## Test URLs:

- After: https://mwpw-201030--mas--adobecom.aem.live/

### DIFF
--- a/studio/src/promotions/mas-promo-variation-geos.css.js
+++ b/studio/src/promotions/mas-promo-variation-geos.css.js
@@ -53,6 +53,16 @@ export const styles = css`
         margin: 0;
     }
 
+    .inherit-hint {
+        color: var(--spectrum-gray-700, #505050);
+        font-size: 12px;
+        margin: 0;
+    }
+
+    .inherit-hint.blocked {
+        color: var(--spectrum-negative-color, #d7373f);
+    }
+
     :host([compact]) .select-geos-content {
         min-width: 220px;
         gap: 8px;
--- a/studio/src/promotions/mas-promo-variation-geos.js
+++ b/studio/src/promotions/mas-promo-variation-geos.js
@@ -7,6 +7,7 @@ class MasPromoVariationGeos extends LitElement {
     static properties = {
         geos: { type: Array },
         disabledGeos: { type: Array },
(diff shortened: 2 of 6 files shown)


3ch023 (approved):

afmicka (approved):

honstar:
I didn't have a chance to add my findings before this PR got merged unfortunately, but I'll add it here nonetheless to serve as a basis for a follow-up ticket:

main issue is that the "one geo-less variation" rule isn't enforced at the service layer - only in transient UI state. The [`findOverlappingGeoTags`](https://github.com/adobecom/mas/blob/main/studio/src/promotions/promotion-variations.js#L112) function is structurally incapable of detecting a "geo-less sibling already exists". It only ever compared geo lists against each other, never presence/absence of the empty-geo case. Nothing in `createPromoVariation` itself asks "does existingVariations already contain an entry with empty pznTags?"

`fragmentHasEmptyGeosVariation` is captured once when the dialog opens and never re-probed before the actual `createPromoVariation()` call fires. Between dialog-open and Continue-click there's an unguarded window: a second browser tab, another user, or even the same user re-opening a stale dialog can create the geo-less sibling in that interval, and this code has no way to notice — it just replays the stale boolean.

So there is a possibility to create more than just one geo-less variation which should not be possible.

*Suggested fix*
Move the invariant into the `createPromotionVariation` itself, next to the existing overlap check, so it's enforced regardless of the caller/UI state:
```
if (!geoTags.length && existingVariations.some((v) => !v.pznTags?.length)) {
    throw new UserFriendlyError('A variation with no geos already exists for this project.');
}
```

mirafedas:
@honstar thank you, I created a follow-up Jira: https://jira.corp.adobe.com/browse/MWPW-201344

PR #1061, opened 2026-07-21
MWPW-201122: Extend cancel-context to OSI substitutions

seanchoi0:
When a promo project sets `cancel-context` as the `substituteOsi` for a base OSI, the WCS transformer now skips the substitution — preserving the original OSI in the HTML body, `fields.osi`, and all WCS token lookups. This lets authors opt out of promo-project OSI replacement on a specific price using the same sentinel already used for promo-code cancellation.

**Note:** This change affects OSI replacement only. Wildcard promoCode from the promo project is a separate concern and is controlled via the existing `cancel-context` promotionCode mechanism in OST.

Resolves https://jira.corp.adobe.com/browse/MWPW-201122

Test URLs:

- After: https://mwpw-201122--mas--adobecom.aem.live/

Test Page (cancel-context OSI substitution):

- https://milo.adobe.com/merch/mas/preview?fragment-id=abaf37fc-dfc4-43d1-a101-f4778245c3eb&content-type=merch-card&locale=en_NG&instant=2027-07-15T05:00:00.000Z&debug.io&clearCaches.io&mas.preview=on&mas-io-url=https://14257-merchatscale-seanchoi.adobeioruntime.net/api/v1/web/MerchAtScale/fragment


npeltier (changes requested):
i don't understand the use of this: rather than doing this, you could just remove that osi from the substitute map, couldn't you?
to me use case here is that you have 123 fragments in a project using an OSI you want to substitute. For some reason, for one of them you want to keep old OSI, but still keep it in the project because of other change.
Then you somehow set cancel context there (needs to be differnentiable from promo context cancel), and you are good.

seanchoi0:
Thanks for the review, @npeltier!

You're right that for a single promo project, omitting the entry would achieve the same result. But would the multi-project case be a problem? When multiple promo projects are merged via `Object.assign({}, ...promoProjects.map(p => p.substituteMap))`, if project A defines `substitute|BASE-OSI|PROMO-OSI|US`, simply not having an entry for `BASE-OSI` in project B wouldn't cancel A's substitution — `Object.assign` would keep A's value. The `cancel-context` sentinel would allow project B to explicitly override it.

This seemed consistent with how the existing `cancel-context` for promo codes works — is that the same rationale, or am I misreading the intent there?

seanchoi0:
Closing this PR — the implementation approach needs to change based on updated requirements. The correct solution is a per-price-element 'Lock OSI' mechanism (data-locked-osi="true" attribute) rather than a project-level cancel-context sentinel. Will reopen with the correct implementation.

PR #1063, opened 2026-07-21
MWPW-201374: fix run-nala AIO namespace (lowercase login + per-user override)

Axelcureno:
Fixes the io/www NALA step deriving a dead `mas-io-url` when a PR author's GitHub login has uppercase or differs from their aio namespace suffix.

- `run-nala.yml` reconstructed the namespace as `AIO_WWW_NS-<github-login>`, using the login verbatim. AIO namespaces are lowercase and a login can differ from the namespace suffix, so `Axelcureno` (namespace `axel`) got a dead URL and Docs NALA hung to the 20-min timeout.
- Switch to the full per-user `AIO_NS_<login>` secret already used by `io-pr.yaml` (deploy) and `io-tests.yaml` (health check) — no prefix concat, no case assumptions, no override branch, no new secret.

Surfaced on #1051 (MWPW-200909), whose Docs NALA was blocked by this.

Resolves https://jira.corp.adobe.com/browse/MWPW-201374

- [ ] C3. Verify all Checks are green
- [ ] C.6 re-read Jira

Test URLs:

- After: n/a (CI workflow change; validated by an io/www PR's Docs NALA resolving the correct namespace)


Axelcureno:
how it works, step by step:

1. github.event.pull_request.user.login → Axelcureno (your GitHub login, capital A).
2. format('AIO_NS_{0}', 'Axelcureno') → the string AIO_NS_Axelcureno.
3. secrets[...] looks up a secret by that name. GitHub secret lookups are case-insensitive, so AIO_NS_Axelcureno resolves to your stored secret AIO_NS_AXELCURENO.
4. That secret's value is 14257-merchatscale-axel (your real, full namespace — set back on 2026-07-17, and already proven correct because your test-deployed/health-check job uses this same secret and passes).
5. So ACTOR_NS = 14257-merchatscale-axel — the correct value, verbatim.
6. The final line builds:
?mas-io-url=https://14257-merchatscale-axel.adobeioruntime.net/api/v1/web/MerchAtScale
6. → a live endpoint (your deployed io action), so the Docs NALA tests actually reach it instead of hanging on a dead URL.

afmicka:
@Axelcureno please add some dummy change to the io/www file to trigger the docs test with mas-io-url to verify this (it will need run nala label). If it starts picking up the correct namespace, remove the io change 

I understood that you are going to create a new namespace but that it did not allow capital later. If you are going to use 'axel' namespace after all, we need to make sure that both io deployment and Nala are taking the same namespace. There should not have been a reason for io deployment and Nala to use different namespaces earlier but could be.

Axelcureno:
Pushed a temp `io/www` change (bbe626869) to trigger the Docs NALA.

On the consistency point: this PR fixes exactly that. Both now read the same secret:
- **Deploy** (`io-pr.yaml` L49): `AIO_NS_<login>`
- **NALA** (`run-nala.yml`, this PR): `AIO_NS_<login>`

Before, NALA used `AIO_WWW_NS-<login>` (reconstructed) while deploy used the full `AIO_NS_<login>` — that was the mismatch. For me: `AIO_NS_AXELCURENO=14257-merchatscale-axel` (my real namespace; deploy/test-deployed already pass with it). No separate namespace anymore.

Blainegunn (approved):

antonio-rmrz (approved):

PR #1069, opened 2026-07-22
MWPW-201011: Geo Promo Variations - Follow up

Andrei4226:
Resolves https://jira.corp.adobe.com/browse/MWPW-201011

Bug fixes — geo-scoped promo variations:
- Fixed `probePromoVariationsForFragment` to probe all variation indices instead of stopping at the first missing one, previously, deleting one variation could make higher-indexed sibling variations "disappear" from Studio's promo-variation list, even though they still existed in AEM.
- Fixed `getNextAvailablePromoVariationIndex` to fill index gaps left by individual deletions, instead of always assuming the next index is existingCount + 1.
- Fixed `createPromoVariation` to force-refresh the promotions store before reading attached fragment paths, so geo-collision checks don't silently no-op when the store hasn't loaded yet.
- Fixed the tie-break logic in `resolveDefaultFragmentForPromoVariation` to correctly prefer the non-suffixed candidate path over a coincidentally-named suffixed fragment.
- Removed `deleteAttachedPromoVariations` and tests.

Code quality (review follow-ups):
- Extracted a shared `geoMatchScore` helper, used by both the personalization (pzn) and promo-variation geo matching logic in `customize.js`, removing duplicated scoring formulas.
- Documented (via comments) the unavoidable duplication of `MAX_PROMO_VARIATIONS_PER_FRAGMENT` between the Studio and io/www runtimes (separate deployables, no shared import path).
- Deduplicated the "`search + select-all + indeterminate checkbox`" list UI logic shared by `mas-promo-variation-geos.js` and `mas-translation-languages.js` into a new shared module (`studio/src/common/utils/selectable-list.js`), with its own dedicated unit tests.
- Fixed a coding-convention issue: `displayPromoVariationGeos()` method converted to a getter (`promoVariationGeosTemplate`), matching the sibling `promoVariationGeoTagsTemplate` getter pattern.

Test hygiene:
- Fixed a broken test (missing `loadPromotions` mock on the repository stub).
- Renamed misleading test titles that still referenced the already-removed "`View promo variation`" UI link.

- `probePromoVariationsForFragment` used to check for existing promo variations by probing up to 50 individual paths one by one (`getByPath` for card, card-2, card-3, ...). Now it does a single folder search (`aem.sites.cf.fragments.search({ path: promoFolder })`) on the promo folder, then matches results locally via a regex derived from the fragment's own leaf name to extract each variation's index — with a guard so a sibling fragment's own name (e.g. card-2 as a separate attached card) never gets misattributed as another card's suffixed variation.

- Also fixed: `getNextAvailablePromoVariationIndex` now skips already-used indices instead of assuming they're contiguous (handles gaps from deleted variations), and `collectAttachedPromoVariations` batches multiple fragments sharing the same promo folder into one search call instead of one per fragment.

Test-side: all AEM mocks that used to stub `getByPath` for probing were converted to stub search as an async generator keyed by folder path — duplicated across 6+ test files, so extracted into a shared `makeSearchStub` helper in the existing `test/helpers/aem-tag-fetch.js`.

About the idea to combine `promotion-model.js` and `promotion-repository.js`:
- Different purpose: `promotion-model.js` = pure functions, no state; `promotions-repository.js` = Store-aware orchestration.
- Different consumers: the former is used by 14 files across the whole studio codebase (many unrelated to promotions); the latter only by 7, strictly within the promotions feature.
- Coupling risk: `aem/fragment.js` (a base file, used everywhere) would become dependent on Store plus the entire promotions feature, despite having nothing to do with it — any code using Fragment would risk breaking because of an unrelated feature.

Test:

Created a promo project: https://mwpw-201011--mas--adobecom.aem.live/studio.html#page=promotions-editor&path=sandbox&promotionId=99cdf5cd-326b-48ec-b4a6-d43af7730314
For these 3 geos: fr_BE, fr_CA and fr_CH 
Created 3 promo variation in this order: fr_BE, fr_CA and fr_CH for this fragment:
https://mas.adobe.com/studio.html#fragmentId=e01d7cbd-c70c-486f-b5a7-5a7272380f88&page=fragment-editor&path=sandbox
Removed the second promo variation(fr_CA).

In main, it looks like this(with issue):
image

In my branch, it looks like this (including the promo variation with the index greater than deleted promo variation):
image

Before: https://mas.adobe.com/studio.html#content-type=merch-card&page=content&path=sandbox&query=e01d7cbd-c70c-486f-b5a7-5a7272380f88

After: 
https://mwpw-201011--mas--adobecom.aem.live/studio.html#content-type=merch-card&page=content&path=sandbox&query=e01d7cbd-c70c-486f-b5a7-5a7272380f88

## 🧪 Nala E2E Tests

### To run Nala tests again:

- Remove the `run nala` label

## Test URLs:

- After: https://mwpw-201011--mas--adobecom.aem.live/


studio/src/common/utils/selectable-list.js
```diff
@@ -0,0 +1,59 @@
...
+ * (e.g. mas-promo-variation-geos, mas-translation-languages).
+ */
+
+/**
+ * Adds the search-box wiring shared by list-selector components (search query state,
+ * input handler, substring filter). Host component still owns its own `searchQuery`
+ * reactive property declaration and initial value.
+ * @param {typeof import('lit').LitElement} Base
+ */
+export const SearchableListMixin = (Base) =>
```
mirafedas: Do we need a mixin? It looks to me we could have just plain utility functions, and pass this.searchQuery as a param to filterBySearchQuery. I think it will decrease complexity (debugging mixins is harder than plain functions)

studio/src/promotions/promotion-variations.js
```diff
@@ -27,24 +29,28 @@ function readPznTags(fragment) {
...
+    const candidates = [];
     for (let index = 1; index <= MAX_PROMO_VARIATIONS_PER_FRAGMENT; index += 1) {
         const suffixIndex = index === 1 ? undefined : index;
         const targetPath = buildPromoVariationPathForTag(defaultPath, promoTagId, suffixIndex);
         if (!targetPath) break;
-        const variation = await getFragmentByPathOrNull(aem.sites.cf.fragments, targetPath);
-        if (!variation?.id) break;
+        candidates.push({ path: targetPath, index });
+    }
+    const variations = await Promise.all(candidates.map(({ path }) => getFragmentByPathOrNull(aem.sites.cf.fragments, path)));
```
mirafedas: Here we will be sending up to 50 parallel requests for every variation - can we consider sending one request to get all fragments in the folder, and then filter out variations? Or some other optimization

mirafedas (changes requested):
Good job 👍

studio/src/common/utils/selectable-list.js
Andrei4226: Done, replaced `SearchableListMixin` with plain functions (`filterBySearchQuery, handleSearchInput`) in `selectable-list.js`. Components call them directly, no shared base class.

studio/src/promotions/promotion-variations.js
Andrei4226: Done, `probePromoVariationsForFragment` now does a single folder search (`aem.sites.cf.fragments.search`) instead of probing each index individually, then matches results locally by leaf name.

mirafedas (approved):

npeltier (approved):
lgtm

afmicka:
@Andrei4226 from the description i understood that you are removing the feature to delete the promo variations when the project is deleted. However, my promo variations HAVE been deleted regardless. Do i need any additional parameter in the URL to test your changes?
Screenshot 2026-08-04 at 14 15 08


We should also remove this warning that the variations are going to be deleted. 
Screenshot 2026-08-04 at 14 13 35

PR #1070, opened 2026-07-22
MWPW-201344: Creating a promo variation without geos - follow-up

mirafedas:
Added a guard to prevent creating a duplicate of an empty-geos promotions variation (there may be only one).

Resolves https://jira.corp.adobe.com/browse/MWPW-201344

## 🧪 Nala E2E Tests

### To run Nala tests again:

- Remove the `run nala` label

## Test URLs:

- After: https://mwpw-201344--mas--adobecom.aem.live/


studio/src/promotions/promotion-variations.js
```diff
@@ -27,35 +31,113 @@ function readPznTags(fragment) {
...
+/**
+ * Probes promo variations for multiple fragments sharing the same promo tag. Fragments whose
+ * variation paths resolve to the same parent folder (e.g. siblings in the same source directory)
+ * are scanned via a single search instead of one per fragment. Results are ordered by index ascending.
+ * @param {import('../aem/aem.js').AEM} aem
+ * @param {string[]} defaultPaths
+ * @param {string} promoTagId
+ * @returns {Promise<Map<string, Array<{ path: string, index: number, id: string, pznTags: string[], status: string, title: string, model: string, fields: Array, tags: Array }>>>}
+ */
+export async function probePromoVariationsForFragments(aem, defaultPaths, promoTagId) {
```
Andrei4226: I saw that `computeProbeTarget` builds a `leafPattern` per fragment, whereas `probePromoVariationsForFragments` groups multiple fragments attached to the same promotion into a single search. 
But if two distinct fragments attached to the same promotion have names like "card" and "card-2", I think the regex for "card" will also match the result for "card-2" and would incorrectly attribute the variation (previously the full path was used and this case couldn't happen)
Optionally, a test to catch this case would be good

Andrei4226 (changes requested):

studio/src/promotions/promotion-variations.js
mirafedas: Very good catch, I haven't thought about this case, thank you :)

mirafedas: @Andrei4226 actually I just saw you fixed the same things in https://github.com/adobecom/mas/pull/1069/ , and our fixes look different. I will remove my fix from this PR

Andrei4226: yes, it's a refactor for the delete part
the common part is part of reading and keeps the search by `getByPath`, I forgot it
plus `getNextAvailablePromoVariationIndex`

Andrei4226 (approved):
LGTM

npeltier (approved):

Axelcureno (approved):
LGTM. The geo-less guard closes the gap the overlap check can't catch — empty geo arrays never register as "overlapping," so without this you could stack two fallback variations. Field access lines up with the probe output, and all three changes have tests. Nice.

One nit: the inline comment in `mas-promotions-items-table.test.js` (~line 110) explaining the disconnect guard — repo style avoids inline comments even in tests. Not blocking.

PR #1076, opened 2026-07-23
MWPW-201713: Apply mas-field promo code to checkout options

Axelcureno:
Resolves https://jira.corp.adobe.com/browse/MWPW-201713 (remaining scope; the io/www rich text OSI promo matching was folded into #1089 with per fragment scoping)

mas-field registers only a price options provider, so CTAs rendered through mas-field resolved checkout URLs without the promo project code. The shopper saw the promo price but checked out at the base price. Adds a checkout options provider mirroring merch-card's; the enclosing mas-field's data-promotion-code is applied unless the element already carries its own.

## Test URLs:

- Before: https://main--da-cc--adobecom.aem.page/products/photoshop?instant=2026-08-19T00:00:00.000Z&mas.preview=on&clearCaches.io
- After: https://main--da-cc--adobecom.aem.page/products/photoshop?instant=2026-08-19T00:00:00.000Z&mas.preview=on&clearCaches.io&maslibs=mwpw-201713&milolibs=mwpw-201713
- Before: https://main--da-cc--adobecom.aem.live/cc-shared/fragments/merch/products/indesign/compare-plans/table/individual/default
- After: https://main--da-cc--adobecom.aem.live/cc-shared/fragments/merch/products/indesign/compare-plans/table/individual/default?maslibs=MWPW-201713
- Before: https://main--da-cc--adobecom.aem.live/drafts/npeltier/table
- After: https://main--da-cc--adobecom.aem.live/drafts/npeltier/table?maslibs=MWPW-201713

## Checklist:

- [x] Tests pass locally (web-components 1545 passing, incl. 3 new provider tests)


Axelcureno:
Related Milo PR for the same ticket: https://github.com/adobecom/milo/pull/6339 — stamps the promo code and loads price styles when merch-card-autoblock unwraps a mas-field.

io/www/src/fragment/transformers/customize.js
```diff
@@ -295,10 +295,30 @@ function mergeVariations(root, customizeContext) {
...
+function collectRichTextOsis(fields) {
+    const osis = [];
+    for (const value of Object.values(fields)) {
+        if (typeof value !== 'string' || !value.includes('data-wcs-osi')) continue;
+        for (const match of value.matchAll(RICH_TEXT_OSI_ATTR)) {
+            osis.push(...match[1].split(','));
+        }
+    }
+    return osis;
+}
```
npeltier: this operation is already done in a different way in later wcs transformer. This is doubling the text processing here for same purpose. 
applyPromoCode runs at the only moment we have finalized output (with customized OSIs), so we can't run it before. 
We should probably though save a map osis -> fragment id, or path, or json path, but i guess this can be done later and we live with that double processing for now.

npeltier (changes requested):
- remark on the code is not for that PR
- please provide mas.io related test URL, also please tell what changes as i can't see it on the preview part
- also we should have explicit logic (and test) when there is a promo code on fields.osi & one in inline osi that is different. I guess this needs PM decision there on what to do (i would make fields.osi prior), and be explicit we do that

io/www/src/fragment/transformers/customize.js
Axelcureno: Ok, yes I agree. `applyPromoCode` needs the finalized customized output. I see you have filed MWPW-201862 as the follow up (thank you!).

Axelcureno: - osi priority: the fragment's own osi is collected first, so an explicit `fields.osi` match always wins over rich text OSIs. Covered by the "prefer the fragment osi match over rich text OSI matches" test. Can make it louder in code if you prefer, lmk.
- mas.io URL added to the description with a note on what shows where: the server pipeline applies the promo only once the window opens (Aug 17), so the visible delta is on the preview URL.

npeltier (changes requested):
sorry @Axelcureno, big collision with https://github.com/adobecom/mas/pull/1089, i guess you should remove io/www part, and just keep here the mas-field related changes (maybe some link to @yesil 's tech debt ticket as this is yet another time this is hitting us)

npeltier (approved):

web-components/src/mas-field.js
```diff
@@ -27,10 +27,26 @@ export function priceOptionsProvider(element, options) {
...
+/**
+ * Applies the enclosing mas-field's promo code to checkout options,
+ * mirroring what merch-card's checkout options provider does for cards.
+ * Without this, CTAs rendered through <mas-field field="ctas"> resolve
+ * checkout URLs without the promotion applied by a promo project.
+ */
+export function checkoutOptionsProvider(element, options) {
+    const masField = element?.closest?.(MAS_FIELD_TAG);
+    if (!masField) return options;
+    if (!options.promotionCode) {
```
yesil, line 70: this is not leveraging the COMPAT_VERSION_GLOBAL_PROMO_CODE, if applied old fragments (not sure realistic since it is for mas-field) can cause issues.

npeltier: i think there was no headless card prior to this.

web-components/src/mas-field.js
```diff
@@ -182,7 +199,11 @@ class MasField extends HTMLElement {
                 'data-promotion-variation-project',
                 fragment.promoVariationProject,
             );
-        if (fragment.fields?.promoCode)
+        if (
+            fragment.fields?.promoCode &&
+            (fragment.fields.compatVersion >= COMPAT_VERSION_GLOBAL_PROMO_CODE ||
```
yesil: in merch-card.js this logic is in the optionProvider functions.
I think this should move into either of priceOptionsProvider and checkoutOptionsProvider.

yesil (approved):

afmicka (approved):

PR #1077, opened 2026-07-23
MWPW-201497: mas-field: style indexed CTAs and emit mas:ready

joaquinrivero:
Resolves https://jira.corp.adobe.com/browse/MWPW-201497

## 🧪 Nala E2E Tests

### To run Nala tests again:

- Remove the `run nala` label

## Test URLs:

- After: https://mwpw-201497--mas--adobecom.aem.live/

### Test Page (maslibs=MWPW-201497):

- Product page (throttle to Slow 3G): https://www.stage.adobe.com/products/photoshop.html?maslibs=MWPW-201497&milolibs=MWPW-201497
- https://stage--da-cc--adobecom.aem.page/ca/drafts/ravneet/dotcom-188566/downloads-test?maslibs=MWPW-201497&milolibs=MWPW-201497


Blainegunn (approved):
bravo

Blainegunn:
i see it working now in all viewports

TsayAdobe (approved):

antonio-rmrz:
**Problem:** Since we now rebuild the indexed CTA instead of reusing the stored anchor, if the checkout-link/service isn't available the button renders styled but dead — no href, never hydrates. The old code degraded to an unstyled but working link. Probably only theoretical since aem:load can't fire without mas.js, but cheap to guard against.

**Potential solve:** In `#buildCtaButton`, when `createCheckoutLink` returns null, fall back to `link.cloneNode(true)` instead of building a bare anchor — that keeps `href` and `is` so the link still works and can upgrade later:

```js
const button =
    CheckoutLink?.createCheckoutLink(link.dataset, link.textContent) ??
    link.cloneNode(true);
```

joaquinrivero:
@antonio-rmrz Good catch, fixed in dfcfc07f. Fallback now clones the link, so `href`/`is`/text survive.

Roycethan (commented):
@joaquinrivero Im still able to reproduce the problem on lower network throttles like on VMs or VPN's with lower throttles:
https://www.stage.adobe.com/products/photoshop.html?maslibs=MWPW-201497

Image

antonio-rmrz (approved):

joaquinrivero:
I did some updates that only adds the event @Blainegunn @TsayAdobe @Roycethan @antonio-rmrz and open this PR on milo: https://github.com/adobecom/milo/pull/6341

Roycethan (approved):

PR #1078, opened 2026-07-23
MWPW-200587: Rename bizpro variant to pro

antonio-rmrz:
Resolves https://jira.corp.adobe.com/browse/MWPW-200587
QA Checklist: https://wiki.corp.adobe.com/display/adobedotcom/M@S+Engineering+QA+Use+Cases

Renames the `bizpro` variant to `pro` so CPro cards use the same template. Full rename across web-components (files, class, fragment mapping, CSS tokens/selectors), Studio (picker, editor gates, whats-included util) and Nala.

Backwards compat until content is migrated: hydrate coerces stored `bizpro` to `pro`, and Studio keeps legacy fragments visible and rewrites them to `pro` on save. Both spots are marked `TODO(MWPW-200587)` and get removed once no `bizpro` fragments remain. Settings fragments with `bizpro` in their `templates` lists need updating before authoring `pro` cards.

Also fixes two latent issues while in there: the row-expand selector was hardcoded to `bizpro`, and the height-sync CSS property name was built dynamically on the producer side but read as a literal in the shadow CSS.

- [x] C1. Cover code with Unit Tests
- [x] C2. Add a Nala test (double check with #fishbags if nala test is needed)
- [ ] C3. Verify all Checks are green (unit tests, nala tests)
- [ ] C4. PR description contains working Test Page link where the feature can be tested
- [ ] C5: you are ready to do a demo from Test Page in PR (bonus: write a working demo script that you'll use on Thursday, you can eventually put in your PR)
- [ ] C.6 read your Jira one more time to validate that you've addressed all AC's and nothing is missing

Note: the Nala suite was renamed to `acom/pro` but the QA fragment still has to be re-saved as `pro` once QA author access is sorted — specs carry a TODO.

## Test URLs:

- Before: https://main--mas--adobecom.aem.live/
- After: https://mwpw-200587--mas--adobecom.aem.live/

PR #1093, opened 2026-07-27
MWPW-201873: Add Theme and Size options to the Pro card editor

antonio-rmrz:
Resolves https://jira.corp.adobe.com/browse/MWPW-201873

Adds two dropdowns to the Pro card editor in M@S Studio so authors can start marking content for the upcoming dark and EDU treatments:

- **Theme** — Light or Dark. We can't add new fields to the Card model right now, so the selection is stored in the card's existing `backgroundColor` field. Picking Dark stamps `background-color="dark"` on the rendered card as a marker; the actual dark styling ships separately with MWPW-200406.
- **Size** — Default, Wide or Edu. Uses the existing `size` field and the editor's existing Size dropdown, which simply wasn't enabled for Pro before. Wide makes the card span two columns today (same as other variants); Edu is stored now and gets its look with MWPW-200407.

Nothing changes visually with this PR — cards look exactly the same until the rendering tickets land. No new model fields, and no changes to hydration, IO, or Milo.

**Every change in this branch, in plain terms:**

- `web-components/src/variants/pro.js` — the Pro card now declares the two options: `size: ['wide', 'edu']`, plus a `backgroundColor` entry labeled "Theme" with Light/Dark values. Declaring them is what makes the dropdowns appear in the editor.
- `studio/src/editors/merch-card-editor.js` — when a variant declares those named theme values, the Background Color control renders as a simple two-option "Theme" picker (no "Transparent"). Every other variant keeps the exact same Background Color control as before. If a card already holds an unexpected stored value, the dropdown shows that raw value instead of pretending it's Light.
- `studio/test/editors/merch-card-editor-pro.test.js` — unit tests: both dropdowns appear for Pro with exactly the right options, selections persist to the fragment, other variants are untouched, and unknown stored values aren't masked.
- `web-components/test/merch-card.pro.test.js` — locks the Pro mapping shape so the option lists can't drift silently.
- `nala/studio/acom/pro/specs/pro_save.spec.js` + `tests/pro_save.test.js` — E2E flow: pick Dark and Edu, save, verify both the editor values and the attributes on the rendered card. Written but not yet run (QA author access still pending, same as #1078).
- `web-components/dist/mas.js`, `merch-card.js`, `merch-card-collection.js` — regenerated bundles (one line each).

Follow-up to #1078 (bizpro → pro rename), which merged today; this branch is rebased on top of it.

**Before merging** we should audit existing pro/bizpro fragments for stray stored `size` / `backgroundColor` values: a stored `wide` would start spanning two columns the moment this ships, and a stored color value would start appearing on the card as an attribute.

- [x] C2. Add a Nala test (written; run pending QA author access)
- [x] C.6 read your Jira one more time to validate that you've addressed all AC's and nothing is missing — AC1 (select dark theme) and AC2 (select wide/edu style) both authorable; final labels ("Edu" vs "EDU") pending the ticket mockup

## Test URLs:

- After: https://mwpw-201873--mas--adobecom.aem.live/


joaquinrivero (approved):

Blainegunn (approved):
génial!

Roycethan (approved):

PR #1094, opened 2026-07-28
MWPW-202151: Validate maslibs and mas-io-url params (VULN-36379)

Axelcureno:
Fixes the MAS side of VULN-36379: the `maslibs` URL parameter was interpolated unvalidated into a remote-script origin, and `mas-io-url` was an unvalidated fetch base carrying the WCS API key.

- Add `getValidatedMasLibsUrl()`: strict `branch` / `branch--repo` / `branch--repo--owner` pattern + `.aem.live`/`.aem.page` hostname assertion; invalid → null → caller default
- Add `isAllowedMasIOUrl()`: hostname allowlist (adobe.com, adobeioruntime.net, aem.live, aem.page; http only for localhost); invalid → default `/mas/io` endpoint
- `aem-fragment.js` and `settings.js` delegate to the validators
- Milo-side fix: separate PR on adobecom/milo, same branch name

Verify: on the After URL, append `?maslibs=evil.com` or `?mas-io-url=https://evil.com` — the network tab shows no request to evil.com; defaults load instead.

Milo counterpart PR: https://github.com/adobecom/milo/pull/6364

Resolves https://jira.corp.adobe.com/browse/MWPW-202151

Test URLs:

- After: https://mwpw-202151--mas--adobecom.aem.live/


web-components/src/utils.js
```diff
@@ -141,3 +141,63 @@ export function shouldHideStPriceLabels(element) {
...
+ * Checks that a mas-io-url value points to an Adobe-controlled host.
+ * The URL becomes the base of fragment requests carrying the WCS api key,
+ * so an attacker-controlled host would leak the key.
+ * @param {string} urlString
+ * @returns {boolean}
+ */
+export function isAllowedMasIOUrl(urlString) {
+    try {
+        const url = new URL(urlString);
+        if (MAS_IO_LOCAL_HOSTS.includes(url.hostname)) return true;
```
honstar: PR description states that only http is allowed for `localhost` but the schema check is entirely skipped in that case. To be consistent, I'd suggest to add a protocol check here as well.

web-components/src/utils.js
```diff
@@ -141,3 +141,63 @@ export function shouldHideStPriceLabels(element) {
...
+
+/**
+ * Validates the maslibs parameter and returns the base URL for MAS libraries.
+ * Only branch, branch--repo and branch--repo--owner shapes are allowed, so
+ * the resulting host always stays under aem.live / aem.page.
+ * @param {string} masLibs raw maslibs parameter value
+ * @param {string} extension aem domain extension: 'live' (default) or 'page'
+ * @returns {string|null} base URL, or null if the value is missing or invalid
+ */
+export function getValidatedMasLibsUrl(masLibs, extension = 'live') {
```
honstar, line 158: The `extension` argument is trusted but not validated. Since we only allow either `live` or `page` as extension, I think it would be safer to validate the extension within this function directly instead of relying on caller discipline.

honstar (changes requested):

web-components/src/utils.js
Axelcureno: Good catch, fixed. The localhost branch now returns `url.protocol === 'http:' || url.protocol === 'https:'` instead of a bare `true`, so `ftp://localhost/mas/io` and friends are rejected. Added a test covering ftp/ws/file on both localhost and 127.0.0.1.

web-components/src/utils.js
Axelcureno: Agreed, done. `getValidatedMasLibsUrl` now checks `extension` against `['live', 'page']` up front and returns null otherwise. Worth noting this was a real hole and not just caller discipline: passing `evil.com` produced `https://main--mas--adobecom.aem.evil.com`. Not reachable from the current caller, but it is now closed at the function. Test added.

honstar (approved):

seanchoi0 (approved):
Nice work on this — adding proper allowlist validation where there was previously none is the right call, and the core logic is solid. Validation runs synchronously before any URL is consumed, so no race conditions to worry about. Approving, but a few things worth a look before this lands:

- **Unit test is failing**

A couple of smaller notes if you have a moment:

- The `maslibs=local` path causes `aem-fragment.js:517` to `import()` from `http://localhost:3000`. That's intentional for dev, but it means anyone with `allowOverride=true` and a local server running could load arbitrary JS from that origin if pointed at a crafted URL. Might be worth an explicit author/preview-context check rather than relying on `allowOverride` alone.

- The `.page` extension branch in `getFragmentClientUrl` is covered at the unit level (`getValidatedMasLibsUrl`), but the wiring through `getFragmentClientUrl` itself doesn't have an integration test for the `.page` case. Worth adding a targeted test given the security context here.

- Minor: `MASLIBS_PATTERN` uses nested quantifiers — in practice the 100-char pre-check limits ReDoS risk in V8, but a `split('--')` approach would be simpler and eliminate the concern entirely.

mirafedas (approved):

afmicka (approved):

PR #1101, opened 2026-07-29
MWPW-202263 reduce promotion project folder listing intensity

npeltier:
## Promotions listing: kill the folder-fetch herd

The old promotions cache was a **single global entry with a fixed 5-min TTL**. Every concurrent
request that missed it — at cold start and at every 5-min expiry, in lockstep across the whole
container fleet — fired its own paginated `/content/dam/mas/promotions` fetch to Odin. That herd
overloads Odin, which returns 504s, which trigger retries, which add more load: a self-reinforcing
loop. This PR makes the read **per-surface + stale-while-revalidate + single-flight** with a
**jittered TTL**.

### The problem, in prod numbers (7 days, 116,870 folder-fetch log lines)

```mermaid
pie showData
    title Promo-folder fetch outcomes — only 1 in 5 succeeds
    "200 OK" : 23568
    "504 + retry (#0/#1/#2)" : 46651
    "Timed out (gave up)" : 46651
```

- **Only 20% of folder fetches succeed.** The other 80% are 504s (retried 3× with 50/100/200 ms
  backoff) or final timeouts.
- **15,115 requests dropped promotions entirely** over the week (timed out, no success). Retries
  recovered just 1,025.
- Even successful listings are slow: p50 83 ms, **p95 877 ms, p99 1.7 s**.
- Fully synchronized & bursty: median 2/min, but peaks of **160 identical fetches/second** and
  **3,547/minute** as the fixed TTL expires fleet-wide at once.

### Before / after load test

Identical k6 run against each build: constant **10 VUs, 15 min** (crosses the ~5-min TTL twice),
3 surfaces, deployed to the same namespace, hitting the action origin directly. Comparable applied
load (10,705 vs 10,502 requests, **0 failures** both).

```mermaid
xychart-beta
    title "Odin promo-list fetches per 15-min run (lower = better)"
    x-axis ["main (before)", "this PR (after)"]
    y-axis "Total fetches" 0 --> 25
    bar [22, 6]
```

| Metric | main (before) | this PR (after) |
|---|---|---|
| **Total promo-list fetches** | **22** | **6**  (−73%) |
| Cold-start burst | **10** fetches in 58 ms | **3** in 39 ms (1 / surface) |
| Expiry events | 9-wide burst, then 3 — **exact 5-min lockstep** | lone fetches **scattered over ~56 s** |
| Per surface | acom 7 / ccd 9 / express 6 | 2 / 2 / 2 |

**Herd shape** — every request that missed used to fetch; now they coalesce:

```mermaid
xychart-beta
    title "Fetches fired at each cache-miss event (peak concurrency)"
    x-axis ["cold start", "expiry 1", "expiry 2"]
    y-axis "Simultaneous fetches" 0 --> 12
    bar [10, 9, 3]
    bar [3, 1, 1]
```
_First series = before, second = after._

- **Single-flight** collapses each burst of concurrent misses to **one fetch per surface**.
- **Stale-while-revalidate** serves the previous list during a refill, so a slow/failed refresh no
  longer drops promotions.
- **Jitter** desyncs per-surface expiry, replacing the synchronized spike with a trickle.

<details>
<summary>Method & caveat</summary>

Requests are stamped `x-request-id=<run>-<surface>-<vu>-<iter>`, which the pipeline propagates to
Odin as `X-Correlation-ID`, so promo-list fetches are counted exactly in Splunk (`fetch` +
`path=/content/dam/mas/promotions&limit=50`), isolated from all other traffic. Test lives at
`scripts/k6/promotions-herd.js`.

At 10 VUs the local runs never overload Odin, so **both show 0 504s** — the 66%-failure /
15k-dropped-promotions win from the prod baseline is *inferred* (single-flight removes the load that
causes those 504s), not reproduced locally. Prod's 160-wide herd collapses far harder than 22 → 6.
</details>

Resolves https://jira.corp.adobe.com/browse/MWPW-202263

## Test URLs:

- After: https://MWPW-202263--mas--adobecom.aem.live/

### regressions:
- https://www.adobe.com/kr/creativecloud/plans.html?mas-io-url=https:\/\/14257-merchatscale-npeltier.adobeioruntime.net/api/v1/web/MerchAtScale
- https://www.adobe.com/products/indesign/plans.html?mas-io-url=https:\/\/14257-merchatscale-npeltier.adobeioruntime.net/api/v1/web/MerchAtScale


io/www/src/fragment/transformers/promotions.js
```diff
@@ -99,18 +120,55 @@ async function fetchProjects(context) {
...
+        logDebug(() => `Using cached promotion projects for surface "${surface}"`, context);
+        return entry.projects;
+    }
+    if (!refills[surface]) {
+        refills[surface] = refillProjects(context, surface).finally(() => {
+            refills[surface] = undefined;
+        });
+    }
+    if (entry) {
+        logDebug(() => `Serving stale promotion projects for surface "${surface}" while refilling`, context);
```
honstar, line 171: As per the ticket's AC: "Preview mode retains blocking refetch behaviour." Looks like this function is a shared code-path for both preview (studio) and published reads. Either this needs to be split or the ticket's AC adjusted to reflect the actual intended behavior.
There is also no corresponding test that advances the clock past the cache TTL for a preview context to check that the refetch is actually blocking and not serving stale cache entries.

io/www/src/fragment/transformers/promotions.js
```diff
@@ -363,22 +418,22 @@ async function hydrateProject(project, { baseUrl, surface, defaultLocale, resolv
...
-
-    // Resolve surface, projects, and defaultLanguage (which carries regionLocale) all in parallel.
+    // The listing is surface-scoped now, so the projects fetch gates on the surface resolved by the
+    // first fragment fetch. defaultLanguage (which carries regionLocale) still resolves in parallel.
     // regionLocale is NOT available on the init-phase context — it is computed by defaultLanguage.init
     // and only placed on context during the process phase. We must read it from the promise.
-    const [{ surface }, projects, defaultLangResult] = await Promise.all([
-        getRequestInfos(context),
+    const surfacePromise = getRequestInfos(context).then((infos) => infos.surface);
+    const projectsPromise = surfacePromise.then((surface) => (surface ? fetchProjects(context, surface) : null));
```
honstar, line 440: Untested branch in case `surface` resolves to `null`. This would break the 100% code coverage requirement for io/www.

io/www/src/fragment/transformers/promotions.js
```diff
@@ -99,18 +120,55 @@ async function fetchProjects(context) {
...
+            endDate: fields?.endDate ?? null,
+            tags: fields?.tags ?? [],
+        }))
+        .filter((project) => project.surfaces.includes(surface));
+}
+
+/**
+ * Refills the per-surface cache from the folder listing. On success, replaces the surface's
+ * entry with a fresh timestamp and jittered TTL; on failure leaves any existing (stale) entry
+ * untouched. Returns the fetched projects, or `null` on failure.
```
honstar, line 140: The doc for this function claims that "on failure, it leaves the existing (stale) entry untouched", but there is no test that actually triggers a refill failure while an entry is stale, to confirm
(a) the stale entry survives and
(b) `refills[surface]` is still cleared afterward so a later request can retry.

honstar (changes requested):

io/www/src/fragment/transformers/promotions.js
npeltier: there is a limit in the code for this (you can't have successful build without 100%) but there is still some coverage "gaps", got you

honstar: Right, but unit-tests were failing (flaky non-related tests) and codecov didn't run at the time.

honstar (approved):
LGTM

mirafedas (approved):

afmicka (approved):

PR #1103, opened 2026-07-30
MWPW-202349: Settings-based placeholder key remap (Option B)

yesil:
### Settings-based placeholder key remap (Option B) — MWPW-202349

A generic `placeholderRemap` setting rewrites a placeholder key (`from:to`, one per line) in `fragment.fields` during the `settings` step — before `replace` — scoped by locale + tag via the existing `resolveSettingEntry`. Lets IL Teams cards resolve `{{annual-billed-monthly}}` to a segment-specific value with no per-fragment authoring; one authored rule covers all current and future Teams cards.

Resolves https://jira.corp.adobe.com/browse/MWPW-202349

### Tests
- io/www unit suite green at 100% coverage (lines/branches/functions/statements).

Test URLs:

Notice the difference in the call out text in the teams card (3rd)

- Before: https://main--da-cc--adobecom.aem.page/drafts/ilyas/abm-to-m2m-test-page?akamaiLocale=il&commerce.env=stage&mas-geo-detection=true

- After: https://main--da-cc--adobecom.aem.page/drafts/ilyas/abm-to-m2m-test-page?akamaiLocale=il&commerce.env=stage&maslibs=mwpw-202349--mas--adobecom&mas-geo-detection=true


io/www/src/fragment/transformers/settings.js
```diff
@@ -212,15 +215,43 @@ export function resolveSettingEntry(fragment, locale, setting) {
...
+
+export function applyPlaceholderRemaps(fragment, remaps) {
+    const entries = Object.entries(remaps);
+    if (!fragment?.fields || entries.length === 0) return;
+    let fieldsString = JSON.stringify(fragment.fields);
+    for (const [from, to] of entries) {
+        const pattern = new RegExp(`{{\\s*${from.replace(/[.*+?^${}()|[\]\\]/g, '\\$&')}\\s*}}`, 'g');
+        fieldsString = fieldsString.replace(pattern, `{{${to}}}`);
+    }
+    fragment.fields = JSON.parse(fieldsString);
```
Andrei4226: maybe we can protect the request with a try/catch block
(I think that in case of a parse failure, `fragment.fields` will be kept unchanged)

honstar: `to` is spliced into the replacement string raw — no escaping for JSON-string context (a " or \ in to produces invalid JSON and an uncaught `JSON.parse` throw) or for `String.replace`'s special patterns (`$&`, `$$`, etc.).

io/www/src/fragment/transformers/settings.js
```diff
@@ -212,15 +215,43 @@ export function resolveSettingEntry(fragment, locale, setting) {
...
+        if (from && to) remaps[from] = to;
+    }
+    return remaps;
+}
+
+export function applyPlaceholderRemaps(fragment, remaps) {
+    const entries = Object.entries(remaps);
+    if (!fragment?.fields || entries.length === 0) return;
+    let fieldsString = JSON.stringify(fragment.fields);
+    for (const [from, to] of entries) {
```
honstar: The settings UI is making use of `sp-textfield` with the `multiline` attribute, so it is possible that an author types multiple `from:to` lines into one `placeholderRemap` entry. This loop would mutate an already mutated string multiple times, effectively leading to rule chaining.
Not sure if this is an issue or not but worth checking to avoid negative side-effects.

yesil: Good catch — fixed in ad5d08b6. Rewrote it as a single pass with one alternation regex (`{{\s*(a|b|c)\s*}}`), so each `{{…}}` is matched once against all keys and a target that happens to equal another rule's source can no longer be re-matched. Added a test with `{a:b, b:c}` asserting `{{a}} {{b}}` → `{{b}} {{c}}`.

io/www/src/fragment/transformers/settings.js
yesil: Fixed in ad5d08b6. Switched to a function replacement (`(match, key) => ...`), so `$&`/`$$` in the target are treated literally instead of as `String.replace` patterns. For the JSON-string case (a `"` or `\` in the target producing invalid JSON) the parse is now wrapped in try/catch — see the reply to Andrei's comment. Tests cover both.

yesil: Done in ad5d08b6 — `JSON.parse` is now wrapped in try/catch; on failure `fragment.fields` is left unchanged (and logged at debug when a context is present). Added a test for the invalid-JSON path.

io/www/src/fragment/transformers/settings.js
```diff
@@ -212,15 +215,49 @@ export function resolveSettingEntry(fragment, locale, setting) {
...
+    const entries = Object.entries(remaps);
+    if (!fragment?.fields || entries.length === 0) return;
+    const escaped = entries.map(([from]) => from.replace(/[.*+?^${}()|[\]\\]/g, '\\$&'));
+    const pattern = new RegExp(`{{\\s*(${escaped.join('|')})\\s*}}`, 'g');
+    const fieldsString = JSON.stringify(fragment.fields).replace(pattern, (match, key) => `{{${remaps[key.trim()]}}}`);
+    try {
+        fragment.fields = JSON.parse(fieldsString);
+    } catch {
+        if (context)
+            logDebug(
```
honstar: This would silently fail on production. Maybe it'd be better to log a warning.

io/www/src/fragment/transformers/settings.js
```diff
@@ -212,15 +215,49 @@ export function resolveSettingEntry(fragment, locale, setting) {
...
+    }
+    return remaps;
+}
+
+export function applyPlaceholderRemaps(fragment, remaps, context) {
+    const entries = Object.entries(remaps);
+    if (!fragment?.fields || entries.length === 0) return;
+    const escaped = entries.map(([from]) => from.replace(/[.*+?^${}()|[\]\\]/g, '\\$&'));
+    const pattern = new RegExp(`{{\\s*(${escaped.join('|')})\\s*}}`, 'g');
+    const fieldsString = JSON.stringify(fragment.fields).replace(pattern, (match, key) => `{{${remaps[key.trim()]}}}`);
```
honstar: Nit: `key.trim()` is redundant since the regex capture group already excludes the surrounding whitespace.

io/www/src/fragment/transformers/settings.js
yesil: ~Agreed — switched to `logError` in a42087b0, so a malformed remap now surfaces in production activation logs regardless of `debugLogs`. Test asserts it logs with `debugLogs: false`.~

io/www/src/fragment/transformers/settings.js
yesil: Right, removed in a42087b0 — the `\s*` is outside the capture group so the key is already trimmed.

io/www/src/fragment/transformers/settings.js
yesil: (For context: `logDebug` is already used several times in this file, and behind the scenes it just calls `console.log` — same Splunk-forwarded sink as `log()`, only gated on `debugLogs`.)

PR #1106, opened 2026-07-30
MWPW-202595: preview reuses the page's commerce WCS config (single call, no prod CORS)

yesil:
In the browser (preview), the merch-card preview ran its own WCS path that ignored the page's commerce environment: `previewFragment` used a hardcoded prod WCS config, so it issued a duplicate prod `web_commerce_artifact` call that CORS-failed for markets with no prod offer (e.g. IL), while the live commerce service correctly called stage.

This makes preview reuse the page's commerce config so a single WCS call is made on the page's environment:

- `previewFragment` derives `wcsConfiguration` from the page's `mas-commerce-service` settings (env + endpoint), instead of the hardcoded prod default.
- `landscape=ALL` on stage, matching the commerce service's own WCS client.
- Drops the `X-Request-ID` header in preview — a custom header forced a CORS preflight that WCS rejects; the server pipeline still sends it.

Server-side behaviour is unchanged (only the browser preview path is affected).

Resolves https://jira.corp.adobe.com/browse/MWPW-202595

### Tests
- io/www unit suite green at 100% coverage (lines/branches/functions/statements).

Test URLs:

- After: https://mwpw-202595--mas--adobecom.aem.live/


honstar (approved):

npeltier (approved):
lgtm

afmicka (approved):

PR #1111, opened 2026-07-31
MWPW-200407: render edu size as standalone two-column pro card

joaquinrivero:
Resolves https://jira.corp.adobe.com/browse/MWPW-200407
QA Checklist: https://wiki.corp.adobe.com/display/adobedotcom/M@S+Engineering+QA+Use+Cases

Please do the steps below before submitting your PR for a code review or QA

- [x] C1. Cover code with Unit Tests
- [ ] C2. Add a Nala test (double check with #fishbags if nala test is needed)
- [ ] C3. Verify all Checks are green (unit tests, nala tests)
- [ ] C4. PR description contains working Test Page link where the feature can be tested
- [ ] C5: you are ready to do a demo from Test Page in PR (bonus: write a working demo script that you'll use on Thursday, you can eventually put in your PR)
- [ ] C.6 read your Jira one more time to validate that you've addressed all AC's and nothing is missing

## 🧪 Nala E2E Tests

Nala tests run automatically when you **open this PR**.

### To run Nala tests again:

1. Add the `run nala` label to this PR (in the right sidebar)
2. Tests will run automatically on the current commit
3. Any future commits will also trigger tests **as long as the label remains**

### To stop automatic Nala tests:

- Remove the `run nala` label

> **Note**: Tests only run on commits if the `run nala` label is present. Add the label whenever you need tests to run on new changes.

## Test URLs:

- Before: https://main--mas--adobecom.aem.live/
- After: https://main--milo--adobecom.aem.page/merch/mas/preview?fragment-id=8229b73d-ba0a-42f6-bd25-8dfbc37b1f0b&content-type=merch-card&locale=en_US&maslibs=mwpw-200407

PR #1113, opened 2026-08-03
MWPW-200406: Add dark theme to the Pro card

antonio-rmrz:
Resolves https://jira.corp.adobe.com/browse/MWPW-200406

Renders the dark Pro card, and ties Border Color to the theme in the editor. #1093 added the Theme=Dark picker (which stamps `background-color="dark"` on the card); this renders that attribute as the Figma dark design. Light is unchanged.

- Colour only — `background-color="dark"` re-points the `--consonant-merch-card-pro-*` tokens to the Figma dark values; no layout change, so all breakpoints are covered, and the old light values stay as fallbacks
- Values from the Figma dark nodes: hero `#000`, card/features/licence dropdown `#131313`, dividers `rgba(0,0,0,.12)`, white / muted-white text, `#dadada` outline CTA, muted-grey secure lock
- Editor (pro only): Border Color now offers just Default + Black, and is disabled while Theme = Dark
- Regenerated the web-components/dist bundles

Built on #1093; branch cut from main.

## Test URLs:

- https://mwpw-200406--mas--adobecom.aem.live/studio.html#fragmentId=248fef33-adea-4c20-ad69-565838afd44c&page=fragment-editor&path=sandbox


Roycethan (changes requested):
@antonio-rmrz Feedbacks:
https://main--da-cc--adobecom.aem.page/drafts/reuben/dark?maslibs=mwpw-200406

1) heading (title) is h5 per figma:
Image

2) Section headers (Apps) are of font weight : 400 and body-sm regular style per figma

Image

3) Clickable links are rendering as Blue instead of white on background
image

4) alignments across the cards needs to be adjusted
image

joaquinrivero (approved):
This is aligned with https://github.com/adobecom/mas/pull/1111

antonio-rmrz:
@Roycethan thanks for catching these.

**2 — fixed.** You're right, the section headers should be regular, not bold. They were hardcoded to bold in the original Biz Pro build and never updated.

While checking that I compared the rest of the card's text to Figma too, and we were adding a hair of letter spacing to all the 14px text (and a bit more on the callout) where Figma has none. That's gone as well — everything matches the type styles now, with only the title and price keeping their -0.48.

Worth a sanity check on your end: none of this is split by theme, so it changes the light card too. Your screenshots are from the dark comp — can you confirm light uses the same type styles? If anything should differ there I'll scope it.

**1 — can we keep the title as `<h3>`?**

It already renders exactly what heading-5 specifies (Adobe Clean Display, Black 900, 24/24, -0.48), so nothing's actually off there. I read "heading-5" as the name of the text style rather than the tag — in the same file the eyebrow layer is called "H3 - Headline" but uses the eyebrow style.

Switching to `<h5>` wouldn't change how it looks, but it would break two things:

- the section headers ("Apps", "Extras") are `<h4>`, so screen readers would rank those above the card title
- Milo reads the product name off the `<h3>` for the buy button's aria-label and for analytics — both would come back empty

Happy to change it if you do mean the tag, just want to be sure first.

